PR #4, opened 2014-07-26
QFJ-802

manureno:
* Merge messages modules (except fix50sp1 and fix50sp2) : new merged module is quickfixj-messages-fix
* Move from quickfixj-core to quickfixj-messages-fix the classes required to compile the code generated in messages modules
* Make quickfixj-core depend on quickfixj-messages-all


manureno:
Another strategy should be implemented for QFJ-802

PR #7, opened 2015-01-09
QFJ-822 Refresh DNS resolution info on connection error.

dale-wilson:
This is a fix for the problem I reported in QFJ_822.
When a connection attempt fails, it clears the cached DNS resolution information so that the next connection attempt will repeat the resolution -- thereby picking up any DNS changes for the host.

PR #8, opened 2015-01-20
Revert "QFJ-822 Refresh DNS resolution info on connection error."

chrjohn:
Reverts quickfix-j/quickfixj#7


chrjohn:
Sorry, had to revert this. InetSocketAddress.getHostString() is only available starting from Java7, but QF/J still uses Java6.

chrjohn:
Should work now after we have switched to JDK 7 compatibility for master branch.

PR #11, opened 2015-02-10
Bump JUnit and Mockito dependencies.

guidomedina:
* JUnit upgraded to 4.10, can't do 4.11+ for now due to order of execution for Java 7+, which can be fixed by annotating each test class though it is tedious to do so now, might be done later when I have more time.
* Mockito upgraded to latest stable which made me gain some tests performace; used to take almost 13 minutes on my laptop to do `mvn clean test` and now it is taking a little bit less than 12 minutes.


chrjohn:
Thanks again. :)

PR #12, opened 2015-02-11
IntelliJ code inspection fixes and optimizations.

guidomedina:
This pull request has the following changes:
* Several fields -local and static- modifiers are now promoted to final to somehow state that they are inmutable.
* Redundant throws `SubclassException` have been removed when the method throws `SuperException`, example: `public void doSomething() throws IOExceptio, UnknownHostException`, in this example `UnknownHostException` is unnecessary because it is a subclass of `IOException`
* Some complex `if` statements have been reduced to a simpler form where one liner return would suffice.
* Some `StringBuilder` were using concatenation instead of consistently using the `append(...)` method.

**Note:** No functional changes made, just code inspections suggestion optimizations, changes kept to minimal.

I could had done more but I'm trying to divide the amount of optimizations needed in small units of work to make code review easy, next I might optimize some `Locking and Synchronization` and instead use more `AtomicReferences`

### DIFF
--- a/quickfixj-codegenerator/src/main/java/org/quickfixj/codegenerator/GenerateMojo.java
+++ b/quickfixj-codegenerator/src/main/java/org/quickfixj/codegenerator/GenerateMojo.java
@@ -83,7 +83,7 @@ public class GenerateMojo extends AbstractMojo {
      *
      * @parameter default-value = "quickfix.field"
      */
-    private String fieldPackage = "quickfix.field";
+    private final String fieldPackage = "quickfix.field";
 
     /**
      * The Maven project to act upon.
--- a/quickfixj-codegenerator/src/main/java/org/quickfixj/codegenerator/MessageCodeGenerator.java
+++ b/quickfixj-codegenerator/src/main/java/org/quickfixj/codegenerator/MessageCodeGenerator.java
@@ -28,7 +28,6 @@ import java.io.PrintStream;
 import java.text.DecimalFormat;
 import java.util.ArrayList;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
@@ -81,8 +80,8 @@ public class MessageCodeGenerator {
         e.printStackTrace();
     }
 
-    private void generateMessageBaseClass(Task task) throws TransformerConfigurationException,
-            FileNotFoundException, ParserConfigurationException, SAXException, IOException,
+    private void generateMessageBaseClass(Task task) throws
+            ParserConfigurationException, SAXException, IOException,
             TransformerFactoryConfigurationError, TransformerException {
         logInfo(task.getName() + ": generating message base class");
         Map<String, String> parameters = new HashMap<String, String>();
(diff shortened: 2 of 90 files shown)


chrjohn:
Thanks for this. Will check it.
Re synchronization changes: I'd rather wait until after 1.6.0 has been released before I merge.
Cheers

guidomedina:
Agreed, I will actually start using QuickFIX/J tomorrow, lots of testing to do for our project, once 1.6.0 is out I will spend a day or two working with the synchronization and try to optimize them where possible.

I have done plenty of concurrency in the past to a reasonable advanced level so it should be easy for me to take over that part.

I see several usages of `ConcurrentMap` without its atomical usage `putIfAbsent` and instead a lock is used generally, those can be replaced by a `LoadingCache` from Google's Guava but such thing would introduce a dependency (a good one but still a dependency)

Here is a detailed explanation of its usage which I have extensively used and it works like a charm, let me know your thoughts about it: https://code.google.com/p/guava-libraries/wiki/CachesExplained

guidomedina:
One thing I don't like about ConcurrentMap is that it assumes a concurrency level of 16 and in theory for very fast multi-threaded operations concurrency level shouldn't exceed the amount of CPU, in some projects I have added `MapBuilder` from Guava and always set the concurrency to `Runtime.getRuntime().getAvailableProcessors()` then tried millions of values read/written and when you exceed the amount of CPU it usually becomes slower, so, coming back to the subject, most servers have from 4 to 8 CPUs, 16 is too much, and it is slower than using exactly the amount of CPU, just saying, I have a fair amount of time working with in-memory caches and concurrent maps so hopefully I can help in this matter.

chrjohn:
Thanks for your suggestions and input. Really appreciated.

chrjohn:
IMHO QuickFIX/J has never been designed to be very performant (although its performance probably is not that bad). But I guess it will not hurt to do some little optimizations where sensible.
Thanks in advance!

guidomedina:
Branch rebased from Pull request #13

chrjohn:
I cannot merge this pull request due to the following chang in Session.java.
The responderSync should be a unique instance per Session class. But it has been converted to a constant string which makes it effectively a class level lock.

guidomedina:
Do you mean changing `new String("SessionResponderSync")` for `"SessionResponderSync"`? Java will create a new String anyway and will still act as a valid lock -using intern() which should be avoided- or do you mean that there is a conflict?

If there is a conflict I can rebase this branch on top of the changes you are planning to add and resolve the conflicts myself which I'm good at.

chrjohn:
No, there is no conflict. I mean that "new String("bla")" will create a new String per Session instance. Whereas writing only "bla" will only create that String once per JVM and all created Session instances will all use the same lock which is not wanted.

guidomedina:
That's a waste of memory, why not just final `Object lock = new Object()`, which is the smallest in-memory lock that you will ever have, for that reason I guessed it was OK to replace it with a literal, that's really bad code, confusing, etc. Let me know if you want me to convert that into just an internal lock. It should take me very few minutes.

guidomedina:
Fixed, renamed `responderSync` for `responderLock` and now it is just a `private final Object responderLock`

chrjohn:
I guess it was there for historical reasons. ;) 
Converting it to a new Object() lock should be fine. Ah, I see you already did it. Thanks!

PR #14, opened 2015-03-13
Bump Maven plugins version

guidomedina:
Bump Maven plugins version, except for `maven-assembly-plugin` which latest version 2.5.3 has a bug and hence was bumped only to 2.5.2, bug described below:

https://jira.codehaus.org/browse/MASSEMBLY-747


chrjohn:
Thanks :)

PR #15, opened 2015-03-13
Bump Maven version to 3.2.5

guidomedina:
Update pre-requisite to latest Maven, this will force contributors to use latest Maven which could be a hassle to update but the benefit is that it will also inherit newer plugins overall, also less whole system overload while building QuickFixJ.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,7 +24,7 @@
 	</licenses>
 
 	<prerequisites>
-		<maven>3.0.4</maven>
+		<maven>3.2.5</maven>
 	</prerequisites>
 
 	<modules>


chrjohn:
Hmm, is there something special about this? I specified 3.2.5 as prerequisite but still can build with Maven 3.0.4?

guidomedina:
Forces the plugin versions related to Maven to be at the minimum required for the prerequisite, for example, `site-plugin` can be at version A.B but saying maven minimum will be 3.2.5 it will force the plugin to be greater than A.B if that's required, also allows to omit the plugin versions.

guidomedina:
Here is kind of explained: http://maven.apache.org/ref/3.2.5/maven-model/maven.html#class_prerequisites

guidomedina:
And what it did in your case is that it downloaded the Maven 3.2.5 needed jars to be able to build.

PR #16, opened 2015-03-26
Maven -Dgenerator.decimal=true will generate classes using BigDecimal

guidomedina:
Maven -Dgenerator.decimal=true will generate classes using BigDecimal.

@chrjohn The other 2 pull requests I sent I can assure you they are harmless and a good thing, for example, IntelliJ has done its best effort to migrate specific classes to use proper Java 6 which is the minimum requirement for QuickFixJ 1.6+

And forcing maven to 3.2.5 is not a bad thing either, Maven 3.3.1 was released so that would be the latest 3.2.x as requirement which does a better job than 3.0.x or 3.1.x

Is there any other parameter that needs to be fixed that isn't working when the project was migrated from Ant to Maven?

As for BigDecimal, in this branch it keeps the default value which is false but this is something that nobody would recommend for financial applications.

### DIFF
--- a/quickfixj-core/pom.xml
+++ b/quickfixj-core/pom.xml
@@ -156,6 +156,7 @@
 							<dictFile>../quickfixj-messages/quickfixj-messages-fixt11/src/main/resources/FIXT11.xml</dictFile>
 							<packaging>quickfix.fixt11</packaging>
 							<fieldPackage>quickfix.field</fieldPackage>
+							<decimal>${generator.decimal}</decimal>
 						</configuration>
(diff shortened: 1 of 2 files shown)


chrjohn:
Thanks. I will have a look at the other pull requests now.
Regarding BigDecimal: there were many discussions around it but IMHO there are more people using the default build than the BigDecimal one. But I guess this largely depends on the fact whether you just transfer values from one system to the other or if you do calculations with the values you receive via FIX. The values are only Strings after all.

PR #18, opened 2015-04-03
Fix timing to set default application version ID at initiator side

siburu:
- There are (at least) 2 threads at each host.
    - IO thread : Receives a message from an opposite and put it onto queue.
    - Application thread : Takes the message from the queue and processes it.
- Probelm in the original implementation
    - At initiator side, _Application thread_ sets default application version ID when receiving a first LOGON message.
    - IO thread needs default application version ID to be set to parse incoming messages.
    - If IO thread attempts to parse a message before Application thread sets default application version ID, it causes error.
- Fixed version
    - At initiator side, _IO thread_ sets default application version ID when receiving a first LOGON message.
    - By the way, at acceptor side, IO thread sets default application version ID from the beginning.

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/mina/initiator/InitiatorIoHandler.java
+++ b/quickfixj-core/src/main/java/quickfix/mina/initiator/InitiatorIoHandler.java
@@ -22,7 +22,12 @@ package quickfix.mina.initiator;
 import org.apache.mina.core.session.IoSession;
 
 import quickfix.Message;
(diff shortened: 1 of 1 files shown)


chrjohn:
Hi,
thanks for the pull request. It seems this is the same problem as in http://www.quickfixj.org/jira/browse/QFJ-592 but there it was only fixed on acceptor side.
Cheers,
Chris.

PR #19, opened 2015-04-14
Moving the setLogoutReceived flag to before we generate a logout response

bilewinters:
This means we can query this flag in the toAdmin callback.


chrjohn:
Thanks for the pull request. Do all tests still pass with your change? I'm just asking since I recall there were some tests around these state flags.

Cheers,
Chris.

bilewinters:
Yes, the current tests pass. I had a look through to see if it was worth adding a test for this, but it didn't make much sense to me, looking at the other tests that are there. I can add one if you think it needs one?

PR #20, opened 2015-04-28
Range of resend request is wrong after GapFill when resends are chunked.

GaetanVaysse:
We may have found an issue with ResendRequest done after a GapFill SequenceReset when resends are chunked.

The SequenceReset NewSeqNo field is the seq no of the next expected field, not of the last message replaced by the SequenceReset; for instance http://btobits.com/fixopaedia/fixdic44/message_Sequence_Reset_4_.html defines NewSeqNo like this:
> The message in all situations specifies NewSeqNo (36) to reset to as the value of the next sequence number to be expected by the message receipient immediately following the messages and/or sequence numbers being skipped." ).
But the ResendRequest is actually done using NewSeqNo+1 in Session.nextSequenceReset(), which mean that one message may be skipped over.

For instance, in fix42/SequenceGapFollowedBySequenceResetWithGapFill.def:

    # Send a GapFill up to the last message 13
    I8=FIX.4.035=434=849=TW52=<TIME>56=ISLD123=Y36=13
       # 36=13 => next message in sequence is 13, so we should request 13  
      
    # Receive Resend request for messages 14 to 15)
    E8=FIX.4.09=6035=234=649=ISLD52=00000000-00:00:0056=TW7=1416=99999910=0
       # 7=14 => but here we request 14


chrjohn:
Thanks for the request, I'll have a look at it.

PR #26, opened 2015-05-19
Add CI build on travisci

kimptoc:
See here for how to enable it for the quickfix-j user - http://docs.travis-ci.com/user/getting-started/

Please wait and see if this build of mine works first :)

https://travis-ci.org/kimptoc/quickfixj/builds/63240864

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,6 @@
+language: java
+jdk:
+#  - oraclejdk8
+#  - oraclejdk7
+#  - openjdk7
+  - openjdk6
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 QuickFIX/J
 ==========
 
+Master: [![Build Status](https://travis-ci.org/quickfix-j/quickfixj.png?branch=master)](https://travis-ci.org/quickfix-j/quickfixj)
+
 This is the official QuickFIX/J project repository.
 
 The Financial Information eXchange (FIX) protocol is a messaging standard developed


kimptoc:
Hmm - it failed, will investigate tomorrow :)

chrjohn:
Hmm, we also have Bamboo build here: http://www.quickfixj.org:8085/browse/QFJ-GIT 
:)

.travis.yml
```diff
@@ -0,0 +1,6 @@
+language: java
+jdk:
+#  - oraclejdk8
+#  - oraclejdk7
+#  - openjdk7
+  - openjdk6
```
chrjohn: Is there a reason why you chose openjdk instead of the Oracle one?

guidomedina:
Try Oracle JDK 7 and remove/comment fork=true on the pom files since the build won't work with that option.

kimptoc:
Whoops - did not notice the Bamboo build - Looks like there is a Bamboo plugin so that its status can be included around the web - https://marketplace.atlassian.com/plugins/com.atlassian.bamboo.plugins.atlassian-bamboo-plugin-build-status-image

chrjohn:
@kimptoc Unfortunately, that plugin works only with Bamboo up to 5.2 but we have 5.4.

kimptoc:
Don't know if you have a server to run this on (e.g. free on heroku :) ) - but this node app seems to do the job - https://github.com/kimptoc/bamboo-status

eg http://overzold1.kimptoc.net:8085/status/QFJ-GIT

chrjohn:
Closed in favour of #121 
Sorry, @kimptoc , totally forgot about this one. :)

PR #28, opened 2015-05-21
Maven plugins updated and compiler fork=true option removed

guidomedina:
Maven plugins updated and compiler `fork=true` option removed because it breaks the build, also, Maven prerequisite updated to latest 3.3.3 to keep default plugins at latest required versions.

All other pending pull requests should build properly after this one is merged, excluding Travis which should be set to Oracle JDK 7 which won't interfere with JDK 6 requirement of the project but will build faster due to JDK 7 performance vs JDK 6.

**Note:** Apache Felix bundle plugin makes build fail if updated to 2.5.4, so I left it at 2.5.3

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,7 +24,7 @@
 	</licenses>
 
 	<prerequisites>
-		<maven>3.2.5</maven>
+		<maven>3.3.3</maven>
 	</prerequisites>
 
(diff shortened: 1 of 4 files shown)


chrjohn:
Thanks, will merge this during the next days. :)

PR #32, opened 2015-07-29
Update FIXMessageEncoder.java

charlesbr1:
Removed "Invalid FIX message object type : " error in encode(...) and support any message type, as the calling code is clean enough.

Biggest allocations removed : 
For each message sent, the previous version create garbages of size of the data to send multiplied by 2 (one bytes array and a direct ByteBuffer instance).
This version creates just one wrapped CharBuffer instance and reuses some thread allocated buffers.
It allocates 16 buffers of 8192 bytes by threads that is sending a message, more buffers are created on fly as before if needed, but only 16 will be kept in memory.

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/mina/message/FIXMessageEncoder.java
+++ b/quickfixj-core/src/main/java/quickfix/mina/message/FIXMessageEncoder.java
@@ -20,11 +20,18 @@
 package quickfix.mina.message;
 
 import java.io.UnsupportedEncodingException;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.Set;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.nio.CharBuffer;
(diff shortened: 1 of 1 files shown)


charlesbr1:
Hi,

I agree with you, either MAX_BUFFERS_BY_THREAD  and DEFAULT_BUFFER_SIZE should be in conf.

Futhermore, the charset convertion may be completly skipped by having a isStringEquivalent() test like I did in other commit. This kind of method should be in CharsetSupport class for instance.

For the buffers pool, we don't have the issue you talk about because the value for previous parameters suit us well. A simple way to solve it is to use a boolean in the SimpleBuffer class, to know if she's was preallocated or not, this way we can easily choose to save it or not.
ie, instead of doing :
if(ioBuffers.size() < MAX_BUFFERS_BY_THREAD)
     ioBuffers.add(this);

We can do something like :
if(this.isPreallocated)
   ioBuffers.add(this)

PR #39, opened 2015-07-31
Message generation length & checksum optimization

charlesbr1:
Use a thread local to reuse the string builder used to generate the fix message.
 Perform message length & checksum calculation after the message has been build, this avoid to iterate over the fields multiple time, and avoid the allocation of some strings many times too.

This is the most important optimization for allocations.

When the charset used to send fix message is not string equivalent to the jvm one, this optimization is not active. (But the code to compute length & checksum in this case will be optimized in a next pull request).

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/Message.java
+++ b/quickfixj-core/src/main/java/quickfix/Message.java
@@ -129,6 +129,21 @@ public class Message extends FieldMap {
         return message;
     }
 
+    private static final class Context {
+        private final BodyLength bodyLength = new BodyLength(100);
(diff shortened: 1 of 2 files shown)


chrjohn:
There are some problems in the MessageTest when parsing incomplete messages, i.e. where BeginString etc have not been added yet. Will improve the code.

charlesbr1:
Great job, thanks you @chrjohn  !

chrjohn:
No problem @charlesbr1 . Thanks for the contribution and sorry again for the slow merge. :)

PR #42, opened 2015-08-01
Fieldmap length and checksum optimization

charlesbr1:
Avoided some temporary IntField creation to compute length.
In fact, the "isStringEquivalent" branch that I added is useless with previous pull request on Message class, which directly compute length & checksum if possible.
Change some 'for each' list iteration by old loop over index to avoid iterator allocations, as the jvm will not always do it for us.

### DIFF
--- a/quickfixj-core/pom.xml
+++ b/quickfixj-core/pom.xml
@@ -59,7 +59,7 @@
         <dependency>
             <groupId>org.apache.mina</groupId>
             <artifactId>mina-core</artifactId>
-            <version>2.0.16</version>
+            <version>2.0.17</version>
         </dependency>
         <dependency>
             <groupId>org.slf4j</groupId>
--- a/quickfixj-core/src/main/java/quickfix/FieldMap.java
+++ b/quickfixj-core/src/main/java/quickfix/FieldMap.java
@@ -31,6 +31,7 @@ import quickfix.field.converter.IntConverter;
 import quickfix.field.converter.UtcDateOnlyConverter;
 import quickfix.field.converter.UtcTimeOnlyConverter;
 import quickfix.field.converter.UtcTimestampConverter;
+import org.quickfixj.CharsetSupport;
 
 import java.io.Serializable;
 import java.math.BigDecimal;
@@ -462,8 +463,9 @@ public abstract class FieldMap implements Serializable {
             } else if (isGroupField(tag) && isOrderedField(tag, fieldOrder)
                     && getGroupCount(tag) > 0) {
                 appendField(buffer, field);
-                for (Group group : getGroups(tag)) {
-                    group.calculateString(buffer, preFields, postFields);
+                List<Group> groups = getGroups(tag);
+                for (int i = 0; i < groups.size(); i++) {
+                    groups.get(i).calculateString(buffer, preFields, postFields);
                 }
             }
         }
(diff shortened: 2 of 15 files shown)


philipwhiuk:
NumbersCache is a bit of a memory/performance trade-off right? This is to do what - convert `9` into `"9"`?

What's the performance of the cache vs the performance of Integer.toString()?
What's the size of the cache. For the small one it's 2 bytes and 48 bytes for the string. Gauss's trick tells us there's ~500K of those so that's 25MB of data.

1000000 also seems arbitrary and high. The common numbers we see are going to be the fields themselves. Every order of magnitude dropped saves you over 90% of memory size (1 character less in the String and 90% less objects total).

charlesbr1:
Hi Philip,

Yes it was a memory trade off.
Getting a number from this cache is simply faster than Integer.toString() but main interest was to avoid String allocations.
It is not only used by the field number themselves, but for all the integer figures that are sent in a fix message (they end up by being converted into string, in the quick fix api). (I used this cache on others PR).
At Natixis they were a lot of garbage to save this way (hence reducing number of stop the world young gen gc), because we were sending thousands of messages each seconds.
I choose an arbitrary size of 1000000 because this value was a good trade off regarding the figures used in a trading environment, and memory size. 25mb of string allocated in old gen is peanuts on modern hardware, but indeed 250mb would be too much only for that...
That's why there is two caches, on for number until 1000000, and a second one for some number above, every 1000, ex : 1001000, 1002000, 5000000, etc… A number not in this cache is allocated the same way as before...

Regards,

Charles

chrjohn:
Hi @charlesbr1 , @philipwhiuk ,

do you think the big numbers cache makes sense?
Having all numbers from 0 to 999999 may make sense since the cache then contains every tag number, seqnum (unless you send really large numbers of messages), checksum and length that is going to be used. However, the values on the FIX messages (e.g. price, qty) are often floating point values so I am not really convinced to use it. Or maybe I am missing something?

Reducing the cached numbers to 99999 would probably be sufficient, too.

Thanks,
Chris.

charlesbr1:
Hi,
Yes you’re right. You can reduce it. The big numbers made sens on the Natixis FX engine, it was tuned for this context.
From what I remember, one should take care that group tag is using it...

From what I remember too, one simple thing you can take too, is the message checksum and lenght calculator. I may be wrong I think it was doing something like constructing the message two time to perform calculation. It was easy to rewrite it in a more efficient way...
Regards,

Charles.

chrjohn:
Hi @charlesbr1 , would it be possible to grant me permission to push to the PR branches of your PRs?

- E.g. on #35 I needed to make changes to the unit test but had to merge the PR on the command line eventually.
- From #40 I created a new PR from your fork so I could build via Travis CI.
But in both cases the commit looks like I authored the changes. :-/ If you grant me the permission I can make changes to the PRs and you would still be listed as the author of the commit.

Thanks,
Chris.

charlesbr1:
Hi Chris,

I just checked box ' Allow edits from maintainers.’ on PRs. 
Let me know if it is not what you requested.

Regards,

Charles

Learn more

chrjohn:
Hi @charlesbr1 
looking good now, at least github says "Add more commits by pushing to the fieldmap-length-and-checksum branch on charlesbr1/quickfixj.". That message was not present before.
Thanks!
Chris.

PR #43, opened 2015-08-09
Fix for QFJ-822: Reset cached DNS resolution information after a connection failure.

csharp:
This is a minor rework of a pull request submitted last January (https://github.com/quickfix-j/quickfixj/pull/7)  

The January request was originally merged but had to be reverted since it would not compile under jdk1.6.

I've done a (very minor) rework and it should build with Java 6 now.

Thanks


chrjohn:
Thank you for the adaption.

PR #44, opened 2015-08-12
SessionSettings loader

lburgazzoli:
This PR aims to: 
- move out of SessionSettings class the loading and parsing of settings file 
- support additional formats for settings file (like YAML which I've added)

 The PR is not 100% complete but I would like to know if there is interest in adding such feature and have feedback on how it has been implemented.


lburgazzoli:
Not actively working with quickfix-j anymore

PR #46, opened 2015-10-24
Session schedule optimization

charlesbr1:
Avoid the creation of many calendars during processing of next messsage


philipwhiuk:
Given we moved to Java 8 it would be nice to get rid of the old date time code entirely. I assume that the Java 8 time stuff is what the conflict is here.

chrjohn:
Superseded by #283

PR #48, opened 2015-11-26
QFJ-730 Corrected sequence number of last gap fill 

flakronv:
Corrected sequence number of last gap fill after all app messages have been sent in a resend request when the session is set up to not store admin messages


chrjohn:
Will merge in due course.

PR #49, opened 2015-12-09
QFJ-821 - acceptor/initiator certificate authentication

the-thing:
Implemented acceptor/initiator certificate authentication.

- new SSL properties:
KeyManagerFactoryAlgorithm (default X509)
KeyStoreType (default JKS)
SocketTrustStore
SocketTrustStorePassword
TrustManagerFactoryAlgorithm (default PKIX)
TrustStoreType (default JKS)
NeedClientAuth (default N)
EnabledProtocols (default - Java supported protocols)
CipherSuites (default - Java default cipher suites)

- removed quickifx.mina.ssl.SimpleTrustManagerFactory class as there is no need for it, MINA provides org.apache.mina.filter.ssl.BogusTrustManagerFactory which
serves the same purpose

- added quickfix.mina.ssl.SSLFilter class which fixes issues with MINA overriding cipher suites to Java enabled cipher suites

- comprehensive tests under quickfix.mina.ssl.SSLCertificateTest

- QFJ-838 and QFJ-854 had to be fix along the way as I encountered the same problems


chrjohn:
Thanks for the PR. :) Much appreciated.

chrjohn:
Hmm, on the Bamboo server two tests failed:

java.lang.AssertionError: No SSL exception thrown	at quickfix.mina.ssl.SSLCertificateTest.shouldFailWhenUsingEmptyServerTrustore(SSLCertificateTest.java:434)

ava.lang.AssertionError: No SSL exception thrown	at quickfix.mina.ssl.SSLCertificateTest.shouldFailWhenUsingEmptyClientTruststore(SSLCertificateTest.java:404)

http://www.quickfixj.org:8085/browse/QFJ-GIT-75/test

Any idea what's wrong there?

the-thing:
I will investigate.

chrjohn:
Hmm, cannot reproduce the errors on my local environment. Will try again on Bamboo.

chrjohn:
Still failing on Bamboo. The debug logging has been enabled. Would be great if you could take a look.
http://www.quickfixj.org:8085/browse/QFJ-GIT-79/test

Thanks

the-thing:
I think the timeout is the key to the problem here. For example on my local machine when I reduce the wait time to 15 seconds 'shouldFailWhenUsingEmptyClientTruststore' the test fails as well, because there is not enough time to raise the exception, although SSL session is not established.

I wonder if this assertion should be made at all. It was borrowed from SecureSocketTest. I will think about a better way of doing this type of assertion.

I'm still investigating 'shouldFailWhenUsingEmptyServerTrustore' failures.

chrjohn:
Hmm, why does it take so long to raise an exception? Does it depend on a message being sent or received? If yes, maybe we should lower the heartbeat interval or send a TestRequest.

the-thing:
SSL exception is raised after calling disconnect in quickfix.Session class, which happens after logon timeout.

```
} else if (state.isLogonAlreadySent() && state.isLogonTimedOut()) {
    disconnect("Timed out waiting for logon response", true);
}
```

the-thing:
The problem is when using empty truststore file with **sun.security.ssl.X509TrustManagerImpl**. **checkClientTrusted** and **checkServerTrusted** methods raise java.lang.RuntimeException instead of java.security.cert.CertificateException which is expected by sun.security.ssl.ClientHandshaker/ServerHandshaker to fail SSL handshake immediately. 

```
java.lang.RuntimeException: Unexpected error: java.security.InvalidAlgorithmParameterException: the trustAnchors parameter must be non-empty
	at sun.security.validator.PKIXValidator.<init>(Unknown Source)
	at sun.security.validator.Validator.getInstance(Unknown Source)
	at sun.security.ssl.X509TrustManagerImpl.getValidator(Unknown Source)
	at sun.security.ssl.X509TrustManagerImpl.checkTrustedInit(Unknown Source)
	at sun.security.ssl.X509TrustManagerImpl.checkTrusted(Unknown Source)
	at sun.security.ssl.X509TrustManagerImpl.checkServerTrusted(Unknown Source)
	at quickfix.mina.ssl.TrustManagerProxy.checkServerTrusted(TrustManagerProxy.java:47)
	at sun.security.ssl.AbstractTrustManagerWrapper.checkServerTrusted(Unknown Source)
	at sun.security.ssl.ClientHandshaker.serverCertificate(Unknown Source)
	at sun.security.ssl.ClientHandshaker.processMessage(Unknown Source)
```	

We could:

1) Increase exception latch timeout for these type of tests to make sure there is enough time to generate an exception on disconnect. Not very nice and QuickFix tests already take a long time.

2) Create X509TrustManager interface delegate which re-throws RuntimeException as CertificateException. Tests could be made a lot faster, but we are coupled with X509TrustManager. On the other hand this is the only algorithm standard JVM supports.
 
3) Get rid of SSLException assertions completely since logon and certificate checks do their thing.

Personally I would go with option 2. Any suggestions?

chrjohn:
Yes, option 2 sounds sensible to me as well.

the-thing:
Changes in https://github.com/quickfix-j/quickfixj/pull/54

dan-corneanu:
Hi @the-thing,
thank you for this PR.

Could it be possible to make the documentation a little bit more clear on what should go inside the trust-store and key-store? I would volunteer for this job if you think it is needed and if someone can review it.

Before this fix the configuration only accepted a SocketKeyStore. How was this working? does it mean that if you wanted both parties to authenticate, the keystore would have to contain both the trusted certificate of the remote-party and the public/private keypair and certificate of the local party?

Regards,
Dan

the-thing:
Hi @dan-corneanu 

I updated parts of the documentation as part of the pull request (https://github.com/quickfix-j/quickfixj/tree/master/quickfixj-core/src/main/doc ). This might not be enough as there are no step by step guidelines what to do.

Initially I would suggest you doing this:

1) Look at the documentation I updated.
2) Download and experiment with keystore tools such as keystore explorer and do some reading about how keystores and truststore work in general.
3) Look at SSLCertificateTest which has many different scenarios for client/server authentication (also use the keytool to look inside the keystores and truststores which are part of the test suite).

Regards,
Marcin

PR #50, opened 2015-12-18
Update Maven plugins, Slf4j & Mina core which fixes some SSL bugs

guidomedina:
Updated Maven plugins, Slf4j & Mina core which fixes some SSL bugs.
It might fix the QFJ SSL bugs, look at Mina core release notes at https://mina.apache.org/mina-project/index.html

Listed below for the lazy:

```
Bugs :

DIRMINA-992 - NioSocketConnector.newHandle throws the wrong exception
DIRMINA-994 - The ConnectionRequest.cancel() method is inconsistent wrt concurrent access
DIRMINA-995 - Deadlock when using SSL and proxy
DIRMINA-1013 -Threading model is supressed by ProtocolCodecFilter
DIRMINA-1016 - Regression with 2.0.9: Missing javax.net.ssl import in manifest
DIRMINA-1017 - SSLEngine BUFFER_OVERFLOW (unwrap)
DIRMINA-1019 - SslHandler flushScheduledEvents race condition

Improvements :

DIRMINA-1018 - fetchAppBuffer shrink
DIRMINA-1020 - Change minimum version for slf4j in MANIFEST
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,7 +24,7 @@
 	</licenses>
 
 	<prerequisites>
-		<maven>3.3.3</maven>
+		<maven>3.3.9</maven>
 	</prerequisites>
 
 	<modules>
@@ -41,7 +41,7 @@
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 		<jdkLevel>1.6</jdkLevel>
-		<slf4j.version>1.7.12</slf4j.version>
+		<slf4j.version>1.7.13</slf4j.version>
 		<mainClass/>
 	</properties>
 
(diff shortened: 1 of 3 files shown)


guidomedina:
I fully re-deployed in our archiva this commit on top of version 1.6.1, as a suggestion you might probably release 1.6.2 doing a cherry-pick.

chrjohn:
I have cherry-picked this into the 1.6.x branch so that it is incorporated into the next 1.6.x release.
Thanks

chrjohn:
Hi @guidomedina ,  could you please split this PR up into the MINA change and the rest of the plugin updates? The build is currently failing because the bundle-plugin 3.0 needs Java 7. But we only support Java 6.
I think we can switch to Java 7 for the 1.7.x branch but not for 1.6.x
Thanks in advance

guidomedina:
OK, I'll give you another PR with 2 commits based on 1.6.x, 1st commit will be this commit and 2nd commit setting the old bundle version.

PR #52, opened 2015-12-19
Revert "Revert "QFJ-822 Refresh DNS resolution info on connection error.""

chrjohn:
Reverts quickfix-j/quickfixj#8


chrjohn:
Should work now after we have switched to JDK 7 compatibility for master branch.

PR #59, opened 2016-01-27
Mina-core, Slf4j and Maven compiler plugin have been updated to latest

guidomedina:
Mina-core, Slf4j and Maven compiler plugin have been updated to latest


guidomedina:
@chrjohn this is the original branch, for QFJ_1_6_x it was a cherry-pick.

PR #62, opened 2016-02-16
Mina-core to 2.0.13

guidomedina:
From https://mina.apache.org/mina-project/index.html

MINA 2.0.13 released posted on February, 16, 2016

Another release to fix a critical SSL bug ( a race condition which could lead to a deadlock in some corner cases).

We urge you to switch to this version if you were using MINA 2.0.12.

Bugs :

DIRMINA-1019 SslHandler flushScheduledEvents race condition
DIRMINA-1027 SSLHandler writes corrupt messages under heavy load

### DIFF
--- a/quickfixj-core/pom.xml
+++ b/quickfixj-core/pom.xml
@@ -59,7 +59,7 @@
 		<dependency>
 			<groupId>org.apache.mina</groupId>
 			<artifactId>mina-core</artifactId>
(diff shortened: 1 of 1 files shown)


guidomedina:
What can I say :(
Mina-core has been releasing like a Yoyo

chrjohn:
Thanks for letting us know :)

PR #63, opened 2016-02-28
QFJ-879: Small Changes: Static compiled pattern, un-sync time and slf4j update

guidomedina:
**Small changes:**

- Synchronization of time in millis removed.
- Regex compiled pattern made static.
- Slf4j updated to 1.7.18

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,7 +69,7 @@
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 		<jdkLevel>1.7</jdkLevel>
-		<slf4j.version>1.7.16</slf4j.version>
+		<slf4j.version>1.7.18</slf4j.version>
 		<mainClass/>
 
         <maven-resources-plugin-version>2.7</maven-resources-plugin-version>
--- a/quickfixj-core/src/main/java/quickfix/SessionID.java
+++ b/quickfixj-core/src/main/java/quickfix/SessionID.java
@@ -38,7 +38,7 @@ import quickfix.field.TargetSubID;
  * but using different FIX versions (and/or session qualifiers).
  */
 public class SessionID implements Serializable {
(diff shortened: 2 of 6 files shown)


chrjohn:
This is for QFJ-879, right?

guidomedina:
Yeah, but decided to also update few minor things in a single commit.

On 29/02/16 08:48, Christoph John wrote:
>
> This is for QFJ-879, right?
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/quickfix-j/quickfixj/pull/63#issuecomment-190106864>.
>


-- 
Squared Financial Services Limited
www.squaredfinancial.com

Switch:       + 353 1 661 0500
Fax:          + 353 1 661 1087
Our LinkedIn: "http://www.linkedin.com/company/squared-financial-services-ltd."


Squared Financial Services Limited is regulated by the Central Bank of Ireland.
Registered Office: Custom House Plaza 2, Harbourmaster Place, IFSC, Dublin 1, Ireland. Registered No. 399289.
Directors: Youssef BARAKAT, Ronnie CULLITON, Philippe GHANEM (Swiss), Jeff GROSSMAN, Andrew MATHERS, Bachir RABBAT (Canadian).

guidomedina:
Yeah, but decided to also update few minor things in a single commit. How do I add a comment to also affect JIRA automatically? is that configured?

chrjohn:
If you include the issue number in the title, JIRA has a link to the pull request on the right hand side. But AFAIK you cannot make github comments appear automatically in JIRA. Or maybe you need a special plugin for that.

guidomedina:
I have changed the synchronize block to create a proper memory barrier (on the write) operation which should clear the L1 cache, because this variable is static the synchronize should be at the class level as it is now (forced push)

guidomedina:
Scratch that, it must be late...the synchronize was on a static method already. Re pushing.

guidomedina:
Re-pushed with volatile, screw it, volatile will keep it correct and thread safe as writes are visible as soon as done, but TBH it is still slower than no volatile at all, though without there is a risk of not seeing the value at all if you change the time source after some other thread have started.

chrjohn:
OK, thank you :)

PR #65, opened 2016-03-26
QFJ-884 Removed quickfixj-messages-all dependency from individual quickfixj-messages-fixXX

guidomedina:
Fixes including all messages dependency into individual messages dependencies.
Made JUnit test version consistent, there is a module with 4.10 and another with 4.12, because ordering behavior was added at 4.11+ the version was set on a centralized place (a maven property) to 4.10


philipwhiuk:
If your 'unit tests' need to be run in a certain order they need re-writing because they aren't unit tests. We should not be downgrading JUnit or not upgrading JUnit (which fixes bugs);

In any case, it's possible to annotate test classes to run in the old order if this causes you an issue: https://github.com/junit-team/junit4/wiki/Test-execution-order

guidomedina:
I haven't written a unit test for QFJ, I'm just made the version consistent, some module has 4.10 and another module 4.12 plus the ordering fact, so I decided to just make it consistent to 4.10, this PR is not fixing that particular problem TBH.

guidomedina:
@philipwhiuk I just changed the PR comment, force pushed and changed the commit message to reflect exactly what I did.

chrjohn:
Hi @guidomedina , thanks for the PR.
I think the JUnit version number can be set to 4.12 for all modules.
I am not so sure about the dependency to quickfixj-messages-all. This has been introduced in #31 . Maybe @manureno  can comment on this. I don't know if there is a cleaner way.

guidomedina:
@chrjohn @manureno

I just force pushed with it at version 4.12, in my project I had before that commit only `quickfixj-messages-fix44` which was pulling `quickfixj-core`, after that commit `quickfixj-messages-all` was also pulled which for some systems might be a bad thing as you have 2 jars with same classes and too many extra classes with no need or reason.

I tried to surgically fix the error since I don't completely understand what was the reason of having `quickfixj-messages-all` as a dependency of every individual message version, then there wouldn't be a reason for them to exist at all.

guidomedina:
@chrjohn also, I ran all the tests with my changes and only SSL tests failed which is happening because some other changes you made before, at least on my environment:

Linux Mint (Ubuntu) 3.19.56 JDK 8u77 64bits.

chrjohn:
@guidomedina the SSL tests fail because on newer JDKs the ciphers used in the tests are disabled due to security reasons.
Unfortunately, there is no automatic unit test which assures that everything works as expected. See https://github.com/quickfix-j/quickfixj/pull/31#issuecomment-193652120

guidomedina:
@chrjohn I don't understand, are you implying that `quickfixj-messages-all` will be needed if you only use `quickfixj-messages-fix44` for example?

What's the point of the modular build and dependency separation that was added?

chrjohn:
I don't think that it was the intention to have it the way it is now. I have only briefly checked that the separate message JARs look correct.
But you are right that from a Maven dependency point of view it is very undesirable to pull the messages-all dependency if you only declare the fix44 dependency.

manureno:
Hi,

Dependancy to message-all has been introduced in QFJ-855 so that all
message source files are merged and then compiled before they are JARed in
order, to reproduce the packaging behavior that existed in the previous ant
build.
Merging the source of all messages before compilation is required since
some FIX messages type differ from one FIX spec to another : messagexx.jar
are thus not inter-changeable if built independantly.

Although this dependancy is not a maven best practice, the jar themselves
should be OK thanks to the following block in the messagexx/pom.xml
(example for fix44):
{code}
<plugin>
                <artifactId>maven-jar-plugin</artifactId>
                <groupId>org.apache.maven.plugins</groupId>
                <version>2.6</version>
                <configuration>

<classesDirectory>${project.basedir}/../quickfixj-messages-all/target/classes/</classesDirectory>
                    <includes>
                        <include>FIX44.xml</include>
                        <include>FIX44.modified.xml</include>
                        <include>quickfix/fix44/**</include>
                        <include>quickfix/field/**</include>
                    </includes>
                    <archive>

<manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
                    </archive>
                </configuration>
            </plugin>
{code}

Looking again a this, a dependancy to the core module may nevertheless
replace the dependancy to message-all if the <classesDirectory> of the jar
plugin configuration is also changed to point to   :
{code}
<classesDirectory>${project.basedir}/../../quickfixj-core/target/classes/</classesDirectory>
{code}
The code generation may then also be removed from the message-all, thus
ensuring that messages are compiled only once and this would also save
build time.

guidomedina:
@manureno I'm not 100% understanding what needs to be done, is it possible for you to pull my master and modify it? the current solution fixes the JAR problem but it doesn't eliminate the extra time consuming steps you are talking about.

Maybe @chrjohn can merge this and then you can modify further? whatever it is easier which either way will fix the problem, as I did other good things to the poms anyway.

guidomedina:
@chrjohn WDYT about:
https://github.com/quickfix-j/quickfixj/pull/65/commits/18b5b710b8259d78576704a212deeb1c90c48c3c

It looks awesome to me, so many if-statements eliminated, the reasoning for this:
http://stackoverflow.com/questions/22110707/how-is-string-in-switch-statement-more-efficient-than-corresponding-if-else-stat

**Note:** This is only good for Java 7 though.

guidomedina:
@chrjohn for FIX44 alone there are 91 message types, that's some if statements it has to go through, basically making the switch be O(1) instead of O(n) for income message.

manureno:
Hi Guido

Sorry, although I've pulled your branch, I still didn't have time to go
through a proper pull request.
If you need to merge before I complete it, please beware that the
individual fix-messages jars modules use code generated and compiled by the
message-all module.
This means that maven must run the message-all before the individual
message modules, hence the dependancy declaration.
If the dependancy is removed, it is not guaranteed that the classes that
need to be jar-ed in the individual modules are actually generated and
compiled before.
Hopes it clarifies...

Best regards,
Manu

guidomedina:
Hi @manureno 

I'm using my own modified version for fix44, and like I explained before it was pulling only fix44 and core but now is pulling fix44, all messages and core which is wrong, thanks for trying to fix it, I would at done it if I would known how.

I would recommend you to base your work on top of my master which has two nice commits, then send a PR with a 3rd commit (yours) in it.

Best regards,

Guido.

guidomedina:
@chrjohn I think this branch has the expected results, I will check it tomorrow but you can go ahead and check it too, @manureno added a commit to it.

guidomedina:
@chrjohn @manureno when did you guys eliminated the possibility of generating decimal vs double fields? was that intended? such functionality has no effect now but I'm not sure when it was removed.

guidomedina:
@chrjohn @manureno maybe it is fixed, I will check later.

guidomedina:
@manureno I just tried `mvn clean install` and at the distribution phase it gets stuck:

[INFO] --- maven-javadoc-plugin:2.10.3:javadoc (javadoc-jar) @ quickfixj-distribution ---

On my Linux PC with 16GB RAM, i5-4670k, SSDs it stays there for more than 5min with 100% CPUs usage, something is wrong there.

chrjohn:
@guidomedina This is where it creates the JavaDocs which is quite a time-consuming process given the thousands of classes for messages, fields, etc.
We could think about leaving that classes out of the JavaDoc generation. But at the moment the generation is still done.

guidomedina:
@chrjohn I know the JavaDocs take a while but after last commit it seems it doesn't finish for me, did you got past it?

chrjohn:
@guidomedina I will check. 
On another note: why did you remove the SupportForMultipleFixVersionsTest??

manureno:
Hi,

The test was removed because not needed anymore.

Before, there was several FIX message generation and compilation executed
between the core and the messageXX modules : this could lead to
inconsistencies ending up in a "VerifyError". The test was there to check
that all these artifacts were consistent.

Now, there is only one FIX message generation and compilation (replicating
what was done in the previous ANT build BTW). The classes are now supposed
to be natively consistent and  the test is no needed any more.

Manureno

chrjohn:
Thanks for the explanation, @manureno 
@guidomedina : you are right, the javadoc generation seems to take forever. I assume it is because the memory for the javadoc generation process is too low and all the time is spent doing GC. But I don't know why it is this way now. Actually, it can only have to do with the changes in this PR but unfortunately I can't explain why. There shouldn't be more classes than before, should it?

guidomedina:
@chrjohn there shouldn't be more classes indeed, remember this PR has 3 commits, with the first 2 it seems to do it just fine, when you add the 3rd commit from @manureno the JavaDoc phase takes forever (or never finishes)

My commits are:
- Remove all-messages dependencies (it seems to fix the problem for me)
- Optimize `MessageFactory` for each FIX version (switch for several string cases is faster than if due to byte code generating a hash table for look ups)

The last commit was from @manureno to fix the code generation but added the JavaDoc issue we have now.

chrjohn:
Hi,
I "fixed" it by making the following configuration for the javadoc plugin (line 158 in pom.xml):

```
                    <configuration>
                        <excludePackageNames>quickfix.fix*</excludePackageNames>
                    </configuration>
```
IMHO it makes no sense anyway to generate javadoc for the generated field classes.

guidomedina:
@chrjohn which POM? I will add it to the last commit and test.

chrjohn:
Sorry, the main POM in the base directory.

guidomedina:
Tested it like this and it worked, please check if this is OK (notice there are 2 configuration sections, one for the plugin and another for an execution goal):

```
                <plugin>
                    <groupId>org.apache.maven.plugins</groupId>
                    <artifactId>maven-javadoc-plugin</artifactId>
                    <version>${maven-javadoc-plugin-version}</version>
                    <configuration>
                        <excludePackageNames>quickfix.fix*</excludePackageNames>
                    </configuration>
                    <executions>
                        <execution>
                            <id>attach-javadocs</id>
                            <goals>
                                <goal>jar</goal>
                            </goals>
                            <configuration>
                                <additionalparam>-Xdoclint:none</additionalparam>
                            </configuration>
                        </execution>
                    </executions>
                </plugin>
```

guidomedina:
I just forced push the last @manureno commit with this change, I think this PR is now good.

chrjohn:
@guidomedina re pom.xml: yes, that's how I tested it here.

I will have to check how I can selectively merge this to 1.6.x later. The packaging changes should also go into 1.6.x but the switch-case-with-string does only work for the 1.7.x.

guidomedina:
Yes, switch-strings will only work on JDK 7+ which will make a huge difference for message factories as all of them have at least 50+ string cases, you can cherry-pick and skip MessageFactory.xsl commit.

guidomedina:
@chrjohn wait for one more commit, I'm eliminating all deprecated APIs, have 1 left out of 50+ now.

guidomedina:
@chrjohn now it should be good to go ;-)

chrjohn:
@guidomedina great, thanks. :)
Thanks also to @manureno for your help along the way. :)

chrjohn:
I have re-added the PermGen settings which were removed. We are still compiling on Java7 on the build server and need that setting.

guidomedina:
Check the place it was removed from in the following commit https://github.com/quickfix-j/quickfixj/pull/65/commits/b4f6847569cc3a8d0715a835d6bb6d93928fb9ee

quickfixj-messages/quickfixj-messages-all/pom.xml
```diff
@@ -59,140 +52,76 @@
...
 				</executions>
 			</plugin>
 			
 			<plugin>
-				<artifactId>maven-surefire-plugin</artifactId>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-jar-plugin</artifactId>
+				<version>${maven-jar-plugin-version}</version>
 				<configuration>
-					<argLine>-Xmx512m -XX:MaxPermSize=128m</argLine>
```
guidomedina: @chrjohn I think you got it right, see here.

guidomedina: @chrjohn no, it is at that line but the plugin should be surefire-plugin not jar-plugin.

guidomedina: @chrjohn 
**Edit:** it was removed from that line but the plugin should be `maven-surefire-plugin` not `maven-jar-plugin`

chrjohn: @guidomedina I have increased the PermGen size for the surefire plugin now.
ce69c6dec2c1b009f8f40804cb59a71cbdb8600a

chrjohn:
OK, now we have a bunch of failing tests. Will have to check where they come from. 
http://www.quickfixj.org:8085/download/QFJ-GIT-JOB1/build_logs/QFJ-GIT-JOB1-133.log
On your system they are all successful?

guidomedina:
On my system on SSL tests failed (around 12 SSL tests)

chrjohn:
Hmm, strange. I think I know where the problem is. Some tests do not find their config files anymore. E.g. the test JdbcLogTest. Example:
```
JdbcTestSupport.loadSQL(connection,
                    "config/sql/hsqldb/messages_log_table.sql",
                    new JdbcTestSupport.HypersonicPreprocessor(null));
```
This is now failing. When I change it to:

```
JdbcTestSupport.loadSQL(connection,
                    "src/main/resources/config/sql/hsqldb/messages_log_table.sql",
                    new JdbcTestSupport.HypersonicPreprocessor(null));
```

the tests succeeds. Seems to have to do with the changed "resources" tree in quickfixj-core/pom.xml

guidomedina:
It all has to do with @manureno commit, you can try without that commit and see that happens, it worked for me without it, by just removing messages-all dep.

chrjohn:
OK, I will check later to untwine this

manureno:
Hi,

I'll try to check the impact of the change of the resource block on the
test run. The issue is that the tests hardly all pass on my machine
(especially the ssl ones) so it's a bit hard to diagnose.
Meanwhile beware that merging the QFJ-884 without the last commit may cause
other issues : in this verison it is not guaranteed that the block that
generates messages is run before the code that packages the messageXX jars.

guidomedina:
We just need to figure out the minor detail, as usual I'm sure is something stupid that is breaking it, these bugs that you spend 3 days looking for them and then boom, you find the 1 liner ;-)

chrjohn:
@manureno Regarding the SSL tests: I assume you are using JDK8 (or a recent JDK7 version). There have been some ciphers disabled on those versions. That is why the tests are failing. Will update the ciphers used in these tests when I find the time.
For the time being, if you want the SSL related tests to work then you can create a file with the following content:

```
jdk.certpath.disabledAlgorithms=true
jdk.tls.disabledAlgorithms=true
jdk.tls.legacyAlgorithms=true

```

And put the following option to the command line when building:
` -Djava.security.properties=<path to created file>`

For example:

`mvn clean package -Djava.security.properties=<path to created file>`

If the tests are now still failing then it is most likely due to the fact that the tests are not finding the certificate. This is the same problem as with the JdbcLogTest mentioned earlier. See https://github.com/quickfix-j/quickfixj/pull/65#issuecomment-212921833

Cheers,
Chris.

guidomedina:
Testing with your suggested file created (JDK 8u92 here), got JDBC tests failures, I suspect it might be JUnit v4.12, testing again with Junit v4.10

manureno:
@chris thanks for the tip about SSL, I'll try that ASAP

guidomedina:
I did the following:

`git checkout 18b5b710b8259d78576704a212deeb1c90c48c3c` checkout up to commit before big pom changes.
`git cherry-pick a31a648dcbef0a6168497d48ab9de851fca91b57` add deprecated commit fixes
`mvn clean test -Djava.security.properties=/home/gmedina/cert.txt`

And only one test failed:
```
Results :

Failed tests: 
  SSLCertificateTest.shouldFailWhenUsingBadClientCertificate:464 No SSL exception thrown

Tests run: 1338, Failures: 1, Errors: 0, Skipped: 0
```

That tells me that @chrjohn is right about config not being found by the tests with the pom changes.

chrjohn:
I found the problem with some of the failing tests. In the quickfixj-core pom.xml it should read:
```
            <testResource>
                <directory>src/main/resources</directory>
            </testResource>
```

Before, `directory` only contained `src/main` and so the resources were not found and hence not copied to the test-classes directory.
Now the sql files and the quickfixj.cert are copied.

What doesn't seem to be required is the whole `resources` tree in the quickfixj-core POM. At least I am not sure for what it is required.

The SSL tests don't seem to find their certificate, tough. Which is strange since it is located in the same directory as the SQL files.

Cheers

chrjohn:
Umm, guys... The last build with https://github.com/quickfix-j/quickfixj/commit/3bc3562a3cf7c1a5fb79926388f81f4c8c8c3ad1 was successful:
http://www.quickfixj.org:8085/browse/QFJ-GIT-134

Strange thing is: on my local system I still get all sorts of failures for the SSL tests (although I have the changed security file in place).
Maybe another JDK8/7 change. Local JDK is 1.8.0_77 on build server there is 1.7.0_45.

guidomedina:
My tests all were successful with your new commit, this is what I executed:

`mvn clean test -Djava.security.properties=/home/gmedina/cert.txt`

chrjohn:
Doesn't work for me. Very strange. At least it does work for you and on the build server. ;)

guidomedina:
Read release notes JDK 8u92, they did something with the encryption algorithms.

chrjohn:
From what I understood they disabled more algorithms, but since I set `jdk.tls.disabledAlgorithms=true` none of them should actually be disabled...

PR #71, opened 2016-05-22
Codec cleanup

amichair:
Split into 3 commits to make it a bit easier to follow. Highlights include getting rid of the BufPos class (lots of unnecessary object allocations), extracting the pattern matching code into a separate class, replacing Character calls with literals (since it deals with bytes, not chars, so there's no need for the overhead) a bit of renaming and little refactoring for clarity.

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/mina/message/FIXMessageEncoder.java
+++ b/quickfixj-core/src/main/java/quickfix/mina/message/FIXMessageEncoder.java
@@ -20,7 +20,7 @@
 package quickfix.mina.message;
 
 import java.io.UnsupportedEncodingException;
-import java.util.Collections;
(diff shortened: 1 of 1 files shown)


chrjohn:
Thank you for the PR.
This has some overlaps with #47 in FixMessageEncoder. Will have to check how this can be integrated.

amichair:
There's actually not much overlap - this PR mostly modifies the decoder, and the other PR only the encoder. The encoder changes here are just in defining the constant type set (no conflict there) and extracting a toBytes() utility method which can be reverted or ignored since it is not of much consequence.

amichair:
I split the last commit further, so now the encoder changes are on a separate commit, in case that helps.

PR #74, opened 2016-06-06
FieldType internal cache class for fast type lookup by name.

guidomedina:
FieldType internal cache class for fast type lookup by name.

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/FileStore.java
+++ b/quickfixj-core/src/main/java/quickfix/FileStore.java
@@ -24,11 +24,11 @@ import java.io.BufferedOutputStream;
 import java.io.Closeable;
 import java.io.DataInputStream;
 import java.io.DataOutputStream;
+import java.io.EOFException;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
-import java.io.OutputStream;
 import java.io.RandomAccessFile;
 import java.util.Calendar;
 import java.util.Collection;
@@ -77,11 +77,7 @@ public class FileStore implements MessageStore, Closeable {
         this.syncWrites = syncWrites;
         this.maxCachedMsgs = maxCachedMsgs;
 
-        if (maxCachedMsgs > 0) {
-            messageIndex = new TreeMap<Long, long[]>();
-        } else {
-            messageIndex = null;
-        }
+        messageIndex = maxCachedMsgs > 0 ? new TreeMap<Long, long[]>() : null;
 
         final String fullPath = new File(path == null ? "." : path).getAbsolutePath();
         final String sessionName = FileUtil.sessionIdFileName(sessionID);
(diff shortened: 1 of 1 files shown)


guidomedina:
@amichair WDYT?

amichair:
It appears that the implicit `Enum.valueOf(String)` bytecode just calls `Enum.valueOf(Class, String)` which in turn does a HashMap lookup (in Oracle JDK 6/7/8 at least), so I'm not sure what the advantage of the explicit cache is - care to elaborate?

guidomedina:
@amichair answering two concerns, the first one not related:
-`switch(string)` are turned into `switch(integer)` using the `String.hashcode()` which Java internally treats as a sort of hashmap lookup.
-`Enum.valueOf(Class, String)` uses a hashmap to locate which Enum class, but the string lookup is done via loop, the fact that it has to do a class lookup first makes even worse than doing a hashmap lookup and at best makes it a bit slower than doing it implicitly.

guidomedina:
@amichair there is no 20+ rule, it is just a limit to determine if a hashmap is going to payoff or just a loop lookup, say if you have 5 values, there is no point making a hashmap, usually people create small static caches if the values you are looking through are more than 15 or 20, in this case there are 28 values which justifies having a static map.

amichair:
@guidomedina I'm not sure I follow. `Enum.valueOf(Class, String)` receives a Class instance - it doesn't need to look it up. If you are referring to how `valueOf(String)` finds the class, then the class parameter is taken from a class constant - this code is generated within the enum subclass which already knows its type at compile time. Given the class, the two-argument method is invoked, and it finds the enum constant from its name using a `HashMap` that maps names to constants. I'm not sure which loop you're referring to. I have a feeling we may be talking about two different things :-)

As for the hashcode, you are correct - I just wanted to point out that there is no actual HashMap instance involved, just a native internal table lookup based on the hashCode().

guidomedina:
@amichair assuming the `Enum.valueOf(String)` is a `HashMap` lookup, the other advantage of my method is that you completely avoid the exception when not found and stack trace creation.

But still most people say `Enum.valueOf(String)` is just a dumb loop until found:
http://stackoverflow.com/a/7165270/1666753

guidomedina:
Maybe they changed it and I'm reading the wrong responses all over, I guess for the exception is not worth the PR, I might close it depending on more feedback: http://bugs.java.com/bugdatabase/view_bug.do?bug_id=6352179

amichair:
@guidomedina Thanks for the links, now I understand your context better. Here's what I understand from them:

- The JDK bug database link and the duplicate issue it links to show that there indeed used to be an issue of using a loop to find the enum constant, but it was fixed in 2004 and released in JDK 6, and someone else reported this in 2005 and was told that it's already been fixed. So this is no longer relevant to any supported JDK, and it's all using HashMaps for the past decade - no need to worry about it.

- In the StackOverflow post, the references to loops are basically just about the code snippet from the question, unrelated to the JDK and quickfixj. My own (possibly biased) takeaway from the various responses and comments is that there is no practical performance difference between catching an exception or not, and it's best to opt for readability, simplicity and maintainability (i.e. just use valueOf) - which I would totally agree with even if there was a slight difference in performance, since the FileType code under discussion here is in no way a performance bottleneck, and would fall under the 'premature optimization' idiom.

btw, I'm actually thinking of changing the existing behavior and just removing the catch block and letting it throw the exception to the caller. I think this isn't supposed to be happening anyway under normal circumstances, so we can keep the code even simpler. What do you think?

In addition, I'll take the opportunity to say thanks for the code review of this commit - I hope you have time to go over the others (and I hope you approve :-) ) - the more eyes on the code the better!

guidomedina:
I think is OK, I usually cache them afraid of JDK folks not doing the best specially on this industry. I ttry to give my input as much as I can, btw, nice tidy up, the Enum conversion was very good, simplifies a lot of things.

But for decision making between catching the exception or let it be thrown will be up to @chrjohn as it is a behavioral change.

PR #75, opened 2016-06-10
QFJ-318 / QFJ-677 regression (FileLog thread safety)

knutae:
Fixes thread race condition regression in http://www.quickfixj.org/jira/browse/QFJ-318. 1.4.0 seems to be the last released version with this fix.

See also: https://sourceforge.net/p/quickfixj/code/826/

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/FileLog.java
+++ b/quickfixj-core/src/main/java/quickfix/FileLog.java
@@ -90,16 +90,18 @@ public class FileLog extends AbstractLog {
         writeMessage(messages, message, false);
     }
 
-    private synchronized void writeMessage(FileOutputStream stream, String message, boolean forceTimestamp) {
+    private void writeMessage(FileOutputStream stream, String message, boolean forceTimestamp) {
         try {
-            if (forceTimestamp || includeTimestampForMessages) {
-                writeTimeStamp(stream);
-            }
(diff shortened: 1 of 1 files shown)


chrjohn:
I don't know if we should synchronize the whole writeMessage() method since that will also block concurrent access to the event log. What about introducing a lock for writing to the message log and for writing to the event log?

knutae:
I agree that synchronizing the message and event log separately sounds better than synchronizing the whole method. Maybe just use the FileOutputStream argument as a lock?

chrjohn:
I thought a little more about it. Actually, the FileOutputStreams `messages` and `events` which are passed into the method are not final. So it could be possible that content of the variables `messages` and `events` change concurrently and it is hence not guaranteed that the synchronized block synchronizes on the same object.

knutae:
Sorry about the delay. Introduced separate objects for locking instead of synchronizing on the FileOutputStream arguments.

guidomedina:
Another _-but more complicated-_ way would be to have 1 consumer thread and a bounded `BlockingQueue`, that way no lock is needed and the publishers won't be blocked, it also adds some back pressure by blocking the publisher when the queue is full _-for slow writes-_

Maybe this is something that can be implemented at a higher level for all message writers, including the ones using distributed data.

PR #76, opened 2016-06-16
Fixes QFJ-895 for 1.7.x and bump plugin versions

guidomedina:
Bump plugin versions, all tests passed, everything generated pretty fast which makes me think `javadoc-plugin` got better, or maybe just a placebo effect.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -73,13 +73,13 @@
 		<junit.version>4.12</junit.version>
 		<mainClass/>
 
-		<maven-resources-plugin-version>2.7</maven-resources-plugin-version>
+		<maven-resources-plugin-version>3.0.1</maven-resources-plugin-version>
 		<maven-compiler-plugin-version>3.5.1</maven-compiler-plugin-version>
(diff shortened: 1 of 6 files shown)


guidomedina:
@amichair can you please take a look at:

http://www.quickfixj.org/jira/browse/QFJ-895

I was able to patch 4 out of 5 files with https://github.com/quickfix-j/quickfixj/pull/76/commits/b0b35b54ffa43fb03225a84f16cbeba72fc30cf9 if you could shed some light as `IoSessionInitiator` has many conflicts after SSL refactor which I couldn't crack.

guidomedina:
@chrjohn @amichair I'll retry tomorrow, this is what happens when you try to code after work, I just noticed I can do it _-I think-_ but will do it tomorrow after lunch that I'm home early.

guidomedina:
@chrjohn @amichair done, it would be nice if the original author of the patched classes could verify this commit.

chrjohn:
Thanks Guido! :)

PR #77, opened 2016-06-23
Fixes QFJ-895 for 1.6.x

guidomedina:
Patched classes provided with ticket that fixes #895 for 1.6.x

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/ThreadedSocketInitiator.java
+++ b/quickfixj-core/src/main/java/quickfix/ThreadedSocketInitiator.java
@@ -81,8 +81,8 @@ public class ThreadedSocketInitiator extends AbstractSocketInitiator {
     }
 
     public void stop(boolean forceDisconnect) {
+        stopInitiators();
         logoutAllSessions(forceDisconnect);
-        stopSessionTimer();
         if (!forceDisconnect) {
             waitForLogout();
         }
--- a/quickfixj-core/src/main/java/quickfix/mina/AbstractIoHandler.java
+++ b/quickfixj-core/src/main/java/quickfix/mina/AbstractIoHandler.java
@@ -19,23 +19,18 @@
 
 package quickfix.mina;
 
-import static quickfix.MessageUtils.parse;
-
-import java.io.IOException;
-
 import org.apache.mina.core.service.IoHandlerAdapter;
 import org.apache.mina.core.session.IoSession;
 import org.apache.mina.filter.codec.ProtocolCodecException;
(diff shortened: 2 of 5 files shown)


chrjohn:
Hi @guidomedina , after merging other SSL-related changes I had to resolve some conflicts manually.
For 1.6.x there is one failure (at least on my system):

`SSLCertificateTest.shouldFailIndividualSessionsWhenInvalidCertificatesUsed:245 No SSL exception thrown`

Does this also occur for you?
Thanks,
Chris.

guidomedina:
There were two PRs because it was a big difficult for 1.7 and I was able to keep it very clean, try to compare the IoSessionInitiator between branches after your conflicts resolved maybe you missed something.

guidomedina:
Testing now on 1.7.x, then I will compare both classes between branches maybe I see something different.

chrjohn:
I have compared them but found no changes (except an unrelated change getHostName() vs. getHostString().
Probably has to do with the newly generated SSL certificates (https://github.com/quickfix-j/quickfixj/commit/508244b8c44cbf398fe6a8cff5eec78fa8da7cb2) but it is beyond me why it does not work on 1.6.x.

guidomedina:
Could be this line an exception being transformed into another exception because this line https://github.com/quickfix-j/quickfixj/blob/master/quickfixj-core/src/main/java/quickfix/mina/initiator/IoSessionInitiator.java#L163 was before outside the try, testing that change.

guidomedina:
ok, that was it for 1.7.x, now I need to understand if I need to catch that exception or not basically from:

```
        private void connect() {
            try {
                lastReconnectAttemptTime = SystemTime.currentTimeMillis();
                SocketAddress nextSocketAddress = getNextSocketAddress();
                if (localAddress == null) {
                    connectFuture = ioConnector.connect(nextSocketAddress);
                } else {
                    // QFJ-482
                    connectFuture = ioConnector.connect(nextSocketAddress, localAddress);
                }
                pollConnectFuture();
            } catch (Throwable e) {
                handleConnectException(e);
            }
        }
```
to:

```
        private void connect() {
            lastReconnectAttemptTime = SystemTime.currentTimeMillis();
            SocketAddress nextSocketAddress = getNextSocketAddress();
            try {
                if (localAddress == null) {
                    connectFuture = ioConnector.connect(nextSocketAddress);
                } else {
                    // QFJ-482
                    connectFuture = ioConnector.connect(nextSocketAddress, localAddress);
                }
                pollConnectFuture();
            } catch (Throwable e) {
                handleConnectException(e);
            }
        }
```

WDYT? should an exception outside the try screw up what's been fixed by it?

PR #78, opened 2016-06-24
Fixes QFJ-895 SSL exception caused by patch.

guidomedina:
@chrjohn this fixed the SSL exception for 1.7.x, see if it also fixes it for 1.6.x, it would be nice if the author of the patch could come by and comment ;-)

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/mina/initiator/IoSessionInitiator.java
+++ b/quickfixj-core/src/main/java/quickfix/mina/initiator/IoSessionInitiator.java
@@ -158,9 +158,9 @@ public class IoSessionInitiator {
         }
 
         private void connect() {
+            lastReconnectAttemptTime = SystemTime.currentTimeMillis();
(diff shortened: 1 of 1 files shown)


guidomedina:
Testing for 1.6.x after cherry-picked

guidomedina:
@chrjohn yeap, same for 1.6.x, all tests passed too, I don't think leaving these two lines out of the try should affect the patch, but it is definitely fixing the SSL tests so I guess that's the correct change.

guidomedina:
You only need to cherry pick this PR into 1.6.x, it merges without conflicts.

guidomedina:
I like how it takes almost 2 less minutes testing on 1.7.x than 1.6.x, I think that's because of the `MessageDecoder` using `switch(String)` with 50+ cases (message type)

guidomedina:
Closing, that test fails sometimes, it just failed for me again, then next time it didn't.

chrjohn:
Yes, same for me. Sometimes it is failing. Strange.

PR #80, opened 2016-09-07
Upgraded MINA core to 2.0.15 and some plugins.

guidomedina:
Bump some plugins and dependencies, specially mina-core to 2.0.14:
https://mina.apache.org/mina-project/index.html

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/mina/SingleThreadedEventHandlingStrategy.java
+++ b/quickfixj-core/src/main/java/quickfix/mina/SingleThreadedEventHandlingStrategy.java
@@ -1,4 +1,5 @@
-/*******************************************************************************
+/*
+ ******************************************************************************
  * Copyright (c) quickfixengine.org  All rights reserved.
  *
  * This file is part of the QuickFIX FIX Engine
@@ -15,19 +16,18 @@
  *
  * Contact ask@quickfixengine.org if any conditions of this licensing
  * are not clear to you.
- ******************************************************************************/
+ ******************************************************************************
+ */
 
 package quickfix.mina;
 
+import quickfix.*;
+
+import java.util.ArrayList;
+import java.util.List;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
-
-import quickfix.LogUtil;
-import quickfix.Message;
-import quickfix.Session;
-import quickfix.SessionID;
-import quickfix.SystemTime;
 
 /**
  * Processes messages for all sessions in a single thread.
@@ -68,9 +68,9 @@ public class SingleThreadedEventHandlingStrategy implements EventHandlingStrateg
             synchronized (this) {
                 if (isStopped) {
                     if (!eventQueue.isEmpty()) {
-                        final LinkedBlockingQueue<SessionMessageEvent> tempQueue = new LinkedBlockingQueue<SessionMessageEvent>();
-                        eventQueue.drainTo(tempQueue);
-                        for (SessionMessageEvent event : tempQueue) {
+                        final List<SessionMessageEvent> tempList = new ArrayList<SessionMessageEvent>();
+                        eventQueue.drainTo(tempList);
+                        for (SessionMessageEvent event : tempList) {
                             event.processMessage();
                         }
                     }
--- a/quickfixj-core/src/main/java/quickfix/mina/ThreadPerSessionEventHandlingStrategy.java
+++ b/quickfixj-core/src/main/java/quickfix/mina/ThreadPerSessionEventHandlingStrategy.java
@@ -19,19 +19,13 @@
 
 package quickfix.mina;
 
-import java.util.Collection;
-import java.util.Iterator;
-import java.util.concurrent.BlockingQueue;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.LinkedBlockingQueue;
-import java.util.concurrent.TimeUnit;
-
 import quickfix.LogUtil;
 import quickfix.Message;
 import quickfix.Session;
 import quickfix.SessionID;
-import static quickfix.mina.EventHandlingStrategy.END_OF_STREAM;
+
+import java.util.*;
+import java.util.concurrent.*;
 
 /**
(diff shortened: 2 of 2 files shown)


guidomedina:
Some tests hitting deadlocks because of proxool, try on your side without merging first and let me know if is happening to you.

guidomedina:
@RockManJoe64 can you take a look at the tests? they seem to be blocking due to proxool changes, I'm not 100%, I can't figure out what changed specifically because of too many files formatted, difficult to diff if you format everything.

guidomedina:
@chrjohn @RockManJoe64 ignore this PR for the moment, the lock seems to be related with the latest mina-core, still testing...I will let you know what I conclude, tests take a while to run so nothing more for today.

chrjohn:
@guidomedina I am also facing this after update to MINA 2.0.14. Maybe it is related to https://github.com/quickfix-j/quickfixj/commit/3e6eebf2d73aad6cfd38f8954d487af1ad65b1a2 (dispose(true) has been added when closing a SocketAcceptor)
But this is only a wild guess. However, https://mina.apache.org/mina-project/apidocs/org/apache/mina/core/service/IoService.html#dispose mentions that calling dispose(true) might lead to a deadlock when called from an IoFutureListener. But AFAIK we do not call it from there.

chrjohn:
I've sent a message to the MINA mailing list. Maybe they will come back with a pointer. Actually, the "dispose(true)" change does not seem to make a difference.

guidomedina:
Yeah, that makes sense, maybe they find a bug and release 2.0.15, tyvm for the efforts.

chrjohn:
No problem, thanks for the PR. :)

guidomedina:
Testing with Mina 2.0.15 which has been released.

guidomedina:
still threads blocking....

chrjohn:
I know. Emmanuel from the MINA team will probably look at it next week since 2.0.15 included a priority SSL issue.
About the blocking: IMHO it does not really "block". The thing I observed is that the tests are waiting for the Logout message which is not received. The counterparty seems to send it but the receiving party does not receive it. Maybe it has to do with either closing the connection too early or not flushing it.

guidomedina:
So maybe that is related with this change? which maybe we will have to change the way we close the connection or the tests?

https://issues.apache.org/jira/browse/DIRMINA-1025

chrjohn:
Yes, that's what I thought also. See mail thread here (already posted some comments above):
http://apache-mina.10907.n7.nabble.com/observing-hangs-after-update-to-MINA-2-0-14-tc51136.html
But didn't have the time yet to play around with the tests. I only ran the tests a few times to see where it is hanging.

chrjohn:
Hi @guidomedina , I think I have found the problem. When closing the connection in IoSessionResponder we are doing a closeNow().
https://github.com/quickfix-j/quickfixj/blob/e2a0fc0636b70d2f735815c6773b5f8dbb993d16/quickfixj-core/src/main/java/quickfix/mina/IoSessionResponder.java#L86
So, as said, it is most likely to be related to https://issues.apache.org/jira/browse/DIRMINA-1025 when closeNow() did allow some messages to still pass through whereas it is really closing the connection now.
Replacing the call by closeOnFlush() should fix it. I am not quite sure whether I should wait for the returned CloseFuture to complete since it might cause deadlocks. Will ask on the MINA dev list.
Cheers,
Chris.

guidomedina:
@chrjohn the `CloseFuture` has always been there so I think what you are suggesting will fix as it will go back to the old _-broken-_ closeNow behavior, we have never waited for the `Future` so I don't think we should now.

guidomedina:
Or, we could close now and wait for that future, but then `await()` with no timeout is not two questions:
- What property do we use for timeout?
- Is it visible in all the cases where is used?

The 2nd can prove to be problematic, I'm testing now with `closeOnFlush()` which is the old _-broken-_ MINA behavior.

guidomedina:
I notice MINA uses `LinkedBlockingQueue` or is it QuickFixJ? I wonder if we can migrate such Java queue so a low GC MpSc queue like https://github.com/JCTools/JCTools/blob/master/jctools-core/src/main/java/org/jctools/queues/MpscChunkedArrayQueue.java

Is it multiple producer single consumer contract? Think about it, we can easily shade JCtools in MINA and have a low latency (because of low GC) product, for things like price streaming, etc.

I'm using JCtools for Akka but unfortunately GC for third party libraries like MINA are out of my hands.

guidomedina:
Actually I see QuickFixJ uses `LinkedBlockingQueue` for events so that should be an easy change for me to replace them with that suggested queue.

I don't think we need to shade that dependency as it is a useful for other people to notice, they will realize there are LMAX like queues available for their projects.

WDTY? shaded or dependency?

guidomedina:
@chrjohn I just downloaded MINA core and I believe they can benefit from higher performance/low GC queues too, they are using by default `ConcurrentLinkedQueue`, since you have a relationship with these guys, it can be of all benefit to propose such idea, WDTY?

btw, tested everything with `closeOnFlush()` and now working with MINA 2.0.15, repushing.

guidomedina:
Follow up on this one:
http://mina.markmail.org/search/?q=#query:+page:1+mid:mcdxlufpcb6ldxkj+state:results

I don't think it applies much for QuickFixJ because of the blocking nature, not the same for MINA because they have a non-blocking behavior.

guidomedina:
@chrjohn I have to redo this branch as you made changes that conflict with mine....I had done it for all closeNow() calls and you only took one...

chrjohn:
@guidomedina I only changed what's necessary. Have to review the other occurrences in the code. Sometimes it is feasible to use closeNow(), e.g. in situations where the counterparty session is unknown, or there is duplicate session trying to log on.

guidomedina:
yes, but remember it was using the old behavior where it would potentially block the thread, by changing to closeOnFlush() it would go back to the old behavior as it is the safest thing we can do at the moment, all tests passed for me.

If later you test the new behavior fully then you can change it, but now I wouldn't recommend it to do it as is not fully tested.

chrjohn:
There were no threads blocked per se. The only thing that was blocking was the acceptance test suite waiting for a Logout message which never came (because it was not flushed). I don't think this will have any relevance in production.
I will do some more testing. Could you please do me a favour and split the POM changes to a separate PR?
Thanks in advance

PR #82, opened 2016-10-21
QFJ-906 Source code language level set to JDK 7.

guidomedina:
Source code language level set JDK 7, can be set to JDK 8 later but first we need to keep it at JDK 7 for easy support of both 1.7.x and 1.6.x

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/Initiator.java
+++ b/quickfixj-core/src/main/java/quickfix/Initiator.java
@@ -30,12 +30,12 @@ public interface Initiator extends Connector {
      *
      * @see quickfix.SessionFactory#SETTING_CONNECTION_TYPE
      */
-    public static final String SETTING_RECONNECT_INTERVAL = "ReconnectInterval";
+    String SETTING_RECONNECT_INTERVAL = "ReconnectInterval";
 
     /**
      * Initiator setting for connection protocol (defaults to "tcp").
      */
-    public static final String SETTING_SOCKET_CONNECT_PROTOCOL = "SocketConnectProtocol";
+    String SETTING_SOCKET_CONNECT_PROTOCOL = "SocketConnectProtocol";
 
     /**
      * Initiator setting for connection host. Only valid when session connection
@@ -43,7 +43,7 @@ public interface Initiator extends Connector {
      *
      * @see quickfix.SessionFactory#SETTING_CONNECTION_TYPE
      */
-    public static final String SETTING_SOCKET_CONNECT_HOST = "SocketConnectHost";
+    String SETTING_SOCKET_CONNECT_HOST = "SocketConnectHost";
 
     /**
      * Initiator setting for connection port. Only valid when session connection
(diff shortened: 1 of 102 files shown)


guidomedina:
I will give you another branch sometime next week or later with JDK 8 optimizations, like using `computeIfAbsent/Present ` for Maps with a loading function, it is much better for caching and stuff with new JDK 8 concurrent changes.

guidomedina:
Added a dom4j change to new version 2.0.0 of this year, under new Maven group:
http://search.maven.org/#artifactdetails%7Corg.dom4j%7Cdom4j%7C2.0.0%7Cjar

quickfixj-core/src/main/java/quickfix/Initiator.java
```diff
@@ -30,42 +30,42 @@
      *
      * @see quickfix.SessionFactory#SETTING_CONNECTION_TYPE
      */
-    public static final String SETTING_RECONNECT_INTERVAL = "ReconnectInterval";
+    String SETTING_RECONNECT_INTERVAL = "ReconnectInterval";
```
chrjohn: Is there any specific reason to make this package-private? I have also seen it for other constants in this PR.

guidomedina:
IntelliJ says that modifiers at interfaces are redundant, in an interface they are always `static final` for constants.

quickfixj-core/src/main/java/quickfix/Initiator.java
guidomedina: IntelliJ says that modifiers at interfaces are redundant, in an interface they are always static final for constants.

quickfixj-core/src/main/java/quickfix/Session.java
```diff
@@ -344,7 +343,7 @@
 
     public static final String SETTING_MAX_SCHEDULED_WRITE_REQUESTS = "MaxScheduledWriteRequests";
```
chrjohn: These constants remain as public...

guidomedina: It is only at interfaces.

quickfixj-core/src/main/java/quickfix/Initiator.java
chrjohn: Of course. Sorry, I did not realize it was an interface. :)

guidomedina: Well, IntelliJ inspection, but that's a Java rule TBH.

quickfixj-core/src/main/java/quickfix/Session.java
chrjohn: Just deleted the comment ;)

chrjohn:
Looks good to me. Thank you.
Will cherry-pick it later.

guidomedina:
Good, next week I will take another look from a Java 8 perspective.

chrjohn:
Thanks, have a nice weekend.

chrjohn:
Cherry picked with commit 629b801446431f06e0426aebcf669a2eca341a69

PR #83, opened 2016-10-24
Post migration to JDK 8: Functions, lambdas and others

guidomedina:
Post migration to JDK 8:
- [x] Functions, lambdas and use of computeIfAbsent to behave like a true cache.
- [ ] Migrate date formatters and calendars to Java 8 JSR-310.

There is a new class org.quickfixj.SimpleCache, is it at the right place?

### DIFF
--- a/quickfixj-core/src/test/resources/quickfix/test/acceptance/definitions/server/fix40/1b_DuplicateIdentity.def
+++ b/quickfixj-core/src/test/resources/quickfix/test/acceptance/definitions/server/fix40/1b_DuplicateIdentity.def
@@ -13,4 +13,3 @@ I2,8=FIX.4.035=A34=149=TW52=<TIME>56=ISLD98=0108=30
 # wait for disconnect for second connection
 e2,DISCONNECT
 # force disconnect on first connection
-i1,DISCONNECT
--- a/quickfixj-core/src/test/resources/quickfix/test/acceptance/definitions/server/fix41/1b_DuplicateIdentity.def
+++ b/quickfixj-core/src/test/resources/quickfix/test/acceptance/definitions/server/fix41/1b_DuplicateIdentity.def
@@ -13,4 +13,3 @@ I2,8=FIX.4.135=A34=149=TW52=<TIME>56=ISLD98=0108=30
 # wait for disconnect for second connection
 e2,DISCONNECT
 # force disconnect on first connection
-i1,DISCONNECT
--- a/quickfixj-core/src/test/resources/quickfix/test/acceptance/definitions/server/fix42/1b_DuplicateIdentity.def
+++ b/quickfixj-core/src/test/resources/quickfix/test/acceptance/definitions/server/fix42/1b_DuplicateIdentity.def
@@ -13,4 +13,3 @@ I2,8=FIX.4.235=A34=149=TW52=<TIME>56=ISLD98=0108=30
 # wait for disconnect for second connection
 e2,DISCONNECT
 # force disconnect on first connection
-i1,DISCONNECT
--- a/quickfixj-core/src/test/resources/quickfix/test/acceptance/definitions/server/fix43/1b_DuplicateIdentity.def
+++ b/quickfixj-core/src/test/resources/quickfix/test/acceptance/definitions/server/fix43/1b_DuplicateIdentity.def
@@ -13,4 +13,3 @@ I2,8=FIX.4.335=A34=149=TW52=<TIME>56=ISLD98=0108=30
 # wait for disconnect for second connection
 e2,DISCONNECT
 # force disconnect on first connection
-i1,DISCONNECT
--- a/quickfixj-core/src/test/resources/quickfix/test/acceptance/definitions/server/fix44/1b_DuplicateIdentity.def
+++ b/quickfixj-core/src/test/resources/quickfix/test/acceptance/definitions/server/fix44/1b_DuplicateIdentity.def
@@ -13,4 +13,3 @@ I2,8=FIX.4.435=A34=149=TW52=<TIME>56=ISLD98=0108=30
 # wait for disconnect for second connection
 e2,DISCONNECT
 # force disconnect on first connection
-i1,DISCONNECT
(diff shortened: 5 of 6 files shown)


guidomedina:
Had to push -f to because I was starting the dispatching thread twice, making the tests go crazy, don't merge yet I might add more commits, but review so that you get a feeling of JDK 8 changes.

guidomedina:
@chrjohn there is a new class:

`org.quickfixj.SimpleCache`

is it at the right place?

quickfixj-core/src/main/java/quickfix/field/converter/UtcTimestampConverter.java
```diff
@@ -19,24 +19,33 @@
...
+    private final static SimpleCache<String, Long> dateCache = new SimpleCache<>(dateString -> {
+        final Calendar c = new GregorianCalendar(1970, 0, 1, 0, 0, 0);
+        c.setTimeZone(SystemTime.UTC_TIMEZONE);
+        final int year = Integer.parseInt(dateString.substring(0, 4));
+        final int month = Integer.parseInt(dateString.substring(4, 6));
+        final int day = Integer.parseInt(dateString.substring(6, 8));
+        c.set(year, month - 1, day);
+        return c.getTimeInMillis();
+    });
+
```
guidomedina: @chrjohn this is an interesting function, I think we don't need a calendar for that, if it could be migrated to JDK 8 so avoid such invocation.

Also, date & time formatters now won't need `ThreadLocal`, I'll see if I can add more of that stuff on Friday, busy days ahead on my other project.

quickfixj-core/src/main/java/quickfix/DefaultDataDictionaryProvider.java
```diff
@@ -20,54 +20,51 @@
...
+                    return new DataDictionary(path);
+                } catch (ConfigError e) {
+                    throw new QFJException(e);
+                }
+            }
+            return null;
+        });
+        applicationDictionaries = new SimpleCache<>(appVersionKey -> {
+            if (findDataDictionaries) {
+                final String beginString = toBeginString(appVersionKey.applVerID);
```
guidomedina: I'm thinking to change this type of key from `AppVersionKey` to `ApplVerID` no instantiation will be needed.

chrjohn:
Hi @guidomedina , thank you, I will look at it later this week or weekend.

quickfixj-core/src/main/java/org/quickfixj/SimpleCache.java
```diff
@@ -0,0 +1,42 @@
...
+
+    public SimpleCache(Function<K, V> loadingFunction) {
+        super(CPUs, 0.7f, CPUs);
+        this.loadingFunction = loadingFunction;
+    }
+
+    public V computeIfAbsent(K key) {
+        /*
+         * We could computeIfAbsent directly but for CPUs < 32 pre-scanning is faster.
+         */
```
chrjohn: Just curious: Shouldn't we then have a code path that directly uses computeIfAbsent if number of CPUs is greater than 32 (for the very rare case ;)).

guidomedina: This is actually the recommendation by the Doug Lea the author `ConcurrentHashMap` for Java 8 (aka CHMv8)

guidomedina: http://jsr166-concurrency.10961.n7.nabble.com/Re-ConcurrentHashMap-computeIfAbsent-td11687.html

guidomedina:
Rebased and added some plugin updates, I was looking at the date time stuff and it will take some effort to do so I think what has been done is good enough for now, later we can comeback to it and migrate more stuff to Java 8.

chrjohn:
Sorry, did not have time yet to take a deeper look, but anyway thanks for the PR. :+1:

Just a quick question: does the conversion from for-each-loop to the functional style have any benefit? I do not find it easier to understand. I mean, maybe there are cases where it might be handy, but in this commit it is just the same logic but with a different header.

guidomedina:
I didn't change any for-each, I explicitly left these out, the only functional things are anonymous classes and functional interfaces which are interfaces with one method are inferred as functional by Java 8 hence allow you to omit boiler plate code.

Almost two years ago I found difficult to understand Java 8, then it becomes very easy.

guidomedina:
I have a custom QFJ 1.7.x compiled version, running what is on master in production for weeks now, no issues, I also introduce a `NullStore` to avoid storing market data messages which are not needed to be resend, I should probably PR it, WDYT?

```
import org.slf4j.LoggerFactory;
import quickfix.*;

import java.util.Collection;
import java.util.Date;

public class NullStore implements MessageStore {

  public static final class NullStoreFactory implements MessageStoreFactory {

    @Override
    public MessageStore create(SessionID sessionID) {
      return new NullStore(sessionID);
    }
  }

  private final SessionID sessionID;

  private Date creationTime = new Date();
  private int nextSenderMsgSeqNum = 1;
  private int nextTargetMsgSeqNum = 1;

  public NullStore(SessionID sessionID) {
    this.sessionID = sessionID;
  }

  public void get(int startSequence, int endSequence, Collection<String> messages) {
  }

  public Date getCreationTime() {
    return creationTime;
  }

  public int getNextSenderMsgSeqNum() {
    return nextSenderMsgSeqNum;
  }

  public int getNextTargetMsgSeqNum() {
    return nextTargetMsgSeqNum;
  }

  public void incrNextSenderMsgSeqNum() {
    nextSenderMsgSeqNum++;
  }

  public void incrNextTargetMsgSeqNum() {
    nextTargetMsgSeqNum++;
  }

  public void reset() {
    creationTime = new Date();
    nextSenderMsgSeqNum = 1;
    nextTargetMsgSeqNum = 1;
  }

  public boolean set(int sequence, String message) {
    return true;
  }

  public void setNextSenderMsgSeqNum(int next) {
    nextSenderMsgSeqNum = next;
  }

  public void setNextTargetMsgSeqNum(int next) {
    nextTargetMsgSeqNum = next;
  }

  public void refresh() {
    final String text = "null store does not support refresh!";
    if (sessionID != null) {
      final Session session = Session.lookupSession(sessionID);
      session.getLog().onErrorEvent("ERROR: " + text);
    } else {
      LoggerFactory.getLogger(quickfix.MemoryStore.class).error(text);
    }
  }
}
```

chrjohn:
Hi @guidomedina,

yes, please open a separate PR. You could also use the PersistMessages=N option to not persist any messages.
But I like your implementation anyway since it is even doing less processing than a MemoryStore used in conjunction with PersistMessages=N.

Cheers,
Chris.

chrjohn:
Hi @guidomedina , could you please reopen this? I didn't yet merge the JDK8 changes.
Thanks

chrjohn:
OK, just saw #88 :)

guidomedina:
I split them so that one doesn't interfere with the other.

PR #90, opened 2016-12-12
No-op stores does nothing hence test class is not needed.

guidomedina:
`NoopStore` does nothing hence all tests will fail as it will not store messages, it will simply return true to any action.


guidomedina:
@chrjohn I figured what it was, `NoopStore` instance created from `NoopStoreFactory` passes the session ID and when refreshing, the session is not corresponding to that session ID is not found.

We can just remove the session ID from that class and that test will pass, but the others won't, that's because the other tests will spec the messages to be present.

The only tests that will pass are the ones related to message sequences.

guidomedina:
Edited previous comment.

guidomedina:
Don't merge yet, I will add another modification to make it even cleaner.

chrjohn:
Okeydokey.

guidomedina:
Done, check now.

guidomedina:
Is this valid with this `NoopStore`?
```
    final MessageStoreFactory messageStoreFactory;
    if (needMessageStore) {
      messageStoreFactory = new MemoryStore.MemoryStoreFactory();
    } else {
      settings.setString(sessionId, SETTING_PERSIST_MESSAGES, "N");
      messageStoreFactory = new NoopStoreFactory();
    }
```

That `MemoryStore` that you see there is similar to the one from QFJ but the `HashMap` is based on FastUtil:

```
  ...
  private final Map<Integer, String> messages = new Int2ObjectOpenHashMap<>(10000, 0.7f);
  ...
```

chrjohn:
Sorry, I do not quite understand the question. Where is that piece of code from?

guidomedina:
That's just an example, not part of QFJ, the question is in regard the use of `NoopStore` combined with the `PersistMessages` option set at the session settings programmatically.

guidomedina:
Ah, I see my question was missing the last statement, the question should had been:
Is the following example with `NooStore` and `PersistMessages` setting OK?

chrjohn:
Yes, that should work IMHO.

PR #93, opened 2017-01-12
removed '(optional)' qualifier for TargetCompID from configuration docs

brobits:
TargetCompIDs are required and should not have an (optional) qualifier in the docs, here: http://www.quickfixj.org/quickfixj/usermanual/1.5.0/usage/configuration.html


chrjohn:
Thanks for the PR :)

PR #94, opened 2017-01-16
Allow a custom SessionSchedule

Arthurm1:
Currently the scheduling options are weekly or daily or to use an external scheduler.

This allows a custom scheduler to be defined by the user by implementing the SessionSchedule interface.

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/DefaultSessionFactory.java
+++ b/quickfixj-core/src/main/java/quickfix/DefaultSessionFactory.java
@@ -46,6 +46,7 @@ public class DefaultSessionFactory implements SessionFactory {
     private final MessageStoreFactory messageStoreFactory;
     private final LogFactory logFactory;
     private final MessageFactory messageFactory;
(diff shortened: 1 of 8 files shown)


chrjohn:
Hi @Arthurm1 , thanks for the PR. Did you implement a special kind of schedule in your application?

Arthurm1:
I was going to if the PR was accepted :-)
I have a daily schedule which I want to work for only Mon->Fri.  Currently I've got an external fudge which occasionally fails.  I was going to create a modified version of the SessionSchedule that allowed days of the week to be configured.

chrjohn:
OK, as far as I can see the changes you did should not break any backward compatibility, right? When no SessionScheduleFactory is passed to the constructor then it will use the DefaultSessionScheduleFactory. So I think we could merge this to 1.6.x branch also.

Arthurm1:
I did rename SessionSchedule to DefaultSessionSchedule but there doesn't seem to be a way for anyone to use a SessionSchedule even though they could create one so I thought that would be OK.  The rest was additional constructors.

PR #96, opened 2017-01-24
QFJ-912: Support use of externally supplied ThreadPools.

jimbogithub:
**Purpose:**

To provide a mechanism for using externally supplied ThreadPools/Executors instead of the internal QFJ managed Threads.  Expected use case is for ResourceAdapter authors to inject the WorkManager (wrapped as an Executor).

**Solution:**

Defines a new quickfix.ExecutoryFactory interface and adds SessionConnector#setExecutorFactory(...) so that externally supplied Executors can be used.  One Executor is for the timer tasks and one for the message processing.

The Acceptors/Initiators pass the message processing Executor through to their EventHandlingStrategy.  The EventHandlingStrategy classes have been adapted to support the external Executor or fallback to the old behavior if no Executor is supplied.

**Compatibility:**

- There are no API breakages, only new optional API.
- The message processing and timer Thread names are different when running with the Executors.  They start with the same name but are suffixed with the original thread-pool supplied name.

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/ExecutorFactory.java
+++ b/quickfixj-core/src/main/java/quickfix/ExecutorFactory.java
@@ -1,0 +1,32 @@
+package quickfix;
+
+import java.util.concurrent.Executor;
+import java.util.concurrent.Executors;
+
+/**
+ * <p>
+ * Two Executors are required. The {@link #getLongLivedExecutor()} is used for message processing and the
+ * {@link #getShortLivedExecutor()} is used for timer tasks. They can both be the same underlying Executor if desired.
+ * Separating them allows for additional control such as with a ResourceAdapter WorkManager where the WorkManager
+ * differentiates between short and long lived Work.
+ * </p>
+ * <p>
+ * By way of example, a single {@link Executors#newCachedThreadPool()} satisfies the requirements but really adds
+ * nothing over the default behavior.
+ * </p>
+ */
+public interface ExecutorFactory {
+
+	/**
+	 * The message processing activities are long-lived so this Executor must have sufficient distinct Threads available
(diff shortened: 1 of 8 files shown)


chrjohn:
Thanks for the PR. Looks like a sensible enhancement when running in a managed environment. Are you passing in a JEE ManagedExecutorService in your application?

jimbogithub:
You certainly could use a JEE ManagedExecutorService.  In my case I'm using a javax.resource.spi.work.WorkManager as I have ExecutorService wrappers for it that pre-date the JEE Concurrency API.  The WorkManager is fully portable across older containers and the ManagedExecutorService lookup from a ResourceAdapter is container specific (it's not available to the RA at the default location).

The ManagedThreadFactory would have actually been an easier fit for this change but I couldn't adapt the WorkManager to that and I think the ExecutorService is more flexible in the long term.

One possible improvement could be to check if the getShortLivedExecutor() implements ScheduledExecutorService and use that directly for the timers.  For the timers I currently still use the existing ScheduledExecutorService and hand that off to another Executor so there are two threads in play.  I didn't want to have to dig too deep into the static singleton nature of the existing timer approach.  Are you brave enough to look at this?  Would the ScheduledExecutorService supplier still need to ensure they have only a single Thread?

chrjohn:
> Would the ScheduledExecutorService supplier still need to ensure they have only a single Thread?
I did not look very long at this but I think it should at least be ensured that there is not more than one thread per Session.
The current approach with having one thread that iterates over all Sessions is actually not ideal (e.g. http://www.quickfixj.org/jira/browse/QFJ-555).

jimbogithub:
I think the ScheduledExecutorService enhancements would require more invasive changes so I'd be keen to proceed with things as they are and leave that for another day.

chrjohn:
Agreed, thank you. :)

PR #100, opened 2017-01-27
Whitespace removed from field number=276 (QuoteCondition)

damianfanaro:
Affected files: FIX50SP1.modified.xml, FIX50SP2.modified.xml, FIX50SP1.xml, FIX50SP2.xml.
Affected value: <value enum=**"f "** description="FAST_MARKET_ETH"/>

I guess this whitespace is a typo.


chrjohn:
Thanks for the PR. It seems that this error is also present in the original specification files from FPL and we imported it from there. :)

PR #101, opened 2017-02-24
QFJ-885 Send logout message before stopping initiators

alphafoobar:
Because logout sets a flag indicating we should logout, we rely on another thread to send the logout message.

If we stop the initiators first that doesn't happen.

See JIRA: (QFJ-885)[http://www.quickfixj.org/jira/browse/QFJ-885]

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/ThreadedSocketInitiator.java
+++ b/quickfixj-core/src/main/java/quickfix/ThreadedSocketInitiator.java
@@ -82,11 +82,12 @@ public class ThreadedSocketInitiator extends AbstractSocketInitiator {
     }
 
     public void stop(boolean forceDisconnect) {
-        stopInitiators();
         logoutAllSessions(forceDisconnect);
+        stopSessionTimer();
         if (!forceDisconnect) {
             waitForLogout();
         }
+        stopInitiators();
         eventHandlingStrategy.stopDispatcherThreads();
         Session.unregisterSessions(getSessions());
     }
--- a/quickfixj-core/src/main/java/quickfix/ThreadedSocketInitiator.java
+++ b/quickfixj-core/src/main/java/quickfix/ThreadedSocketInitiator.java
@@ -83,10 +83,6 @@ public class ThreadedSocketInitiator extends AbstractSocketInitiator {
 
     public void stop(boolean forceDisconnect) {
         logoutAllSessions(forceDisconnect);
-        stopSessionTimer();
-        if (!forceDisconnect) {
-            waitForLogout();
-        }
         stopInitiators();
         eventHandlingStrategy.stopDispatcherThreads();
         Session.unregisterSessions(getSessions());
--- a/quickfixj-core/src/main/java/quickfix/SocketAcceptor.java
+++ b/quickfixj-core/src/main/java/quickfix/SocketAcceptor.java
@@ -105,7 +105,7 @@ public class SocketAcceptor extends AbstractSocketAcceptor {
 
     @Override
     public void stop(boolean forceDisconnect) {
-        eventHandlingStrategy.stop();
+        eventHandlingStrategy.stopHandlingMessages();
         synchronized (lock) {
             try {
                 try {
@@ -115,7 +115,6 @@ public class SocketAcceptor extends AbstractSocketAcceptor {
                 }
                 logoutAllSessions(forceDisconnect);
                 stopSessionTimer();
-                eventHandlingStrategy.stopHandlingMessages();
             } finally {
                 Session.unregisterSessions(getSessions());
                 isStarted = Boolean.FALSE;
--- a/quickfixj-core/src/main/java/quickfix/SocketAcceptor.java
+++ b/quickfixj-core/src/main/java/quickfix/SocketAcceptor.java
@@ -105,7 +105,7 @@ public class SocketAcceptor extends AbstractSocketAcceptor {
 
     @Override
     public void stop(boolean forceDisconnect) {
-        eventHandlingStrategy.stopHandlingMessages();
+        eventHandlingStrategy.stop();
         synchronized (lock) {
             try {
                 try {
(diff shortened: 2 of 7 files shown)


quickfixj-core/src/main/java/quickfix/ThreadedSocketInitiator.java
```diff
@@ -82,11 +82,12 @@ public void stop() {
     }
 
     public void stop(boolean forceDisconnect) {
-        stopInitiators();
         logoutAllSessions(forceDisconnect);
+        stopSessionTimer();
         if (!forceDisconnect) {
             waitForLogout();
```
alphafoobar: This isn't required because `logoutAllSessions(forceDisconnect)` does exactly the same, unless the sessions weren't initialised... in which case any point waiting for logout?

chrjohn:
Hi @alphafoobar 
Thanks for the PR. I'll have a look.
I recall doing the same changes on a private branch but there was some unit test failing afterwards. Will check again.
Cheers,
Chris.

quickfixj-core/src/main/java/quickfix/ThreadedSocketInitiator.java
```diff
@@ -82,11 +82,12 @@ public void stop() {
     }
 
     public void stop(boolean forceDisconnect) {
-        stopInitiators();
         logoutAllSessions(forceDisconnect);
+        stopSessionTimer();
```
chrjohn: Yeah, call to stopSessionTimer() has been removed in https://github.com/quickfix-j/quickfixj/commit/5f365094aec7b529f91ae518313c8b12517bdede

quickfixj-core/src/main/java/quickfix/ThreadedSocketInitiator.java
chrjohn: You are right, it is not necessary.

alphafoobar:
Hi @chrjohn 
NP, I'll take a look at what unit testing is around this. If I didn't break one, then I should add one.
Cheers,
James

chrjohn:
Hi @alphafoobar , regarding the unit test: I checked it and there was no existing test failing but one that I wrote to check for the correct bahaviour.
However, there have to be more changes to correct the behaviour since you were only changing the ThreadedSocketInitiator. But the problems were reported against the default SocketInitiator. Did you observe the problematic behaviour only with the ThreadedSocketInitiator or did you not try with the default SocketInitiator?

Thanks,
Chris.

alphafoobar:
Hi @chrjohn, 

Thanks for checking, we only use the `ThreadedSocketInitiator`.

Cheers
James

alphafoobar:
Hi @chrjohn 

I have added some unit testing that shows a logout is happening, I was able to add those tests onto an unpatched branch and show that a logout was not sent.

I also looked at adding this for the `SocketInitiator` as I saw this does class does exhibit the same symptoms. But that appears to require more work for the acceptance tests to work, so I backed that change out.

Thanks,
James

chrjohn:
Thanks @alphafoobar , I will take a look shortly.

chrjohn:
Hi @alphafoobar, does your build run successfully with this changes? I have several unit test failures, e.g. in SocketInitiatorTest.testDoubleStartOfInitiator.
Thanks,
Chris.

alphafoobar:
Hi @chrjohn,

Sorry about that, I am seeing that unit test fail. I'll look into why that's happening.

Cheers,
James

alphafoobar:
Hi @chrjohn,

With the changes in this PR, the `QFJ Message Processor` threads weren't dying as quickly. So the problem wasn't that `SocketInitiatorTest.testDoubleStartOfInitiator` had initiated two sessions; but instead that the others tests may still have had lingering Acceptor threads.

Sending the `END_OF_STREAM` message before the synchronization block is good for emptying the message queue, but not quite as good for closing the collection quickly.

Since the acceptor change is breaking tests I've backed it out. 

Cheers,
James

chrjohn:
Hi @alphafoobar , thanks for your input. I am thinking about relaxing the synchronization around the `start()/stop()` methods of the Initiator to prevent the deadlock when `stopHandlingMessages()` is called on an Initiator in the `block()` method. The synchronization is merely done so that you cannot call `stop()` and `start()` concurrently. However, the `SingleThreadedEventHandlingStrategy` itself already checks if there already is a message processing thread active and throws an Exception in that case.

What do you think?
Cheers,
Chris.

alphafoobar:
Hi @chrjohn 

It sounds sensible, starting and stopping the initiator/acceptors are activities that are unlikely to see much contention and the `ThreadedSocketInitiator` start/stop methods aren't synchronized.

I've been thinking that it would make more sense for the blocking thread to add the END_OF_STREAM message on the end of the queue once it realises it has been stopped, for example in `SingleThreadedEventHandlingStrategy.block()`:

```java
if (!eventQueue.isEmpty()) {
    final List<SessionMessageEvent> tempList = new ArrayList<>();
    eventQueue.drainTo(tempList);
    for (SessionMessageEvent event : tempList) {
        event.processMessage();
    }
    for (Session quickfixSession : sessionConnector.getManagedSessions()) {
        new SessionMessageEvent(quickfixSession, END_OF_STREAM).processMessage();
    }
}
```

I'm happy to help,
James

chrjohn:
Hi  @alphafoobar ,
after reviewing your changes again, I think I found an easier (i.e. less invasive) way to correct the behaviour.
One part of the solution were definitely the changes in https://github.com/quickfix-j/quickfixj/pull/101/files#diff-76633e507f7b944754a248ac2d00434c
Will check again and come back.
Thanks,
Chris.

alphafoobar:
Hi Chris, sounds good. Cheers!

chrjohn:
Hi @alphafoobar,
I have created a new pull request #109 based on your changes. After fiddling around with your pull request and doing some other changes it was a bit of a mess. ;) So I hope it is OK with you to close this PR and that #109 resolves your issues.
Thanks for all your help and cheers,
Chris.

alphafoobar:
Cool, thanks @chrjohn !

PR #102, opened 2017-02-28
QFJ-917: multi-threaded disconnect fails to call onLogout

upchuck:
If multiple threads call disconnect, the logonReceived/logonSent can be updated by multiple threads.  The first thread might never call onLogout if the thread was interrupted after releasing the lock and before caching the state.


chrjohn:
Thank you.

upchuck:
@boxhead201 Good point on setting twice.  I updated to initialize with false.

chrjohn:
Hi @upchuck  and @boxhead201 , thanks for your input. I am going to close this now in favour of #104 since it also has a unit test.
Cheers,
Chris

PR #103, opened 2017-03-02
bugfix:  real field-ending SOH

btomas:


chrjohn:
Thanks for the PR. Could you please describe shortly which bug you did encounter? Are you able to provide a unit test?
Thanks,
Chris.

guidomedina:
This is very suspicious as the real check if happening after an exception is thrown which is bad since exceptions are slower than normal flow.

You shouldn't relay on exceptions to process normal flow of an execution but use them for exactly that, exceptions.

chrjohn:
Thanks @guidomedina  for the review. You are completely right.

PR #104, opened 2017-03-09
QFJ-917: multi-threaded disconnect fails to call onLogout (Take #2)

boxhead201:
This is a variation of upchuck's pull request with a unit test to verify that the issue is fixed.

The gist of this pull request:
1.  Move logonReceived and logonSent local variables in Session.disconnect(String,boolean) to top of method
2.  Pull up PausableThreadPoolExecutor into its own top level class that can be leveraged by other test classes.
3.  Create unit test in SessionDisconnectConcurrentlyTest.  In order to get this problem to come out sometimes I have to bombard the disconnect method with a thousand threads while in the context of a for loop.

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/Session.java
+++ b/quickfixj-core/src/main/java/quickfix/Session.java
@@ -23,37 +23,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import quickfix.Message.Header;
 import quickfix.SessionState.ResendRange;
-import quickfix.field.ApplVerID;
-import quickfix.field.BeginSeqNo;
-import quickfix.field.BeginString;
-import quickfix.field.BusinessRejectReason;
-import quickfix.field.DefaultApplVerID;
-import quickfix.field.EncryptMethod;
-import quickfix.field.EndSeqNo;
-import quickfix.field.GapFillFlag;
-import quickfix.field.HeartBtInt;
-import quickfix.field.LastMsgSeqNumProcessed;
-import quickfix.field.MsgSeqNum;
-import quickfix.field.MsgType;
-import quickfix.field.NewSeqNo;
-import quickfix.field.NextExpectedMsgSeqNum;
-import quickfix.field.OrigSendingTime;
-import quickfix.field.PossDupFlag;
-import quickfix.field.RefMsgType;
-import quickfix.field.RefSeqNum;
-import quickfix.field.RefTagID;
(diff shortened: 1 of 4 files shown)


upchuck:
Thanks for all the hard work.  I agree with your new changes.

chrjohn:
Hi @boxhead201 , I have changed the unit test to make provoking the error a little more likely (IMHO a sleep should increase the likeliness of context switching).
Before the changes the test was running about three minutes on my machine, now it's about 20 seconds and I was still able to provoke the erroneous behaviour (when I reverted your fix).

Umm, actually, I wanted to push the changes to your PR branch but for some reason I was too dumb. ;)
Anyway, the only change I made was adding a `sleep(10)` to the `disconnect()` method of the `UnitTestResponder`. That way I could reduce the number of loops to 1000 instead of 100000.
What do you think? Can you still provoke the error using my changes?

Thanks,
Chris.

boxhead201:
Hi Chris;

  Initially when I was working on creating a unit test for this issue, I broke up the Session.disconnect() method into 3 methods; a method that set the responder to null, another method that calls Application.onLogout(), and the parent method that calls the 2 aforementioned methods.  Then in the unit test I subclassed Session.disconnect() and inserted Thread.yield() at various places in order to make the problem appear.  This unit test worked better because the problem came out right away; with less threads and loops.  But I did not like this unit test because it involved too much change to the Session.disconnect() method, and it also involves that the unit test subclass Session.disconnect() with special behavior just for the unit test.  I wanted the change to Session.disconnect() to be as simple as possible so I threw away this work and came up with the unit test in this pull request instead.  The downside of the unit test in this pull request is the results vary widely depending on environment.  I think for the long term, Session.disconnect() refactored to do less; it is doing too much now.

  I experimented with putting a Thread.sleep() and a Thread.yield() in the UnitTestResponder this morning and my results at this addition were mixed at best.  Sometimes the test failed right away, sometimes the test did not fail at all.  When I leave the UnitTestResponder unchanged, then my tests fail faster, contrary to your experience.  I added Thread.sleep() into the UnitTestResponder to facilitate your findings.  Also, FYI, I am using 100 loops in the parent test method, and 1000 threads.

boxhead201:
Hi Chris;  

  Do you know if you will be merging this pull request anytime soon?  Or do you anticipate further changes?

chrjohn:
Hi Paul, I am ill currently but will be able to do it when back in the office. 
Cheers

PR #105, opened 2017-03-23
Updates to allow for custom dictionary provider in DefaultSessionFactory

msbg:
Good morning, small update here that allows for custom dictionary providers to be used with the DefaultSessionFactory.  If you're running a custom codegen from an altered XML file, it makes it easy to use a custom DictionaryProvider that can refer to the altered XML file directly.  Any thoughts greatly appreciated.

Regards,

- monica

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/DefaultDataDictionaryProvider.java
+++ b/quickfixj-core/src/main/java/quickfix/DefaultDataDictionaryProvider.java
@@ -25,7 +25,7 @@ import quickfix.field.ApplVerID;
 
 import static quickfix.MessageUtils.toBeginString;
 
-public class DefaultDataDictionaryProvider implements DataDictionaryProvider {
+public class DefaultDataDictionaryProvider implements EditableDataDictionaryProvider {
     private final SimpleCache<String, DataDictionary> transportDictionaries;
     private final SimpleCache<ApplVerID, DataDictionary> applicationDictionaries;
 
--- a/quickfixj-core/src/main/java/quickfix/DefaultSessionFactory.java
+++ b/quickfixj-core/src/main/java/quickfix/DefaultSessionFactory.java
@@ -47,6 +47,7 @@ public class DefaultSessionFactory implements SessionFactory {
     private final LogFactory logFactory;
     private final MessageFactory messageFactory;
     private final SessionScheduleFactory sessionScheduleFactory;
+    private final EditableDataDictionaryProvider editableDataDictionaryProvider;
 
     public DefaultSessionFactory(Application application, MessageStoreFactory messageStoreFactory,
             LogFactory logFactory) {
@@ -55,6 +56,7 @@ public class DefaultSessionFactory implements SessionFactory {
         this.logFactory = logFactory;
         this.messageFactory = new DefaultMessageFactory();
         this.sessionScheduleFactory = new DefaultSessionScheduleFactory();
+        this.editableDataDictionaryProvider = new DefaultDataDictionaryProvider();
     }
 
     public DefaultSessionFactory(Application application, MessageStoreFactory messageStoreFactory,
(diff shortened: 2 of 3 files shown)


chrjohn:
Thanks for the PR, will have a look later on...
Cheers,
Chris.

chrjohn:
Just a question: you did not add the methods `addTransportDictionary()` and `addApplicationDictionary()` to the `DataDictionaryProvider` interface to maintain backward compatibility?
I am currently a bit slow on the uptake. Can you give me an example where this is useful? Do you want to load dictionaries after the Session has been constructed?
Thanks,
Chris.

msbg:
Correct, I didn't touch the existing interface for reasons of backward compatibility - but can change if you prefer.  The use case is when I have manually altered a datadictionary xml file locally, and therefore renamed it (so that I can tell which file is what!), but the default data dictionary provider assumes the xml name to be the same as the appVerId. I can easily create my own DataDictionaryProvider to adjust the xml filename mapping, but without this update I can't assign my implementation into the session without a custom version of the DefaultSessionFactory. Of course I may be missing an alternative solution, and apologies if so!
Many thanks for looking at this, it's really appreciated,

Regards,

- monica

quickfixj-core/src/main/java/quickfix/DefaultDataDictionaryProvider.java
```diff
@@ -25,7 +25,7 @@
 
 import static quickfix.MessageUtils.toBeginString;
 
-public class DefaultDataDictionaryProvider implements DataDictionaryProvider {
+public class DefaultDataDictionaryProvider implements EditableDataDictionaryProvider {
```
philipwhiuk: Please update the methods being overridden to add `@Override` notation.

quickfixj-core/src/main/java/quickfix/EditableDataDictionaryProvider.java
```diff
@@ -0,0 +1,13 @@
+package quickfix;
+
+
+import quickfix.field.ApplVerID;
+
+/**
+ * Editable dictionary provider used by the DefaultSessionFactory
```
philipwhiuk: I'm not sure this Javadoc is particularly useful and if the DefaultSessionFactory gets re-factored nobody will remember to update this comment (at which point it will be actively unhelpful).

quickfixj-core/src/main/java/quickfix/DefaultSessionFactory.java
```diff
@@ -131,9 +147,13 @@ public Session create(SessionID sessionID, SessionSettings settings) throws Conf
...
             }
 
-            DefaultDataDictionaryProvider dataDictionaryProvider = null;
+            EditableDataDictionaryProvider dataDictionaryProvider = null;
             if (useDataDictionary) {
-                dataDictionaryProvider = new DefaultDataDictionaryProvider();
+                if(editableDataDictionaryProvider == null) {
+                    dataDictionaryProvider = new DefaultDataDictionaryProvider();
+                } else {
+                    dataDictionaryProvider = editableDataDictionaryProvider;
```
philipwhiuk: You now have a dataDictionaryProvider per SessionFactory rather than per Session. Do we need to think about this?

Perhaps we should be passing in a DataDictionaryProviderFactory.

chrjohn:
Thanks @philipwhiuk for taking the time for a review.

@msbg : I have to admit I still don't get the whole picture. I did not test this by myself but looked at the QFJ code:
IMHO I can add a data dictionary at runtime by using `Session.getDataDictionaryProvider().addTransportDictionary()`. Of course, the returned provider needs to be cast to a `DefaultDataDictionaryProvider` first.
Could you maybe give some code examples , otherwise I'm afraid I will never get it. ;)

msbg:
Many thanks for the comments - made me think of a different approach.  I'm going to re-submit an alternative shortly.  
Thanks a lot!

- monica

chrjohn:
OK, thanks for your feedback. Please open a new PR then.
Cheers,
Chris.

PR #118, opened 2017-07-04
Fix for race condition when receiving logon response before logon request has been processed by MessageStore

jonfreedman:
As discussed on the [mailing list](https://sourceforge.net/p/quickfixj/mailman/quickfixj-users/thread/1974CA77F2BC6F4AA91619016DCFC9EF7DE94FD7%40P2EXCH01.rivagecapital.com) there appears to be a race condition between storage of logon request and receipt of logon response when validating the expected SeqNum.  `SessionDisconnectTest` simulates this scenario and locking around the call to `SessionState#getNextSenderMsgSeqNum` in `Session#nextLogon` fixes the specific issue.

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/Session.java
+++ b/quickfixj-core/src/main/java/quickfix/Session.java
@@ -2058,7 +2058,7 @@ public class Session implements Closeable {
 
             final int targetWantsNextSeqNumToBe = logon.getInt(NextExpectedMsgSeqNum.FIELD);
             state.lockSenderMsgSeqNum();
-            final int actualNextNum = state.getMessageStore().getNextSenderMsgSeqNum();
+            final int actualNextNum = state.getNextSenderMsgSeqNum();
             state.unlockSenderMsgSeqNum();
             // Is the 789 we received too high ??
             if (targetWantsNextSeqNumToBe > actualNextNum) {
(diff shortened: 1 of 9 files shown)


chrjohn:
Thanks for the PR. The Builder looks nice. Have to check how I can merge the tests to 1.6.x which is still on JDK7. ;)
Cheers

quickfixj-core/src/main/java/quickfix/Session.java
```diff
@@ -2057,7 +2057,9 @@ private void nextLogon(Message logon) throws FieldNotFound, RejectLogon, Incorre
         if (logon.isSetField(NextExpectedMsgSeqNum.FIELD) && enableNextExpectedMsgSeqNum) {
 
             final int targetWantsNextSeqNumToBe = logon.getInt(NextExpectedMsgSeqNum.FIELD);
-            final int actualNextNum = state.getMessageStore().getNextSenderMsgSeqNum();
+            state.lockSenderMsgSeqNum();
+            final int actualNextNum = state.getNextSenderMsgSeqNum();
```
chrjohn: Could you please wrap this in a try-finally-block? Thanks.

jonfreedman: Sorry, I realised on my way out the door last night I forgot this

jonfreedman:
Regarding JDK7 it appears the only module with tests is quickfixj-core so you could introduce a new quickfixj-tests module which is compiled using JDK8?

PR #121, opened 2017-07-07
Use TravisC CI

jonfreedman:
Set up a travis-ci build which will run on every check-in.  Before merging this please complete steps 1-2 [here](https://docs.travis-ci.com/user/getting-started#To-get-started-with-Travis-CI).  Took a while to get this working as the default build runs out of heap space and produces more than 4MB of stdout.  This supersedes  https://github.com/quickfix-j/quickfixj/pull/26.

Adds a travis build badge and also a Maven badge to the readme because why not...

Successful builds
- https://travis-ci.org/jonfreedman/quickfixj/builds/251075138
- https://travis-ci.org/jonfreedman/quickfixj/builds/251086018


chrjohn:
Just a question: QFJ already uses Bamboo as build server. What is the advantage when using travis here? Is it because of the badge? ;) Have no experience with travis yet, so just asking...
Thanks,
Chris.

jonfreedman:
The main reason I can see is that it will run tests on PRs automatically, I'm sure you could configure Bamboo to do that and integrate with GitHub but it doesn't look it's doing that yet.

[Here is an example PR from one of my projects](https://github.com/symphonyoss/hubot-symphony/pull/54) - you can see the green tick under the commit expands to show various status checks include Travis.

chrjohn:
Ah OK. Sounds cool :)

chrjohn:
Now I remember why we didn't use this before. The build is failing too often. ;)
IIRC there are some tests which fail because sometimes the initiator/acceptor cannot be started in quick succession because something blocks.

Will take a look some time later.

jonfreedman:
If you pull out a list of fragile tests it should hopefully be easy to work through them to fix. 

Cheers 

Jon 

---- On Wed, 12 Jul 2017 10:56:11 +0000 notifications@github.com wrote ----

Now I remember why we didn't use this before. The build is failing too often. ;)
IIRC there are some tests which fail because sometimes the initiator/acceptor cannot be started in quick succession because something blocks.

Will take a look some time later.

PR #127, opened 2017-07-24
IntelliJ inspections corrected and some dependencies updated.

guidomedina:
IntelliJ inspections corrected and some dependencies updated.

### DIFF
--- a/quickfixj-core/src/test/java/quickfix/SocketInitiatorTest.java
+++ b/quickfixj-core/src/test/java/quickfix/SocketInitiatorTest.java
@@ -268,29 +268,27 @@ public class SocketInitiatorTest {
         final int port = serverSocket.getLocalPort();
 
         final AtomicBoolean socketConnected = new AtomicBoolean(false);
-        Thread socketThread = new Thread() {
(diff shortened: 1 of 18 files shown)


quickfixj-core/src/test/java/quickfix/SocketInitiatorTest.java
```diff
@@ -145,7 +145,7 @@ public void testBlockLogoffAfterLogon() throws Exception {
                 clientApplication.stopAfterLogon(initiator);
                 clientApplication.setUpLogonExpectation();
 
-                initiator.block();
+                initiator.start();
```
chrjohn: I think it does not make sense to change `block()` to `start()` here since the test specifically tests for `block()` (although test is ignored). Maybe mark it as deprecated so it can be cleaned up when the `block()` implementation is cleaned up.

guidomedina: I have reverted this particular line, later on you can decide if to keep the test or remove it completely along with the block method.

PR #132, opened 2017-08-18
add create_all

welloncn:
maybe someone need a full table create sql for mysql


chrjohn:
Thanks for the PR. But only the first commit is needed. The second (seems to be a merge) not.

PR #133, opened 2017-08-20
Weekday Session Schedule

Arthurm1:
This is a weekday based session schedule that can be used instead of `DefaultSessionSchedule`.

It uses a new session setting ID `Weekdays` which is a comma-delimited list of the days to _**start**_ the session on.

A standard western trading week example...
```
StartTime=22:05:00 UTC
EndTime=21:55:00 UTC
Weekdays=Sun,Mon,Tue,Wed,Thu
```
`StartDay `and `EndDay `are not used.

To tell QuickFIX/J to use the weekly schedule, a `WeekdaySessionScheduleFactory` instance must be passed to the `DefaultSessionFactory` constructor.

### DIFF
--- a/quickfixj-core/src/main/doc/usermanual/usage/configuration.html
+++ b/quickfixj-core/src/main/doc/usermanual/usage/configuration.html
@@ -202,15 +202,29 @@
   </TR>
   <TR ALIGN="left" VALIGN="middle">
     <TD> <I>StartDay</I> </TD>
-    <TD> For week long sessions, the starting day of week for the session. Use in combination with StartTime. </TD>
+    <TD> For week long sessions, the starting day of week for the session.
+      <br>Use in combination with StartTime.
+      <br>Incompatible with Weekdays</TD>
     <TD> Day of week in the default locale (e.g. Monday, mon, lundi, lun. etc.)</TD>
 
     <TD>&nbsp; </TD>
   </TR>
   <TR ALIGN="left" VALIGN="middle">
     <TD> <I>EndDay</I> </TD>
-    <TD> For week long sessions, the ending day of week for the session. Use in combination with EndTime. </TD>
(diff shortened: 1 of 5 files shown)


philipwhiuk:
Hmm..

Did you consider just adding the logic to the DefaultSessionSchedule? It feels like this is a comparable behavioural change to Non-Stop-Session or Weekly Sessions to me. Why not just have it be used if Weekdays is set, rather than having to pick a different SessionSchedule and then that Schedule throw an error when it's not set?

By making it a factory choice you have to pick for an entire QFJ session factory which will be used, rather than picking on a session by session basis.

Arthurm1:
When I wrote this it was outside the QuickFIXJ codebase so I didn't have the option to refactor DefaultSessionSchedule.

You're probably right though.  I could have a look at integrating it into DefaultSessionSchedule if people prefer that and it's likely to be accepted.

chrjohn:
Hi @Arthurm1 , I just started to work on unifying the WeekdaySessionSchedule and DefaultSessionSchedule. So if you could just add the missing documentation that would be great.
Thanks,
Chris.

Arthurm1:
@chrjohn That's great.  I was going to take a look at it myself next week but if you've already started.

I'm good to add `Weekdays` into the config docs.

Looking at how `StartDay`, `EndDay `are translated using `DayConverter `I think I've put too much validation into the `WeeklySessionSchedule`.  It will reject any days that are not 3 letters (e.g. "Mon") but I don't think that is consistent with how `StartDay`, `EndDay `work and may not cope with non-English days so you could probably just take that out.

chrjohn:
Hi @Arthurm1 , see #134 :)
Thanks again for the PR!

Arthurm1:
Thanks for putting it in the DefaultSessionSchedule - that looks better than my extra classes.

PR #135, opened 2017-09-18
Use helper function to extract epochMillis

philipwhiuk:
Remove some of the boilerplate in this test class.


philipwhiuk:
> SSLCertificateTest.shouldFailIndividualSessionsWhenInvalidCertificatesUsed:257 No SSL exception thrown

Odd. My code doesn't change that.

chrjohn:
Hi @philipwhiuk ,
don't worry, the failing build has nothing to do with your changes. It is failing from time to time in the SSL-related tests. But I didn't get around to finding out why. I suspect it has to do with the SSLFilter. Will open a separate issue for that.
Thanks for the PR!
Cheers,
Chris.

PR #138, opened 2017-09-21
Dependencies and plugins updated to latest.

guidomedina:
Dependencies and plugins updated to latest and some IntelliJ inspections suggestions applied.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -74,14 +74,14 @@
 		<mainClass/>
 
 		<maven-resources-plugin-version>3.0.2</maven-resources-plugin-version>
-		<maven-compiler-plugin-version>3.6.1</maven-compiler-plugin-version>
+		<maven-compiler-plugin-version>3.7.0</maven-compiler-plugin-version>
 		<maven-jar-plugin-version>3.0.2</maven-jar-plugin-version>
-		<maven-surefire-plugin-version>2.20</maven-surefire-plugin-version>
+		<maven-surefire-plugin-version>2.20.1</maven-surefire-plugin-version>
 		<maven-pmd-plugin-version>3.8</maven-pmd-plugin-version>
 		<maven-source-plugin-version>3.0.1</maven-source-plugin-version>
 		<maven-javadoc-plugin-version>2.10.4</maven-javadoc-plugin-version>
-		<maven-shade-plugin-version>3.0.0</maven-shade-plugin-version>
-		<maven-assembly-plugin-version>3.0.0</maven-assembly-plugin-version>
+		<maven-shade-plugin-version>3.1.0</maven-shade-plugin-version>
+		<maven-assembly-plugin-version>3.1.0</maven-assembly-plugin-version>
 		<maven-bundle-plugin-version>3.3.0</maven-bundle-plugin-version>
 		<maven-gpg-plugin-version>1.6</maven-gpg-plugin-version>
 		<maven-deploy-plugin-version>2.8.2</maven-deploy-plugin-version>
--- a/quickfixj-codegenerator/pom.xml
+++ b/quickfixj-codegenerator/pom.xml
@@ -35,7 +35,7 @@
 		<dependency>
 		    <groupId>net.sf.saxon</groupId>
 		    <artifactId>Saxon-HE</artifactId>
-		    <version>9.8.0-3</version>
+		    <version>9.8.0-4</version>
 		</dependency>
 	</dependencies>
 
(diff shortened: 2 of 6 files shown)


guidomedina:
@chrjohn I have this running with a custom version which includes current master, for few days now and it looks good.

quickfixj-core/src/test/java/quickfix/SocketInitiatorTest.java
```diff
@@ -145,7 +145,7 @@ public void testBlockLogoffAfterLogon() throws Exception {
                 clientApplication.stopAfterLogon(initiator);
                 clientApplication.setUpLogonExpectation();
 
-                initiator.block();
+                initiator.start();
```
philipwhiuk: Doesn't this make for a potiential race condition between the initiator and the line below?

guidomedina: Not really, it will simply block until it has actually logged in, tests passed twice after this change, on my PC and Jenkins.

guidomedina: Also, the reason for the change was the deprecation of the `block()` method:
```
    @Override
    @Deprecated
    /**
     * Use SocketInitiator.start().
     */
    public void block() throws ConfigError, RuntimeError {
        initialize(false);
    }
```

philipwhiuk: Isn't the point of this code to test the deprecated functionality. This test no longer tests what it says it tests.  It says it's testing `block()`.  You're removing test coverage for deprecated, yet still functional code. We should still test APIs we deprecate yet don't remove.

I would assume `start()` doesn't block and `block()` does. That would be the whole point, no? So you then have a race between start calling logoff and initiator.isLoggedOn() being true.

guidomedina: You may have a point, let me undo this part.

philipwhiuk:
👍

guidomedina:
What confused me about block and start is that they both lead to:

```
            if (blockInThread) {
                eventHandlingStrategy.blockInThread();
            } else {
                eventHandlingStrategy.block();
            }
```

I'm wondering if one will block initiating thread and the other the processing thread, I have to look deeper and if it is actually blocking either way.

chrjohn:
IMHO you can remove the deprecated methods and tests. But maybe that should be a separate PR. Thanks, Chris.

Am 24. September 2017 22:26:10 MESZ schrieb Guido Medina <notifications@github.com>:

guidomedina:
@chrjohn agreed, let's leave that for later, I have to actually spend time on looking what to optimize, I have updated the `StringBuilder` with `ThreadLocal` branch, that's a more interesting thing to look at.

I haven't PR it because I need volunteers to kind of benchmark it and see if it actually makes a difference, it is currently rebased on top of this branch:
https://github.com/guidomedina/quickfixj/tree/stringbuilder-threadlocal-improvement

guidomedina:
Updated to point to actual commit instead of branch: https://github.com/guidomedina/quickfixj/commit/f2c5629087e3db1a17b3ac1f729a1de56749a4e5

chrjohn:
Hi @guidomedina ,
I will take care of the @Deprecated stuff. There are other places in the code that I want to clean up as well.

W.r.t. the StringBuilder changes: did you try to create some benchmarks with JMH?
IMHO the only things that might be sped up (since they are constantly called) are the methods in the Message and Field classes. On the other hand, I am not certain that a Message instance will be handled by threads concurrently since it belongs to a dedicated Session which is normally handled by one thread.
But just off the top of my head, didn't dig through the code...
Cheers,
Chris.

guidomedina:
Yes, this branch doesn't deal with the deprecated stuff anymore, it should be good for merging.

MartyIX:
@chrjohn My understanding is that `StringBuilderHelper` is good because there will be much less objects created for JVM to collect later. I would expect lower GC times in general.

guidomedina:
@MartyIX that's the idea yes, to re-use the same `char[]` within the same thread, it should dramatically reduce GC for price streaming as the initiator or acceptor thread will constantly re-use it.

I have been personally being using such mod for almost 2 months now in production, no issues, I know in theory it should help but I haven't measured it.

chrjohn:
Hi,
@MartyIX , of course, you are right. I was only reacting to the "benchmark" idea and forgot about the re-usage. And yes I would also **expect** lower GC times... :) 
But shouldn't take more than having two unit tests (one with and one without optimization) which are running for say an hour and have GC logging turned on...

chrjohn:
Hi @guidomedina , thanks for the PR. Will merge later this week, I promise. :) First have to sort out some other QFJ issues.

PR #145, opened 2017-10-09
Changed logging to use formatting anchor {} where applicable.

chrjohn:
- https://www.slf4j.org/faq.html#logging_performance


quickfixj-core/src/main/java/org/quickfixj/jmx/JmxExporter.java
```diff
@@ -132,12 +132,12 @@ public void registerMBean(Object mbean, ObjectName objectName) throws JMExceptio
         } catch (final InstanceAlreadyExistsException ex) {
             if (registrationBehaviour == REGISTRATION_IGNORE_EXISTING) {
                 if (log.isDebugEnabled()) {
```
MartyIX: This line may not be necessary as `log.debug` should do the same check (https://stackoverflow.com/a/963681/99256)

chrjohn: You are of course right. I did not pay much attention to the context when doing the replacement. Sorry ;)
I will remove the additional log.isDebugEnabled() checks in the code.
Thanks

PR #146, opened 2017-10-15
QFJ-902: Fixing a bug with attempt to close stream multiple times

alexwibowo:
* When a session window is not 'open', the quickfix timer thread will try to reset the filestore, which also initialises the creation time on the filestore.
* When there is another thread that tries to login (outside the session 'open' window), that thread will try to write something to the filestore, namely the header info, such as sequence number
* If the operation is interleaved in the following manner:
  1. Quickfix timer thread closes the underlying filestore
  2. Thread that accepts the login attempt to update the sequence number (which happens to be using buffered output stream, by the way)
  3. Quickfix timer thread tries to close the underlying filestore again
  Then the 3rd step above will fail. As part of the close, standard Java buffered output stream will see if there is anything on the buffer to be flushed out. In this scenario, there is something (an update to the sequence number). Therefore it will try to write to the output stream. However, the underlying stream is already closed (in step 1). Hence stream closed event error.
  Unfortunately, this last step (step 3) is also where quickfix tries to initialise the filestore creation time. Hence, the creation time is never updated (due to the exception). So any further attempt to login from client will get rejected with 'login attempt outside session window' error.

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/FileStore.java
+++ b/quickfixj-core/src/main/java/quickfix/FileStore.java
@@ -97,10 +97,10 @@ public class FileStore implements MessageStore, Closeable {
     }
 
     void initialize(boolean deleteFiles) throws IOException {
-        close();
-
         if (deleteFiles) {
-            deleteFiles();
+            closeAndDeleteFiles();
+        } else {
+            close();
         }
 
         String mode = READ_OPTION + WRITE_OPTION + (syncWrites ? SYNC_OPTION : NOSYNC_OPTION);
@@ -219,7 +219,7 @@ public class FileStore implements MessageStore, Closeable {
         }
     }
 
-    public void deleteFiles() throws IOException {
+    public void closeAndDeleteFiles() throws IOException {
         close();
         deleteFile(headerFileName);
         deleteFile(msgFileName);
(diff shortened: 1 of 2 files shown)


chrjohn:
Hi, thanks for the PR.
Seems there are tests which are still failing:
```
[ERROR]   FileStoreLowMemTest>FileStoreTest.testResetShouldNeverFail:105
[ERROR]   FileStoreZeroMemTest>FileStoreTest.testResetShouldNeverFail:105
```

alexwibowo:
Do you know how am I meant to run this test from my IDE (intellij)? I run that test (FileStoreLowMemTest) and it passes. I cant find any special surefire setting for this 'LowMemTest'.

chrjohn:
Hmm, just guessing but the following statement could also be false when two consecutive runs are completed in exact the same millisecond, right?

`assertTrue(newCreationTime.after(creationTime));`

alexwibowo:
Hmmm... true. Ok...flaky test...

chrjohn:
Hi @alexwibowo ,
I have slightly changed your unit test since it passed even without your FileStore changes. The problem with doing the `sleep(1)` is that you also reduce the likelihood of the thread interfering with the `store.reset()` call.
Please have a look at https://gist.github.com/chrjohn/af3adc25b2585ebd899ab232b1ba9e8d
I'm in a little of a hurry at the moment, so if you could do the changes that would be great.
Thanks in advance and cheers,
Chris.

jimmysyss:
The root cause should be due to concurrent update of the FileStore object, Since the Random Access File is not thread safe, we need to explicitly Lock the File with Read Write Lock. 

http://tutorials.jenkov.com/java-util-concurrent/readwritelock.html

PR #147, opened 2017-10-25
Better error messages in dictionary generator

aferrandi:
I was creating a dictionary from a repository and got:

- NullpointerExceptions because NotReqXML was written with wrong letter case
- NoSuchElementException because it was not finding the session messages.

This change gives better errors in both cases, so that it is easier to find the reason of the problem.

### DIFF
--- a/quickfixj-dictgenerator/src/main/java/org/quickfixj/dictgenerator/Generator.java
+++ b/quickfixj-dictgenerator/src/main/java/org/quickfixj/dictgenerator/Generator.java
@@ -218,13 +218,11 @@ public class Generator {
     }
 
     private MsgType getNextMsgType(Map<String, MsgType> msgTypes) {
-        try
-        {
-            return msgTypes.values().iterator().next();
-        }
-        catch(NoSuchElementException ex)
-        {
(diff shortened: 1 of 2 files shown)


quickfixj-dictgenerator/src/main/java/org/quickfixj/dictgenerator/Generator.java
```diff
@@ -222,6 +217,17 @@ private void generateDictionary(String file, boolean admin, boolean merged) {
         }
     }
 
+    private MsgType getNextMsgType(Map<String, MsgType> msgTypes) {
+        try
```
philipwhiuk: Code style:

```
try {

} catch {

}
```

chrjohn: I'd also prefer to check for the presence of a next element instead of catching the Exception.

aferrandi:
I applied the requested changes and tested against the repository I was working on. 
It works.

chrjohn:
Great, thanks. :)

PR #149, opened 2017-11-06

pbyrne413:
…ould throw a NumberFormatException.


MartyIX:
There is a typo: `NumberFormatException: Repeating group count requires an <WHAT> but found:`

pbyrne413:
not ready for merge

PR #150, opened 2017-11-06

pbyrne413:
…ould throw a NumberFormatException

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/Message.java
+++ b/quickfixj-core/src/main/java/quickfix/Message.java
@@ -578,7 +578,13 @@ public class Message extends FieldMap {
         final int[] fieldOrder = groupDataDictionary.getOrderedFields();
         int previousOffset = -1;
         final int groupCountTag = field.getField();
-        final int declaredGroupCount = Integer.parseInt(field.getValue());
+		// QFJ-533
+        int declaredGroupCount = 0;
+        try {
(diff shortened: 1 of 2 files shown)


philipwhiuk:
See https://github.com/quickfix-j/quickfixj/pull/151#pullrequestreview-74600756

quickfixj-core/src/main/java/quickfix/Message.java
```diff
@@ -581,9 +581,11 @@ private void parseGroup(String msgType, StringField field, DataDictionary dd, Da
...
         int declaredGroupCount = 0;
         try {
-          declaredGroupCount = Integer.parseInt(field.getValue());
+             declaredGroupCount = Integer.parseInt(field.getValue());
         } catch (final NumberFormatException e) {
-            throw new InvalidMessage("NumberFormatException: Repeating group count requires an Integer but found:" +field.getValue());
+			 InvalidMessage invalidMessage = new InvalidMessage("Repeating group count requires an Integer but found:" +field.getValue());
+			 invalidMessage.initCause(e);
+		     throw invalidMessage;
         }
```
philipwhiuk: Does InvalidMessage not have a constructor `InvalidMessage(String message, Throwable cause)`

pbyrne413: No I can just find the constructor:  InvalidMessage(String message)

PR #151, opened 2017-11-06

pbyrne413:
…ould throw a NumberFormatException

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/Message.java
+++ b/quickfixj-core/src/main/java/quickfix/Message.java
@@ -578,7 +578,13 @@ public class Message extends FieldMap {
         final int[] fieldOrder = groupDataDictionary.getOrderedFields();
         int previousOffset = -1;
         final int groupCountTag = field.getField();
-        final int declaredGroupCount = Integer.parseInt(field.getValue());
+       // QFJ-533
+        int declaredGroupCount = 0;
+        try {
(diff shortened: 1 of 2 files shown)


quickfixj-core/src/main/java/quickfix/Message.java
```diff
@@ -578,7 +578,13 @@ private void parseGroup(String msgType, StringField field, DataDictionary dd, Da
         final int[] fieldOrder = groupDataDictionary.getOrderedFields();
         int previousOffset = -1;
         final int groupCountTag = field.getField();
-        final int declaredGroupCount = Integer.parseInt(field.getValue());
+       // QFJ-533
+        int declaredGroupCount = 0;
+        try {
+          declaredGroupCount = Integer.parseInt(field.getValue());
+        } catch (final NumberFormatException e) {
+            throw new InvalidMessage("NumberFormatException: Repeating group count requires an Integer but found:" +field.getValue());
```
philipwhiuk: Instead of making "NumberFormatException" part of the string, you should provide it as the cause.

pbyrne413: good point, I'll amend the PRs

chrjohn:
Hi, you don't need to open PRs against the 1.6.x branch. There will be no more 1.6.x releases. Thank you.

PR #152, opened 2017-11-13
After scheduled logout/disconnection FIX session starts connection attempts from the 1st host on the list

mmitya:
**Example**
Session configuration snippet:
```
SocketConnectHost=host_0
SocketConnectPort=port_0
SocketConnectHost1=host_1
SocketConnectPort1=port_1
StartTime=07:30:00
EndTime=17:30:00
```
**Issue**
FIX session establishes connection to host_0 at 07:30:00
FIX session disconnects at 17:30:00
Next day FIX session established connection to host_1 at 07:30:00, but it is expected to be host_0

For my own reference: PROM-4719

### DIFF
--- a/quickfixj-core/src/main/doc/usermanual/usage/configuration.html
+++ b/quickfixj-core/src/main/doc/usermanual/usage/configuration.html
@@ -498,11 +498,29 @@
     <TD>"TCP" </TD>
   </TR>
   <TR ALIGN="left" VALIGN="middle">
+    <TD> <I>SocketConnectPort&lt;n&gt;</I> </TD>
+    <TD> Alternate socket port(s) for connecting to a session for failover or load balancing,
+         where <B>n</B> is a positive integer, i.e. SocketConnectPort1, SocketConnectPort2, etc.
+         Must be consecutive and have a matching SocketConnectHost&lt;n&gt;
+    </TD>
+    <TD> positive integer </TD>
+    <TD>&nbsp; </TD>
+  </TR>
+  <TR ALIGN="left" VALIGN="middle">
 
     <TD> <I>SocketConnectHost&lt;n&gt;</I> </TD>
-    <TD> Alternate socket hosts for connecting to a session for failover, where <B>n</B> is a
-         positive integer. (i.e.) SocketConnectHost1, SocketConnectHost2... must be consecutive
-         and have a matching SocketConnectPort[n]</TD>
+    <TD> Alternate socket host(s) for connecting to a session for failover or load balancing,
+         where <B>n</B> is a positive integer, i.e. SocketConnectHost1, SocketConnectHost2, etc.
+         Must be consecutive and have a matching SocketConnectPort&lt;n&gt;
+         <p>
+      Connection list iteration rules:
+      <ul>
+        <li>Connections are tried one after another until one is successful: SocketConnectHost:SocketConnectPort,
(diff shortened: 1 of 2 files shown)


chrjohn:
Thanks for the PR, seems sensible.

chrjohn:
One further thought: wouldn't this also reset the address back to the first address in the following case:
E.g. we have 3 addresses:

* Connection to 1 fails.
* Connection to 2 succeeds and sets the address back to the first address upon successful connection. When connection 2 is dropped, I think that with this PR connection is initiated to address 1 but I would expect 3.

I didn't look this up in the code, just was thinking about it.

mmitya:
@chrjohn , yes, you are right - it will try FIX connection to 1 first.
But I'm not sure this is wrong.
What if host 1 was not available, but is available now?

In my case host 1 is primary, host 2 is backup. QFJ connects to primary (host 1) on Mon, but on Tue it connects to backup (host 2). If primary host 1 is not available - it doesn't mean we shouldn't try it again next day.

chrjohn:
Hi @mmitya 
but doesn't this also happen during the day? E.g. if connection 2 disconnects during the day, shouldn't connection 3 be taken next instead of 1? I would expect that it cycles through the connections intra-day.

mmitya:
Good afternoon @chrjohn ,
I don't think there is a single answer to this question.
I think it should be OK to cycle from the 1st host after successful FIX connection.

chrjohn:
You are probably right ;) I'll give it a thought or two.

mmitya:
Great, thanks @chrjohn .

mmitya:
Good afternoon, @chrjohn.
Just decided to summarize my proposal.

Connection algorithm:
0. Wait for session has to connect.
1. Connect to host_1 : port_1, success - goto 0, failure - goto 2.
2. Connect to host_2 : port_2, success - goto 0, failure - goto 3.
...
N. Connect to host_N : port_N, success - goto 0, failure - goto 1.

Real case example:
- host_1 - primary server (cluster)
- host_2 - back-up server (cluster)

If host_1 is not available at some point we should try to connect it again next time.

After every successful FIX connection and subsequent disconnection FIX engine starts from the beginning of the host list.

Hope it was helpful.

P.S. I don't think this connection establishment logic can satisfy everyone. It would be better to extract interface and have different implementations - round-robin for LB, etc. But it is a separate feature, not a quick change.

chrjohn:
Hi @mmitya , thanks for the summary. Would it be possible to add a short description of the algorithm to https://github.com/quickfix-j/quickfixj/blob/master/quickfixj-core/src/main/doc/usermanual/usage/configuration.html ??
I think something along the lines of:
* connections are tried one after another until one is successful
* next connection attempt after a successful connection will start at first defined connection again
Thanks,
Chris.

mmitya:
Hello @chrjohn ,
I update the configuration page, please have a look when you have time.
TA.

chrjohn:
Hi @mmitya ,
thanks for the update and for adding SocketConnectPort<n> :)
Will merge once 2.0.0 has been released.

Thanks,
Chris.

mmitya:
Hello @chrjohn !
Great news, thanks.
When do you plan to release 2.0.1?

chrjohn:
Hi @mmitya , there is currently no plan. Luckily I got 2.0.0 out of the door before Christmas. ;)

PR #160, opened 2017-11-28
minor changes around unit tests

chrjohn:
- AbstractSocketInitiator: stop initiators before clearing map
 - whitespace changes
 - minor simplifications

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/mina/initiator/AbstractSocketInitiator.java
+++ b/quickfixj-core/src/main/java/quickfix/mina/initiator/AbstractSocketInitiator.java
@@ -46,6 +46,7 @@ import quickfix.mina.ssl.SSLSupport;
 import java.net.SocketAddress;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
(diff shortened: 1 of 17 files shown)


philipwhiuk:
I'm fine with this change, but the commit message doesn't explain that you're actually changing app code. In this case you're actually changing behaviour.

Having a commit that claims to only change test code changing app code is annoying when working out what's changed.

quickfixj-core/src/main/java/quickfix/mina/initiator/AbstractSocketInitiator.java
```diff
@@ -79,6 +79,9 @@ protected void createSessionInitiators()
         try {
             // QFJ698: clear() is needed on restart, otherwise the set gets filled up with
```
philipwhiuk: This comment is now no longer attached to the right line

quickfixj-core/src/main/java/quickfix/mina/initiator/AbstractSocketInitiator.java
```diff
@@ -79,6 +79,9 @@ protected void createSessionInitiators()
...
             // more and more initiators which are not equal because the local port differs
+            initiators.forEach((initiator) -> {
```
philipwhiuk: It would be nice to test this change.

quickfixj-core/src/main/java/quickfix/mina/initiator/IoSessionInitiator.java
```diff
@@ -338,6 +338,9 @@ private void resetIoConnector() {
                         connectFuture.cancel();
                     }
                     connectFuture = null;
+                    if ( !ioSession.isClosing()) {
```
philipwhiuk: Ditto testing.

chrjohn:
Hi @philipwhiuk , thanks for your comments.
Maybe I should have been more expressive. Actually I am investigating some failing unit tests that only seem to fail on the build server. So the easiest way for me is to create some PRs and have them built by Travis.
I'll see which changes will be merged at all.
Cheers,
Chris.

chrjohn:
This is shit. After enabling SSL debug logging to see why the tests fail, there are no more SSL errors.

PR #161, opened 2017-11-30
Pass SessionID to MemoryStore constructor from MemoryStoreFactory

jonfreedman:
Looking at the code the only change in behavior based on this will be that the "memory store does not support refresh!" error generated if #refresh is called will be logged at the session level.

The upside is it will make debugging a lot easier.


jonfreedman:
I think the TravisCI build timed out rather than failed

PR #163, opened 2017-12-04
Add full removeGroup method

philipwhiuk:
Add method that takes a boolean to also remove the count.

As per conversation on QFJ mailing list

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/FieldMap.java
+++ b/quickfixj-core/src/main/java/quickfix/FieldMap.java
@@ -614,6 +614,16 @@ public abstract class FieldMap implements Serializable {
         getGroups(field).clear();
         removeField(field);
     }
+    
+    public void removeGroup(int field, boolean removeCount) {
+        if(!removeCount) {
+            removeGroup(field);
+        } else {
+            getGroups(field).clear();
+            groups.remove(new Integer(field));
+            removeField(field);
+        }
+    }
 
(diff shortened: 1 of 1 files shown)


chrjohn:
Hi @philipwhiuk ,
thanks for the PR. Could you please add a tiny test? ;) I do not recall from which discussion this was. Was it "Message.getGroups() may mutate message instance"?
Thanks,
Chris.

philipwhiuk:
It was yes - I should have time to update this today :+1:

chrjohn:
Hi @philipwhiuk ,

I followed this up today. Given your comment on the mailing list thread I was trying to reproduce it.

> 
> So in the below example:
> 
> > System.out.println(msg.hasGroup(1));
> > System.out.println(msg.getGroups(1));
> > System.out.println(msg.hasGroup(1));
> 
> actually adds a tag, |1=0| to the message.

I could not reproduce that. Here is the output:

```
false
[]
true
9=010=167
```

The second call to `hasGroup()` returns true, but the group does not occur in the `toString()` output.
I don't know, maybe this was caused by the latest changes but I'm not sure.

Where do we go from here?

Thanks,
Chris.

philipwhiuk:
I'll have a look and try and diagnose this some time this week - sorry I've not got back to this.

philipwhiuk:
Closing in favour of #414

PR #165, opened 2017-12-11
Add new header field to header list: ApplExtID

philipwhiuk:
Header field present in FIX spec but missing in QFJ

See: http://fiximate.fixtrading.org/latestEP/en/FIX.5.0SP2_EP236/tag1156.html


philipwhiuk:
I'm unclear why Travis failed, I don't think it's related.

david-gibbs-ig:
I see the same error on another pull request, the tests seem to be unreliable.

PR #168, opened 2018-01-28
QFJ-943: optional watermarks-based back pressure propagation from inbo…

mrbald:
The background for this change is described in the QFJ-943

To make sure writes to the inbound FIX Message queue don't block the MINA IO loop a blocking queue can be replaced with watermarks-based (configured with lower and upper watermark values instead of fixed queue capacity.

When using watermarks the queue is no longer limited in size.
Instead, when queue grows above the upper watermark, the socket reads of the respective session are suspended by unsubscribing from reads with MINA selector (IoSession.suspendRead()), nicely propagating back pressure onto the remote end of the TCP session by the TCP flow control.

When the queue drains below the lower watermark the socket is resubscribed for reads (IoSession.resumeRead()and the back pressure is relieved.

The functionality is implemented with a standalone class WatermarkTracker.

Watermarks - enabled queue in SPSC scenario shows somewhat near 300ns enqueue/dequeue times, which should not be a problem considering general latencies of QuickFIXj, so I left it with JDK collections.

The implementation is sandboxed / benchmarked in https://github.com/mrbald/ufwj. Additionally, the threaded initiator tested under load in a prod-like environment

NB: watermarks do not save the day if one side of the fix session outperforms the other side on the average. If socket reads get disabled for long enough the receiver starts missing heartbeats and eventually drops the link.

### DIFF
--- a/quickfixj-core/pom.xml
+++ b/quickfixj-core/pom.xml
@@ -401,24 +401,6 @@
                     </systemPropertyVariables>
                 </configuration>
             </plugin>
-            <plugin>
-                <groupId>org.codehaus.mojo</groupId>
-                <artifactId>build-helper-maven-plugin</artifactId>
-                <version>1.9.1</version>
-                <executions>
-                    <execution>
-                        <phase>generate-sources</phase>
-                        <goals>
-                            <goal>add-source</goal>
-                        </goals>
-                        <configuration>
-                            <sources>
-                                <source>${project.build.directory}/generated-sources</source>
-                            </sources>
-                        </configuration>
-                    </execution>
-                </executions>
-            </plugin>
         </plugins>
     </build>
 
--- a/quickfixj-core/pom.xml
+++ b/quickfixj-core/pom.xml
@@ -401,6 +401,24 @@
                     </systemPropertyVariables>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>build-helper-maven-plugin</artifactId>
+                <version>1.9.1</version>
+                <executions>
+                    <execution>
+                        <phase>generate-sources</phase>
+                        <goals>
+                            <goal>add-source</goal>
+                        </goals>
+                        <configuration>
+                            <sources>
+                                <source>${project.build.directory}/generated-sources</source>
+                            </sources>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
         </plugins>
     </build>
 
--- a/quickfixj-core/src/main/java/quickfix/mina/SingleThreadedEventHandlingStrategy.java
+++ b/quickfixj-core/src/main/java/quickfix/mina/SingleThreadedEventHandlingStrategy.java
@@ -33,6 +33,7 @@ import java.util.concurrent.Executor;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
 
+import static java.lang.String.format;
 import static quickfix.mina.EventHandlingStrategy.lookupIoSession;
 
 /**
@@ -65,16 +66,14 @@ public class SingleThreadedEventHandlingStrategy implements EventHandlingStrateg
                     final IoSession ioSession = lookupIoSession(qfSession);
                     if (ioSession != null && ioSession.isReadSuspended()) {
                         ioSession.resumeRead();
-                        LOG.info("{}: inbound queue size < lower watermark ({}), socket reads resumed",
-                                qfSession.getSessionID(), queueLowerWatermark);
+                        qfSession.getLog().onEvent(format(LOWER_WATERMARK_FMT, queueLowerWatermark));
                     }
                 },
                 qfSession -> { // upper watermark crossed up, while reads active
                     final IoSession ioSession = lookupIoSession(qfSession);
                     if (ioSession != null && !ioSession.isReadSuspended()) {
                         ioSession.suspendRead();
-                        LOG.info("{}: inbound queue size > upper watermark ({}), socket reads suspended",
-                                qfSession.getSessionID(), queueUpperWatermark);
+                        qfSession.getLog().onEvent(format(UPPER_WATERMARK_FMT, queueUpperWatermark));
                     }
                 });
     }
--- a/quickfixj-core/src/main/java/quickfix/mina/SingleThreadedEventHandlingStrategy.java
+++ b/quickfixj-core/src/main/java/quickfix/mina/SingleThreadedEventHandlingStrategy.java
@@ -20,9 +20,6 @@
 
 package quickfix.mina;
 
-import org.apache.mina.core.session.IoSession;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import quickfix.*;
 
 import java.util.ArrayList;
@@ -33,15 +30,13 @@ import java.util.concurrent.Executor;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
 
-import static java.lang.String.format;
-import static quickfix.mina.EventHandlingStrategy.lookupIoSession;
+import static quickfix.mina.QueueTrackers.newDefaultQueueTracker;
+import static quickfix.mina.QueueTrackers.newMultiSessionWatermarkTracker;
 
 /**
  * Processes messages for all sessions in a single thread.
  */
 public class SingleThreadedEventHandlingStrategy implements EventHandlingStrategy {
-    private static final Logger LOG = LoggerFactory.getLogger(EventHandlingStrategy.class);
-
     public static final String MESSAGE_PROCESSOR_THREAD_NAME = "QFJ Message Processor";
     private final BlockingQueue<SessionMessageEvent> eventQueue;
     private final QueueTracker<SessionMessageEvent> queueTracker;
@@ -54,28 +49,14 @@ public class SingleThreadedEventHandlingStrategy implements EventHandlingStrateg
     public SingleThreadedEventHandlingStrategy(SessionConnector connector, int queueCapacity) {
         sessionConnector = connector;
         eventQueue = new LinkedBlockingQueue<>(queueCapacity);
-        queueTracker = QueueTracker.wrap(eventQueue);
+        queueTracker = newDefaultQueueTracker(eventQueue);
     }
 
     public SingleThreadedEventHandlingStrategy(SessionConnector connector, int queueLowerWatermark, int queueUpperWatermark) {
         sessionConnector = connector;
         eventQueue = new LinkedBlockingQueue<>();
-        queueTracker = WatermarkTracker.newMulti(eventQueue, queueLowerWatermark, queueUpperWatermark,
-                evt -> evt.quickfixSession,
-                qfSession -> { // lower watermark crossed down, while reads suspended
-                    final IoSession ioSession = lookupIoSession(qfSession);
-                    if (ioSession != null && ioSession.isReadSuspended()) {
-                        ioSession.resumeRead();
-                        qfSession.getLog().onEvent(format(LOWER_WATERMARK_FMT, queueLowerWatermark));
-                    }
-                },
-                qfSession -> { // upper watermark crossed up, while reads active
-                    final IoSession ioSession = lookupIoSession(qfSession);
-                    if (ioSession != null && !ioSession.isReadSuspended()) {
-                        ioSession.suspendRead();
-                        qfSession.getLog().onEvent(format(UPPER_WATERMARK_FMT, queueUpperWatermark));
-                    }
-                });
+        queueTracker = newMultiSessionWatermarkTracker(eventQueue, queueLowerWatermark, queueUpperWatermark,
+                evt -> evt.quickfixSession);
     }
 
     public void setExecutor(Executor executor) {
--- a/quickfixj-core/src/main/java/quickfix/mina/SingleThreadedEventHandlingStrategy.java
+++ b/quickfixj-core/src/main/java/quickfix/mina/SingleThreadedEventHandlingStrategy.java
@@ -20,11 +20,10 @@
 
 package quickfix.mina;
 
-import quickfix.LogUtil;
-import quickfix.Message;
-import quickfix.Session;
-import quickfix.SessionID;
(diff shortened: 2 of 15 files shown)


mrbald:
When using watermarks the queue is no longer limited in size.
Instead, when queue grows above the upper watermark, the socket reads of the respective session are suspended by unsubscribing from reads with MINA selector  (`IoSession.suspendRead()`), nicely propagating back pressure onto the remote end of the TCP session by the TCP flow control.

When the queue drains below the lower watermark the socket is resubscribed for reads (`IoSession.resumeRead()`and the back pressure is relieved.

The functionality is implemented with a standalone class `WatermarkTracker`.

Watermarks - enabled queue in SPSC scenario shows somewhat near 300ns enqueue/dequeue times, which should not be a problem considering general latencies of QuickFIXj, so I left it with JDK collections

chrjohn:
Hi @mrbald , thanks for the PR and the benchmarks you did.
Will take a more thorough look at it soon.
Cheers
Chris.

quickfixj-core/src/main/java/quickfix/mina/WatermarkTracker.java
```diff
@@ -0,0 +1,185 @@
...
+            Consumer<S> onLowerWatermarkCrossed, Consumer<S> onUpperWatermarkCrossed) {
+        return new WatermarkTracker<>(queue, lowerWatermark, upperWatermark, classifier, onLowerWatermarkCrossed, onUpperWatermarkCrossed);
+    }
+
+    private WatermarkTracker(
+            BlockingQueue<E> queue,
+            long lowerWatermark, long upperWatermark,
+            Function<E, S> classifier,
+            Consumer<S> onLowerWatermarkCrossed, Consumer<S> onUpperWatermarkCrossed) {
+        assert lowerWatermark >= 0 && lowerWatermark < upperWatermark;
```
chrjohn: I am not a fan of assertions in non-test code. Assertions are disabled by default anyway. Could you please remove them? Thanks

quickfixj-core/src/main/java/quickfix/mina/WatermarkTracker.java
```diff
@@ -0,0 +1,185 @@
...
+        final Map<S, StreamTracker> trackerMap = new ConcurrentHashMap<>();
+
+        this.trackerSupplier = key -> trackerMap.computeIfAbsent(key, StreamTracker::new);
+    }
+
+    private WatermarkTracker(
+            BlockingQueue<E> queue,
+            long lowerWatermark, long upperWatermark,
+            Runnable onLowerWatermarkCrossed, Runnable onUpperWatermarkCrossed) {
+        assert lowerWatermark >= 0 && lowerWatermark < upperWatermark;
```
chrjohn: I am not a fan of assertions in non-test code. Assertions are disabled by default anyway. Could you please remove them? Thanks

quickfixj-core/src/main/java/quickfix/mina/WatermarkTracker.java
```diff
@@ -0,0 +1,185 @@
...
+ * PARTICULAR PURPOSE.
+ *
+ * See http://www.quickfixengine.org/LICENSE for licensing information.
+ *
+ * Contact ask@quickfixengine.org if any conditions of this licensing
+ * are not clear to you.
+ ******************************************************************************/
+
+/*
+ * Copyright (c) 2018 Vladimir Lysyy (mrbald@github)
```
chrjohn: I am not sure if we can mix different licenses and copyrights here. Will have to check.

quickfixj-core/src/main/java/quickfix/mina/SingleThreadedEventHandlingStrategy.java
```diff
@@ -49,6 +51,39 @@
...
+                qfSession -> { // upper watermark crossed up, while reads active
+                    final IoSession ioSession = lookupIoSession(qfSession);
+                    if (ioSession != null && !ioSession.isReadSuspended()) {
+                        ioSession.suspendRead();
+                        LOG.info("{}: inbound queue size > upper watermark ({}), socket reads suspended",
+                                qfSession.getSessionID(), queueUpperWatermark);
+                    }
+                });
+    }
+    private static IoSession lookupIoSession(Session qfSession) {
```
chrjohn: There is a similar method in the other EventHandlingStrategy. Can this be moved to the super-interface maybe?

quickfixj-core/src/main/java/quickfix/mina/SingleThreadedEventHandlingStrategy.java
```diff
@@ -49,6 +51,39 @@
...
+    public SingleThreadedEventHandlingStrategy(SessionConnector connector, int queueLowerWatermark, int queueUpperWatermark) {
+        sessionConnector = connector;
+        eventQueue = new LinkedBlockingQueue<>();
+        queueTracker = WatermarkTracker.newMulti(eventQueue, queueLowerWatermark, queueUpperWatermark,
+                evt -> evt.quickfixSession,
+                qfSession -> { // lower watermark crossed down, while reads suspended
+                    final IoSession ioSession = lookupIoSession(qfSession);
+                    if (ioSession != null && ioSession.isReadSuspended()) {
+                        ioSession.resumeRead();
+                        LOG.info("{}: inbound queue size < lower watermark ({}), socket reads resumed",
```
chrjohn: IMHO this should be DEBUG logging.

quickfixj-core/src/main/java/quickfix/mina/SingleThreadedEventHandlingStrategy.java
```diff
@@ -49,6 +51,39 @@
...
+                                qfSession.getSessionID(), queueLowerWatermark);
+                    }
+                },
+                qfSession -> { // upper watermark crossed up, while reads active
+                    final IoSession ioSession = lookupIoSession(qfSession);
+                    if (ioSession != null && !ioSession.isReadSuspended()) {
+                        ioSession.suspendRead();
+                        LOG.info("{}: inbound queue size > upper watermark ({}), socket reads suspended",
```
chrjohn: IMHO this should be DEBUG logging.

quickfixj-core/pom.xml
```diff
@@ -401,6 +401,24 @@
...
+                <version>1.9.1</version>
+                <executions>
+                    <execution>
+                        <phase>generate-sources</phase>
+                        <goals>
+                            <goal>add-source</goal>
+                        </goals>
+                        <configuration>
+                            <sources>
+                                <source>${project.build.directory}/generated-sources</source>
```
chrjohn: Does this add all the generated source files to the JAR file? Is this needed? They are generated on a fresh build anyway.

mrbald: this helps IntelliJ automatically detect the generated files (this was the reason), if they also get added to the *-sources.jar I'd love it too as generated or not, sometimes it's nice to click down to sources to see what they do.

Having said the above - if there are reservations about this change, let's drop it.

quickfixj-core/src/main/java/quickfix/mina/SingleThreadedEventHandlingStrategy.java
mrbald: They are sign of a slow consumer - potentially need to be picked up by prod monitoring and actioned upon (at least in the last 3 places I was engaged with, and I think it's correct), therefore definitely not debug. Also if they are logged frequently (up/down/up/down) it is a sign that they should be reconfigured. But I understand there may be established rules for what to log when in QuickFIX, so happy to change. Let me know.

quickfixj-core/src/main/java/quickfix/mina/SingleThreadedEventHandlingStrategy.java
mrbald: They are sign of a slow consumer - potentially need to be picked up by prod monitoring and actioned upon (at least in the last 3 places I was engaged with, and I think it's correct), therefore definitely not debug. Also if they are logged frequently (up/down/up/down) it is a sign that they should be reconfigured. But I understand there may be established rules for what to log when in QuickFIX, so happy to change. Let me know.

quickfixj-core/src/main/java/quickfix/mina/WatermarkTracker.java
mrbald: Happy to "donate" it to any license and strip this comment. Will include in the next commit.

quickfixj-core/src/main/java/quickfix/mina/WatermarkTracker.java
mrbald: It's more of a documenting assertion (as you said, they are disabled in prod). I'll remove in the next commit.

quickfixj-core/src/main/java/quickfix/mina/WatermarkTracker.java
mrbald: It's more of a documenting assertion (as you said, they are disabled in prod). I'll remove in the next commit.

quickfixj-core/src/main/java/quickfix/mina/SingleThreadedEventHandlingStrategy.java
mrbald: should be possible. objections against static final methods on interfaces?

mrbald:
@chrjohn - pushed the requested changes

chrjohn:
Hi @mrbald , there will not be any 1.6.x release anymore. Even this change is probably more a 2.1.x change since it introduces new functionality.

chrjohn:
Hi @mrbald ,
regarding the benchmarks. How do I run them? I tried the usual JMH way with building the JAR and then doing `java -jar <benchmarkfile>` but it did not work.
Thanks,
Chris.

mrbald:
Hi @chrjohn,

Benchmarks - I just run them from the IDE - open the benchmark source code (say [net/bobah/ufwj/queue/WatermarkTrackerBenchmark.java](https://github.com/mrbald/ufwj/blob/master/base-benckmarks/src/main/java/net/bobah/ufwj/queue/WatermarkTrackerBenchmark.java), and "Run as Java Application...". It forks the JVM, so the fact or running from the IDE should not disturb numbers.

Importantly, JMH does code generation during compilation so the IDE should be made to do the same when building (IntelliJ does if the repo is imported as Maven project).

It should be possible to run the same class from the command line, or with an appropriate maven plugin as well, just that I haven't done it that way in that repo.

chrjohn:
Hi @mrbald ,
I use Netbeans but I had to change the POM for it to work but looking good now.
Thanks

mrbald:
@chrjohn, logging updated, please review

quickfixj-core/src/main/java/quickfix/mina/SingleThreadedEventHandlingStrategy.java
```diff
@@ -34,12 +33,18 @@
...
 import java.util.concurrent.TimeUnit;
 
+import static java.lang.String.format;
+import static quickfix.mina.EventHandlingStrategy.lookupIoSession;
+
 /**
  * Processes messages for all sessions in a single thread.
  */
 public class SingleThreadedEventHandlingStrategy implements EventHandlingStrategy {
+    private static final Logger LOG = LoggerFactory.getLogger(EventHandlingStrategy.class);
```
chrjohn: LOG shouldn't be required anymore now.

quickfixj-core/src/main/java/quickfix/mina/ThreadPerSessionEventHandlingStrategy.java
```diff
@@ -35,19 +35,34 @@
...
 import java.util.concurrent.TimeUnit;
 
+import static java.lang.String.format;
+import static quickfix.mina.EventHandlingStrategy.lookupIoSession;
+
 /**
  * Processes messages in a session-specific thread.
  */
 public class ThreadPerSessionEventHandlingStrategy implements EventHandlingStrategy {
+    private static final Logger LOG = LoggerFactory.getLogger(EventHandlingStrategy.class);
```
chrjohn: LOG shouldn't be required anymore now.

quickfixj-core/src/main/java/quickfix/mina/SingleThreadedEventHandlingStrategy.java
```diff
@@ -49,6 +54,28 @@
...
+        queueTracker = QueueTracker.wrap(eventQueue);
+    }
+
+    public SingleThreadedEventHandlingStrategy(SessionConnector connector, int queueLowerWatermark, int queueUpperWatermark) {
+        sessionConnector = connector;
+        eventQueue = new LinkedBlockingQueue<>();
+        queueTracker = WatermarkTracker.newMulti(eventQueue, queueLowerWatermark, queueUpperWatermark,
+                evt -> evt.quickfixSession,
+                qfSession -> { // lower watermark crossed down, while reads suspended
+                    final IoSession ioSession = lookupIoSession(qfSession);
```
chrjohn: The onWatermarkCrossed actions are duplicated in SingleThreadedEventHandlingStrategy and ThreadPerSessionEventHandlingStrategy. Can this be unified somehow?

chrjohn:
Hi @mrbald ,
I had to add the following to the benchmark POM:
 ```
<plugin>
                <groupId>org.codehaus.mojo</groupId>
                <artifactId>exec-maven-plugin</artifactId>
                <executions>
                    <execution>
                        <id>run-benchmarks</id>
                        <phase>integration-test</phase>
                        <goals>
                            <goal>exec</goal>
                        </goals>
                        <configuration>
                            <classpathScope>test</classpathScope>
                            <executable>java</executable>
                            <arguments>
                                <argument>-classpath</argument>
                                <classpath />
                                <argument>org.openjdk.jmh.Main</argument>
                                <argument>.*</argument>
                            </arguments>
                        </configuration>
                    </execution>
                </executions>
            </plugin>
```

This ran the benchmarks on build. I could then also run the benchmarks alone because the classes and information in META-INF were generated. I had the following error without these changes:
`Unable to find the resource: /META-INF/BenchmarkList`

Another question: which one of the benchmarks is the one where the "threaded initiator tested under load in a prod-like environment" (as mentioned in description of this PR)?

Thanks in advance!

chrjohn:
Thanks @mrbald

PR #170, opened 2018-01-31
do not throw IncorrectDataFormat exception if value is empty and ValidateFieldsHaveValues=N

jonfreedman:
Some vendors use the same repeating groups in different parts of a message conditional on the value of other fields e.g. different product types


jonfreedman:
This PR has picked up a second change I have which deals with the unfortunate situation where a vendor sends a blank value for a timestamp field.  Commit: https://github.com/jonfreedman/quickfixj/commit/4ea7e2a6a2ce7db7d95ff8fae9d24c50b8a445c8

chrjohn:
Hi @jonfreedman ,

thanks for the PR. I'm not very familiar with XSL. Could you please shortly describe how the generated code changes? Maybe you can attach an example class before/after the change.

Thanks,
Chris.

jonfreedman:
I started writing some tests around validating the output .java files compile and it seems that my change doesn't do what I want, and there would be quite a bit of work needed to make the code generator behave how I wanted.  I'll remove the xslt change, although it probably has no ill effects.

chrjohn:
Hi @jonfreedman ,
if it's not too much to ask: would it be possible to create a separate PR for the "value is empty" case?
Thanks,
Chris.

jonfreedman:
I've created https://github.com/quickfix-j/quickfixj/pull/207 & https://github.com/quickfix-j/quickfixj/pull/208

Closing this

chrjohn:
Thanks @jonfreedman ,
I will add #207 to the next release due to be released in a few days.
I will have a look at #208 later.
Hope that is OK with you.

Cheers,
Chris.

jonfreedman:
Yep makes sense, with #208 have a think if the core module is the right place for it, could be useful to show people how to write their own dictionaries, but that's getting very niche.

PR #171, opened 2018-02-01
QFJ-888: DefaultMessageFactory does not support FIX5.0 SP1/2

ckolek:
The background for this change is described in [QFJ-888](https://www.quickfixj.org/jira/browse/QFJ-888).

Changes are as follows:
1. Added a new create method to `MessageFactory` which accepts an `ApplVerID`. The default behavior of the new method is to call `MessageFactory.create(beginString, msgType)` in order to maintain backwards compatibility.
2. Modified `MessageUtils.parse` to call the new create method, passing in the `ApplVerID` acquired from the incoming message or the session. Other calls to `MessageFactory.create` have remained as-is.
3. Added a new constructor to `DefaultMessageFactory` which accepts an `ApplVerID` value to be used as the default when `create(beginString, msgType)` or `create(beginString, null, msgType)` is called. The existing constructor calls the new one, passing in `ApplVerID.FIX50` in order to maintain backwards compatibility.
4. Implementation of message creation in `DefaultMessageFactory` has been moved to the new create method. The old create method calls `create(beginString, null, msgType)`. In the case where `applVerID` is null, the `defaultApplVerID` is used instead.

Ultimately, these changes provide the following:
1. Developers who use `DefaultMessageFactory` can now specify a default `ApplVerID` to be used for FIXT message construction.
2. FIXT message construction is now done dynamically according to the `ApplVerID` in the incoming message/session.

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/DefaultMessageFactory.java
+++ b/quickfixj-core/src/main/java/quickfix/DefaultMessageFactory.java
@@ -19,9 +19,11 @@
 
 package quickfix;
 
+import quickfix.field.ApplVerID;
 import quickfix.field.MsgType;
 
 import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 
 import static quickfix.FixVersions.BEGINSTRING_FIX40;
@@ -40,6 +42,8 @@ import static quickfix.FixVersions.FIX50SP2;
 public class DefaultMessageFactory implements MessageFactory {
     private final Map<String, MessageFactory> messageFactories = new ConcurrentHashMap<>();
 
+    private final ApplVerID defaultApplVerID;
+
     /**
      * Constructs a DefaultMessageFactory, which dynamically loads and delegates to
      * the default version-specific message factories, if they are available at runtime.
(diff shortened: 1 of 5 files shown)


chrjohn:
Hi @ckolek ,
thanks for the PR. Looks good to me at first glance, will look at it more closely later.
One thing: there are two acceptance tests failing because of this change (the SSL related tests can be ignored for now):

```
[ERROR] fix50/2r_UnregisteredMsgType.def(quickfix.test.acceptance.AcceptanceTestSuite$AcceptanceTest)  Time elapsed: 5.221 s  <<< FAILURE!
org.junit.ComparisonFailure: wrong msg type expected:<[j]> but was:<[0]>
	at quickfix.test.acceptance.AcceptanceTestSuite$AcceptanceTest.run(AcceptanceTestSuite.java:78)
[ERROR] fix50/8_AdminAndApplicationMessages-FIX50SP2.def(quickfix.test.acceptance.AcceptanceTestSuite$AcceptanceTest)  Time elapsed: 5.95 s  <<< FAILURE!
org.junit.ComparisonFailure: wrong msg type expected:<[D]> but was:<[0]>
	at quickfix.test.acceptance.AcceptanceTestSuite$AcceptanceTest.run(AcceptanceTestSuite.java:78)
```
It looks like the session is only heartbeating instead of sending the expected messages.
If you need help fixing this please let me know so I can take a closer look. 

Cheers,
Chris.

ckolek:
I made the mistake of passing in the FIX50 BeginString rather than the FIX50 ApplVerID in the DefaultMessageFactory default constructor. I'll update the PR.

PR #172, opened 2018-02-02
QFJ-940 - Event log omits parts of received message in case of validation errors

chrjohn:
- added method Message.toRawString() which returns the message string as it was passed to the Message class
- in case of errors: log raw message data instead of incompletely parsed message
- added unit test
- improved some log messages in Session class


philipwhiuk:
This is awesome @chrjohn - I had a plan to add an `Application` callback when a message was unable to be properly parsed (i.e. `getException() != null`) but toRawString probably removes the need for this - and even if so this is obviously usable by it!

chrjohn:
Hi Philip,

I hope it is still awesome although your comment was deleted in the meantime. ;)
Do you think it is feasible to include this into 2.0.1? It solves a bug but I had to add a new 
public method to the Message class.

Thanks,
Chris.

philipwhiuk:
It's a pretty minor method as it's just exposing an existing various variable via a getter. I think it's okay for a patch release. :+1:

PR #173, opened 2018-02-08
Minor logging change and correction of and-clause in unit test

chrjohn:


quickfixj-core/src/test/java/quickfix/mina/acceptor/AcceptorIoHandlerTest.java
```diff
@@ -198,7 +198,7 @@ public void testLogonWithoutHeartBtInt() throws Exception {
         try {
             handler.processMessage(mockIoSession, message);
         } catch (Exception e) {
-            fail("No exception should be thrown!");
+            fail("No exception should be thrown! " + e);
```
alphafoobar: Hi @chrjohn - A little late and a minor point, but there isn't any need to catch an `Exception` here. If the exception is thrown the unit test will `fail` and by not catching the exception you will benefit from seeing the full stacktrace.

Cheers!

chrjohn: Hi @alphafoobar , you are of course right.
That happens when you don't think about what you're doing. ;)

Thanks,
Chris.

PR #174, opened 2018-02-15
Some calls to isSetting() don't specify the sessionID

jasonps01:
I discovered this bug while trying to set different JDBC tables for different sessions.


chrjohn:
Thanks for the PR. Good catch!

PR #177, opened 2018-03-07
Update MINA to 2.0.17

chrjohn:
to be merged when it has been relased

Then also https://www.quickfixj.org/jira/browse/QFJ-937 (MaxScheduledWriteRequests is ineffective) can be closed.

PR #180, opened 2018-03-12
QFJ-932 - data dictionary validation format independent

the-thing:
Data dictionary validation is format independent now. The following tags are not allowed to be empty (with exception to \<header/> and \<trailer/> for FIX version 5.0 and higher).

- \<header/>
- \<trailer/>
- \<fields/>
- \<message/>
- \<messages/>

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/DataDictionary.java
+++ b/quickfixj-core/src/main/java/quickfix/DataDictionary.java
@@ -812,6 +812,18 @@ public class DataDictionary {
         }
     }
 
+    private int countElementNodes(NodeList nodes) {
+        int elementNodesCount = 0;
+
+        for (int i = 0; i < nodes.getLength(); i++) {
+            if (nodes.item(i).getNodeType() == Node.ELEMENT_NODE) {
+                elementNodesCount++;
+            }
+        }
+
+        return elementNodesCount;
+    }
+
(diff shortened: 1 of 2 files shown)


chrjohn:
Hi @the-thing , thanks for the PR. There is a NPE in the SSLCertificateTest which should be unrelated but I also saw this once in my local environment.
The other error can be ignored, I assume it was because of the session reset at midnight.

the-thing:
Hi Christoph,

It runs on my container ;)
https://travis-ci.org/the-thing/quickfixj/builds/352590390

I doubt that any of these test failures are related to this change. It's possible that there are still flaky tests in many areas. For example I often get errors on master branch for this particular test case

[ERROR] Failures: 
[ERROR]   SocketInitiatorTest.testInitiatorStopStartFileLog:184->doTestOfRestart:393->assertLoggedOn:488 Expected logon did not occur

https://travis-ci.org/the-thing/quickfixj/builds/352514755

I will try to debug some of these soon.

chrjohn:
Hi Marcin,

OK, for further analysis I have activated the complete stacktrace in surefire, just in case it occurs again.

W.r.t. the SocketInitiatorTest: it could be that this will be solved with #177 and/or #178 .

Thanks,
Chris.

PR #185, opened 2018-04-03
Code cleanup

Ganet:
Simplify if statements.
Add final keyword where possible.
Remove redundant null checks.
Convert class fields to local variables were it is possible.
Removed unused imports.
Converted JUnit3 old TestCase test implementation to JUnit 4.
Simplify junit assertions from assertEquals to assertTrue/assertFalse/fail and more.	 
Use lambda in simple anonymous classes for test (LogFactory).
Replace long if else statements in DataDictionary with switch for improved performance (switch for strings use internal hashtable).
Replace StringBuilder usage with simple string concatenation in DataDictionary so simplify code whilst keeping performance.
Inline redundant local variables.
Remove deprecated API usage from tests.
Remove redundant character escape in RegEx.
Remove redundant contains operation when performing .remove
Remove String.intern() used in compile-time constants.

I have run all the tests successfully after the changes.


chrjohn:
Thanks for the PR!

PR #186, opened 2018-04-20
Added additional construction to support supplying a custom sessionDi…

pbyrne413:
…ctionary.

Added test case to cover using the new feature. 

Test checks that when using custom SessionDictionary new custom fields are correctly picked up and identified as header fields.


chrjohn:
Thank you for the PR. There are some other PRs to take care of but will merge soon.
Cheers

PR #205, opened 2018-07-28
Fix typo in CompositeTypeFactory

bezpredel:


chrjohn:
Hi @bezpredel , thanks for the PR. Could you please elaborate on the change you made?
Thanks,
Chris.

bezpredel:
Hi John,

If fixes a typo that was mitigated vy the fact that current code that
populates these two arrays gets called with the same value for both.

If you disagree or are not sure that nothi g relies on it, just discard it)

chrjohn:
Hi @bezpredel ,
looks good to me.
Thanks again.

PR #206, opened 2018-07-30
QFJ-950 reject garbled messages

chrjohn:
Provide configuration to have garbled messages rejected instead of ignored.

### DIFF
--- a/quickfixj-core/src/main/doc/usermanual/usage/configuration.html
+++ b/quickfixj-core/src/main/doc/usermanual/usage/configuration.html
@@ -422,7 +422,7 @@
   </TR>
   <TR ALIGN="left" VALIGN="middle">
     <TD> <I>RejectGarbledMessage</I> </TD>
-    <TD> If RejectGarbledMessage is set to Y, garbled messages will be rejected instead of ignored.<br>
+    <TD> If RejectGarbledMessage is set to Y, garbled messages will be rejected (with a generic error message in 58/Text field) instead of ignored.<br>
         This is only working for messages that pass the FIX decoder and reach the engine.<br>
         Messages that cannot be considered a real FIX message (i.e. not starting with 8=FIX or not ending with 10=xxx) will be ignored in any case.</TD>
     <TD> Y<br>N</TD>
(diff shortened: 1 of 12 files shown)


philipwhiuk:
I feel like the documentation needs to be clearer in terms of what a Garbled Message is vs an Invalid Message, especially as the code seems to use `MessageUtils.newInvalidMessageException()` to handle a garbled message yet we have two properties, RejectInvalidMessage and RejectGarbledMessage. Perhaps we should have a table with some examples and the output (procesed, dropped, rejected?) - e.g [but probably not accurate]:

| Issue  | RejectInvalidMessage | RejectGarbledMessage | Output |
| ------------- | ------------- |------------- |------------- |
| Missing mandatory field  | N | N | Ignored |
|   | N  | Y | Ignored |
|   | Y | N | Ignored |
|   | Y  | Y | Rejected |
| Field with no data  | N | N | Ignored |
|   | N  | Y | Ignored |
|   | Y | N | Ignored |
|   | Y  | Y | Rejected |
| Incorrect repeating group count  | N | N | Processed |
|   | N  | Y | Ignored |
|   | Y | N | Ignored |
|   | Y  | Y | Rejected |
| Repeating group count not a number  | N | N | Ignored |
|   | N  | Y | Ignored |
|   | Y | N | Ignored |
|   | Y  | Y | Rejected |

chrjohn:
Hi @philipwhiuk ,
thanks for your comment. Initially I planned to use RejectInvalidMessage but that was already taken. :-( So I went for garbled message because all cases that were signalled with a InvalidMessage Exception were ignored and the FIX spec states that garbled messages should be ignored.
However, I think an update to the documentation is a good idea! I will add such a table.
Thanks,
Chris.

chrjohn:
Hi @philipwhiuk ,
I have added some explanations to the documentation. Maybe you could review it? I did not go with the table structure but decided to give a little background information on the two settings (data dictionary validation failure vs. FIX protocol failure).
Let me know what you think.

Thanks,
Chris.

philipwhiuk:
Looks good to me :+1:

chrjohn:
Great, thanks.

PR #207, opened 2018-07-31
Do not throw IncorrectDataFormat exception if value is empty and ValidateFieldsHaveValues=N

jonfreedman:


chrjohn:
Thanks @jonfreedman ,
I will add #207 to the next release due to be released in a few days.
I will have a look at #208 later.
Hope that is OK with you.

Cheers,
Chris.

PR #210, opened 2018-08-26
QFJ-636 - Custom fields in component blocks are not added to parent

the-thing:
Hi @chrjohn,

Custom fields and groups are added to parent now (ie. setting, getting and copying is supported). Please keep in mind that when support of custom fields and groups is required new methods must be used. Previous methods:

- quickfix.FieldMap#setComponent(MessageComponent component)
- quickfix.FieldMap#getComponent(MessageComponent component)
- quickfix.MessageComponent#copyFrom(FieldMap fields)
- quickfix.MessageComponent#copyTo(FieldMap fields)

are backwards compatible - we don't want to break existing behaviour on production + there is an extra array creation required to copy all or selected fields/groups.

There are new methods that allow copying all fields and groups:

- quickfix.FieldMap#getComponent(MessageComponent component, boolean includeAllFields, boolean includeAllGroups)
- quickfix.FieldMap#getComponent(MessageComponent component, int[] includeFields, int[] includeGroups)
- quickfix.FieldMap#setComponent(MessageComponent component, boolean includeAllFields, boolean includeAllGroups)
- quickfix.FieldMap#setComponent(MessageComponent component, int[] includeFields, int[] includeGroups)
- quickfix.MessageComponent#copyFrom(FieldMap fields, boolean includeAllFields, boolean includeAllGroups)
- quickfix.MessageComponent#copyFrom(FieldMap fields, boolean includeAllFields, boolean includeAllGroups)
- quickfix.MessageComponent#copyTo(FieldMap fields, boolean includeAllFields, boolean includeAllGroups)
- quickfix.MessageComponent#copyTo(FieldMap fields, boolean includeAllFields, boolean includeAllGroups)

Also new methods are generated for each component, message and fix dictionary:

- #set(XXX component, boolean includeAllFields, boolean includeAllGroups)
- #set(XXX component, int[] includeFields, int[] includeGroups)
- #get(XXX component, boolean includeAllFields, boolean includeAllGroups)
- #get(XXX component, int[] includeFields, int[] includeGroups)
- #getXXX(boolean includeAllFields, boolean includeAllGroups)
- #getXXX(int[] includeFields, int[] includeGroups)

It seems that quickfixj-all.jar file is 1.2MB larger (size 22.6 MB, compared to quickfixj-all-2.1.0.jar which is 21.4 MB)

### DIFF
--- a/quickfixj-codegenerator/src/main/resources/org/quickfixj/codegenerator/MessageSubclass.xsl
+++ b/quickfixj-codegenerator/src/main/resources/org/quickfixj/codegenerator/MessageSubclass.xsl
@@ -271,18 +271,44 @@ import quickfix.Group;</xsl:if>
 
 <xsl:template name="component-accessor-template">
 	<xsl:variable name="type" select="concat($messagePackage,'.component.',@name)"/>
-	public void set(<xsl:value-of select="$type"/> component) {
-		setComponent(component);
-	}
+`    public void set(<xsl:value-of select="$type"/> component) {
+        setComponent(component);
+    }
 
-	public <xsl:value-of select="$type"/> get(<xsl:value-of select="$type"/> component) throws FieldNotFound {
-		getComponent(component);
-		return component;
-	}
+    public void set(<xsl:value-of select="$type"/> component, boolean includeAllFields, boolean includeAllGroups) {
+        setComponent(component, includeAllFields, includeAllGroups);
+    }
 
-	public <xsl:value-of select="$type"/> get<xsl:value-of select="@name"/>() throws FieldNotFound {
-		return get(new <xsl:value-of select="$type"/>());
-	}
+    public void set(<xsl:value-of select="$type"/> component, int[] includeFields, int[] includeGroups) {
+        setComponent(component, includeFields, includeGroups);
+    }
+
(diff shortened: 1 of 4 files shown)


chrjohn:
Hi @the-thing ,

thanks for the PR. Will have a look at it later.
W.r.t. the message size: you mean that after your changes the JAR file size increases due to the newly generated methods in every message class?

Thanks,
Chris.

the-thing:
Hi @chrjohn,

The FIX message size will not change. I made a note that about increased jar file sizes. I'm still wondering what is your opinion on creating new methods against only changing the existing ones.

Thanks,
Marcin

chrjohn:
Hi @the-thing ,
I was referring to the generated message classes, not the FIX messages on the wire. :)
Hmm, if I understood correctly the current behaviour (i.e. not copying everything from the message) is a bug. So there probably is nothing wrong with changing the existing methods to copy everything. However, the extra array creation might be unexpected.
Chris.

the-thing:
Actually I changed the default method to include all fields and now components have access to all message fields. This is definitely something not desired. I will close to pull request for the time being as I need to think what is a better solution to fix this.

PR #212, opened 2018-10-29
Functional adapters for Application and ApplicationExtended interfaces

jasition:
Provides adapter implementation of quickfix.Application and quickfix.ApplicationExtended that

1. allows registering multiple listeners for each operation (e.g. onLogon, onCreate) separately using Java 8's functional interfaces like Consumer, BiConsumer and Predicate. FromAdmin, FromApp and ToApp requires custom interfaces due to checked Exceptions. Also allows removing registered listeners.

2. invokes each registered listeners in a FIFO manner

3. for Predicate, it fail-fast when encounters the first Predicate returning false

4. for checked Exception listeners, it fail-fast when encounters the first Exception thrown

5. is completely optional to Quickfix users

6. is backward compatible

7. is thread-safe with the use of CopyOnWriteArrayList

8. does not add extra lib dependency

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/ApplicationFunctionalAdapter.java
+++ b/quickfixj-core/src/main/java/quickfix/ApplicationFunctionalAdapter.java
@@ -1,5 +1,7 @@
 package quickfix;
 
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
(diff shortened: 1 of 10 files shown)


charlesbr1:
Hi, it's clean code. But what about the MessageCraker and type safe style ?

jasition:
> Hi, it's clean code. But what about the MessageCraker and type safe style ?

This pull request is mainly concerned about the use of functional interfaces. I could have a look at being type safe as the next step.

Although it was highly recommended to choose the most type safe way to handle quickfix message, it should not be mandatory. And if the users wish to, they could mix in MessageCracker in the FromAppListener implementation by something like this:

new FromAppMessageCracker(new FromAppListenerImpl());

MessageCracker will scan the first parameter and register the invokers to be called when the type is matched.

jasition:
If we were to attempt to use MessageCracker as type-safe delegation for fromApp, toApp, fromAdmin and toAdmin, the first thing we'll find is that we'll need to pass in java.lang.Object to have its methods scanned on "onMessage" and the "@Handler" annotation. And lambda expressions no longer work because the compiler expects an interface as lambda's target.

Then how about generic type?, something like

public class FromAppListener<T extends Message> {
   public void onMessage(T message, SessionID sessionID)
}

but due to Erasure of java generics, there is no way to resolve what T actually is at runtime.

The only workaround I could think of is to provide the Class when adding the listener. It would look like this:

adapter.addFromAppListener(NewOrderSingler.class, (NewOrderSingle m, SessionID s) -> {});

and the adapter will keep a Map<Class, List<FromAppListener>> for type safe delegation. In this usage, there is no need to use MessageCracker. No need to conform to "onMessage" or annotate the method with "@Handler".

To conclude this, I think I could at least modify the custom interfaces to have <T extends Message> in preparation for the type-safe delegation. And the actual type-safe implementation should not be too far ahead, if you agree adding the Class in the argument of listener registration is acceptable.

jasition:
I have added support for type-safe listeners for toApp, fromApp, toAdmin and fromAdmin, as mentioned in my previous comment.

chrjohn:
Hi @jasition , thanks for the PR.

Since your changes do not break anything they should be good to go. However, I'll wait some time if anyone does a review.

Cheers,
Chris.

jasition:
Thank you Chris. If this is going to be accepted, I would update the receiving_messages.html to add the usage example.

chrjohn:
Yes, that would be great. If it is not too much effort you could also add a short example to the top-level README.md. But I could also do that later on.

jasition:
I have added the example to README.md and receiving_messages.html. However, feel free to modify, organise and integrate the new section with other sections in the document however you like.

jasition:
Thanks for your comment. Having reviewed the cases that we really would create a new queue for the message type are only when
1. Registering to a new Message Type the first time, and
2. Receiving a new Message Type the first time

With the double-checked locking technique, I could make it lock-free if the queue already exists. A new pull request is coming with all the grammar corrections. The mistakes in README.md and receiving_messages.html were the same since the content was the same. So I will correct both.

quickfixj-core/src/main/java/quickfix/ApplicationFunctionalAdapter.java
```diff
@@ -0,0 +1,301 @@
...
+
+    private <T> List<T> getList(Map<Class, List<T>> multimap, Class clazz) {
+        List<T> list = multimap.get(clazz);
+
+        if (list != null) {
+            // Return without synchronization if the list was found
+            return list;
+        }
+
+        synchronized (multimap) {
```
chrjohn: Hi @jasition , @colinduplantis 

please excuse me if I misunderstood the intention, but why don't we just use a `ConcurrentMap` and utilize `putIfAbsent()`? That way we do not need explicit synchronization and perform an atomic update.

Thanks,
Chris.

jasition: Hi @chrjohn, 

I completely overlooked the `computeIfAbsent()` method which will eliminate the while loop to put the list to ConcurrentMap in previous commit. 

I personally would prefer `ConcurrentMap` because it is atomic. With the use of `computeIfAbsent()`, I can make an one-liner like this in the `getList` method:

`return multimap.computeIfAbsent(k -> new CopyOnWriteArrayList())`

and that should also address @colinduplantis 's concern on potential infinite loop.

jasition:
Hi @charlesbr1, I haven't encountered the infinite loop in my experience. But I could see how `get()` and `put()` could be invoked at the same time due to the double-checked locking in the old code. The `get()` wasn't in the `synchronized` block! The current implementation using `computeIfAbsent` is much simpler and safe.

gbirchmeier:
Is there an example of an application that uses this?  I've searched pretty hard and can't find one.

Specifically, how am I supposed to instantiate ApplicationFunctionalAdapter?  What is the relationship between AFA and Quickfix.Application?

It looks neat and I think it might be useful in my app, but... the readme is just not quite enough to get me there.  I feel like I've got all the pieces in front of me, I just don't know how to put them together.  An example would help a lot.

gbirchmeier:
I did not see that test, thanks.  Not sure how I missed it.

I'd still like to see a concrete instantiation of FromAppListener, and an example of how `accept()` is supposed to be used (the javadoc for it is blank).

I'll have to hack up a demo app and experiment.

Does AFA still throw UnsupportedMessageType if no listener has a function assigned to handle a message type?  I don't see a test that covers that.

PR #215, opened 2018-11-14
Format messages output

mohamedchebbiQH:
std::string FormatMessage(const std::string& value)
	{
		std::string ret(value);
		std::replace(ret.begin(), ret.end(), '\001', '|');
		return ret;
	}
        void onIncoming( const std::string& value )
        { m_messages << UtcTimeStampConvertor::convert(UtcTimeStamp(), 9) << " : Incoming: " <<FormatMessage(value) << std::endl; }
  
       void onOutgoing( const std::string& value )
       { m_messages << UtcTimeStampConvertor::convert(UtcTimeStamp(), 9) << " : Outgoing: " <<FormatMessage(value) << std::endl; }


chrjohn:
Hi @mohamedchebbiQH , I think you are on the wrong project. Did you maybe mean to add this to QuickFIX (C++)??
https://github.com/quickfix/quickfix

Chris.

PR #217, opened 2018-11-22
Bugfix for QFJ-963

alexwibowo:
This is a PR that fixes bug described in https://www.quickfixj.org/jira/browse/QFJ-963


alexwibowo:
This fix depends on the fix i've done on QFJ-964.

quickfixj-core/src/test/java/quickfix/mina/SessionConnectorTest.java
```diff
@@ -144,7 +126,7 @@ public void testOneSessionLoggedOnOneSessionNotLoggedOne() throws Exception {
                     SessionFactory.ACCEPTOR_CONNECTION_TYPE);
             try (Session session2 = connector.createSession(sessionID2)) {
                 assertNotNull(session2);
-                sessions.put(session2.getSessionID(), session2);
+                connector.addDynamicSession(session2);
```
chrjohn: Is there a reason why this was changed?

alexwibowo: I think the way this test was set up in the first place is a bit weird. Looking at the production code, connection#setSession() is only called during the construction of Acceptor / Initiator, from a local HashMap. In other word, no other thread can modify this HashMap post the setSession() call.

Secondly (and probably a lesser reason to the change on the test), I've changed the implementation of setSession() on the connector so that when it is called, it will clear the internal sessions, and copy from the map that being passed in. But as you can see in this test, it calls setSession(), followed by adding more session to the original map (i.e. not reflective on how it is being used in production code). This caused the test to fail.

Thirdly.. it is kind of adding dynamic session. If you read how the test flows...it creates a single session, register it with the connector... followed by adding more session. Well.. i might be wrong.. but that looks 'dynamic' to me :)

PR #218, opened 2018-11-22
QFJ-965 Fixed regression when using non-default charset

amichair:
The optimizations in commit 6229d0722d8c6f6e7b2b251772a92caf2f4bdc25 cause a regression where using fields with a non-default charset such as UTF-8 causes message corruption (even if the charset is properly set via CharsetSupport.setCharset()).

Further, the test that caught this regression (MessageTest.testMessageWithEncodedField via its helper method) was modified in the same commit to work around the test failure, so that the test passes but the bug still occurs in real applications.


charlesbr1:
sorry for this bug & thx for your correction that still keep the improvement

PR #221, opened 2019-03-14
Optimize DataDictionary

ZaitsevY:
DataDictionary keeps some data in Map<IntStringPair, .V> structures. A new instance of IntStringPair is created every time to search the value in the map. When parsing market data (with a lot of repeating groups) it causes a lot of IntStringPair allocations (see the attached profiler screenshot, taken after ~1000 MarketDataSnapshotFullRefresh messages received).
before
Changing implementation to Map<Integer, Map<String, V>> nested maps allows to get rid of IntStringPair at all. The results are on another screenshot.
after

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/DataDictionary.java
+++ b/quickfixj-core/src/main/java/quickfix/DataDictionary.java
@@ -84,8 +84,8 @@ public class DataDictionary {
     private final Map<Integer, Set<String>> fieldValues = new HashMap<>();
     private final Map<Integer, String> fieldNames = new HashMap<>();
     private final Map<String, Integer> names = new HashMap<>();
-    private final Map<IntStringPair, String> valueNames = new HashMap<>();
-    private final Map<IntStringPair, GroupInfo> groups = new HashMap<>();
+    private final IntegerStringMap<String> valueNames = new IntegerStringMap<>();
+    private final StringIntegerMap<GroupInfo> groups = new StringIntegerMap<>();
     private final Map<String, Node> components = new HashMap<>();
 
     private DataDictionary() {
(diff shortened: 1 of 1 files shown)


chrjohn:
Hi,
thanks for the PR and sorry for the long delay. I will check it during the next days but it seems it has already gotten a review. Thanks @philipwhiuk 

Cheers,
Chris.

philipwhiuk:
I didn't do a detailed review of this - but I've seen DataDictionary be a huge hog when debugging stuff, so I was glad to see work on it. I can't think that we particularly care too much about the internal data structure so given it passed all the existing tests I was happy enough.

Hazines:
Hi, 
I'm just a beginner and I need to implement a solution using fix protocol 5.0. I will use quickfix/J but I didn't understand how it works exactly. I read the guidelines but they weren't very clear for me.
Can you please tell me if Data Dictionary is the XML file (fix50.xml) ?  And do I need to create a java class for each type of message and for each field ?
I would appreciate your help.
Thank you.

chrjohn:
Hi @Hazines ,
this is not the correct place for such questions.
A basic introduction is even on the main page of this project https://github.com/quickfix-j/quickfixj or here: https://www.quickfixj.org/usermanual/2.1.0/
For further questions (after checking the documentation), please use the mailing list: https://sourceforge.net/projects/quickfixj/lists/quickfixj-users
Thanks,
Chris.

Hazines:
Hi,
Okay thank you very much.
I wanted to ask a question on the mailing list but i didn't find how!!

Cordially.



Le ven. 12 juil. 2019 à 10:38, Christoph John <notifications@github.com> a
écrit :

chrjohn:
Hi @Hazines , you can subscribe to the mailing list via the link I posted in my former comment:
https://sourceforge.net/projects/quickfixj/lists/quickfixj-users
When you have successfully subscribed to the mailing list you can simply send mails to quickfixj-users@lists.sourceforge.net 
Cheers,
Chris.

PR #222, opened 2019-05-01
Avoid NullPointerException when unknown tag goes before the first field in repeating group (QFJ-980)

MAXbrainRUS:


chrjohn:
Hi,
thanks for the PR and sorry for the long delay. And thanks again for the review, @philipwhiuk

PR #224, opened 2019-07-19
QFJ-981 Validation settings do not get applied on default AppDataDictionary

chrjohn:
- adapted DefaultSessionFactory and unit test


chrjohn:
Some of the acceptance tests are failing now, probably due to caching of data dictionaries. Will investigate.

chrjohn:
This is covered by #245

PR #225, opened 2019-07-24
QFJ-976: GarbledMessage without MsgType

WojciechZankowski:
* Added the same test for all garbled messages that we have in AcceptorIoHandler
* I used Optional, but I couldn't use ifPresent block because this block of code is throwing checked exception
* In AcceptorIoHandler theoretically we could have the same issue, but check for message type is 'guarded' by session check 

```java
Session qfSession = (Session) protocolSession.getAttribute(SessionConnector.QF_SESSION);
        if (qfSession == null) {
            if (message.getHeader().getString(MsgType.FIELD).equals(MsgType.LOGON)) {
```

I don't know if it is valid use case that we could receive GarbledMessage before Logon message. If it is valid then I can adjust this class as well.

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/mina/acceptor/AcceptorIoHandler.java
+++ b/quickfixj-core/src/main/java/quickfix/mina/acceptor/AcceptorIoHandler.java
@@ -19,12 +19,7 @@
 
 package quickfix.mina.acceptor;
 
-import java.net.InetAddress;
-import java.net.InetSocketAddress;
-import java.net.SocketAddress;
-
 import org.apache.mina.core.session.IoSession;
-
 import quickfix.Log;
 import quickfix.Message;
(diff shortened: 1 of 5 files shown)


chrjohn:
Hi @WojciechZankowski ,
thanks for the PR. Looks good to me. 
It would be great if you could also change the AcceptorIoHandler to use the Optional. That way we are on the safe side. Normally that case should not happen (although it is of course possible to provoke it) but if it does then we should handle it gracefully.

Thanks and cheers,
Chris.

WojciechZankowski:
Sure, I will do the same for AcceptIoHandler after my work.

chrjohn:
No need to hurry. :) Thanks in advance

WojciechZankowski:
I commited fix for AcceptorIoHandler as well, cheers.

PR #227, opened 2019-07-27
QFJ-973: Provide configuration to turn off checksum validation

WojciechZankowski:
Added:

* Setting and checksum validation flag
* Acceptance Test
* New entry in documentation


chrjohn:
Hi @WojciechZankowski ,
thanks for the PR! :) I will add some comments later.
Cheers,
Chris

quickfixj-core/src/main/doc/usermanual/usage/configuration.html
```diff
@@ -456,6 +456,14 @@ <H3>QuickFIX Settings</H3>
     <TD> Y<br>N</TD>
     <TD> Y </TD>
   </TR>
+  <TR ALIGN="left" VALIGN="middle">
+      <TD> <I>ProcessMessageWithInvalidChecksum</I> </TD>
```
chrjohn: Just a gut feeling, but I think this setting should rather be named `ValidateChecksum`. Unfortunately the logic and default value needs to be the other way round then.
The flags and methods inside the Message class are also named `validateChecksum` already.
Maybe wait some days if one of the other regulars has another opinion on this. @philipwhiuk maybe?

quickfixj-core/src/main/java/quickfix/DefaultSessionFactory.java
```diff
@@ -84,6 +84,14 @@ public Session create(SessionID sessionID, SessionSettings settings) throws Conf
             final boolean rejectGarbledMessage = getSetting(settings, sessionID,
                     Session.SETTING_REJECT_GARBLED_MESSAGE, false);
 
+            final boolean processMessageWithInvalidChecksum = getSetting(settings, sessionID,
+                    Session.SETTING_PROCESS_MESSAGE_WITH_INVALID_CHECKSUM, false);
+
+            if (rejectGarbledMessage && processMessageWithInvalidChecksum) {
+                throw new ConfigError("Not possible to reject garbled message and process " +
+                        "messages with invalid cheksum at the same time.");
```
chrjohn:
```suggestion
                        "messages with invalid checksum at the same time.");
```

quickfixj-core/src/main/doc/usermanual/usage/configuration.html
WojciechZankowski: Validate Checksum works for me as well. I will wait for opinion from others.

quickfixj-core/src/test/java/quickfix/DefaultSessionFactoryTest.java
```diff
@@ -254,4 +259,13 @@ public void testTimestampPrecision() throws Exception {
         createSessionAndAssertConfigError("no exception", ".*No enum constant quickfix.UtcTimestampPrecision.PICOS.*");
     }
 
+    // QFJ-973
+    @Test
+    public void testRejectGarbledMessageAndProcessInvalidChecksumError() {
+        settings.setString(Session.SETTING_REJECT_GARBLED_MESSAGE, "Y");
+        settings.setString(Session.SETTING_PROCESS_MESSAGE_WITH_INVALID_CHECKSUM, "Y");
+        createSessionAndAssertConfigError("no exception", ".*Not possible to reject " +
+                "garbled message and process messages with invalid cheksum at the same time.*");
```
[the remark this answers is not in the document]
chrjohn: Nevermind, just corrected it.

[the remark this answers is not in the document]
WojciechZankowski: Thanks! Too much code coverage 😅

quickfixj-core/src/main/doc/usermanual/usage/configuration.html
chrjohn: I think you can go ahead.
Thanks

WojciechZankowski:
I changed the name and I had to reverse the behavior in couple places. Rebased to newest master. Please take a look if I didn't miss anything, thanks!

quickfixj-core/src/main/java/quickfix/DefaultSessionFactory.java
```diff
@@ -84,10 +84,10 @@ public Session create(SessionID sessionID, SessionSettings settings) throws Conf
             final boolean rejectGarbledMessage = getSetting(settings, sessionID,
                     Session.SETTING_REJECT_GARBLED_MESSAGE, false);
 
-            final boolean processMessageWithInvalidChecksum = getSetting(settings, sessionID,
-                    Session.SETTING_PROCESS_MESSAGE_WITH_INVALID_CHECKSUM, false);
+            final boolean validateChecksum = getSetting(settings, sessionID,
+                    Session.SETTING_VALIDATE_CHECKSUM, false);
```
chrjohn: I think this setting should have "true" as default.

chrjohn: This should hopefully fix the failing tests.

philipwhiuk:
The name is fine. I can't say I'm likely to use this - we have test tools that auto-correct the checksum of manually copied FIX messages instead and I would not want it on in any real environment. 

I'm not sure I'm a huge fan of the constructor growth - it's getting to the point we have to consider a MessageBuilder. But that's a change we can put in QFJ 3.0

chrjohn:
Hi @philipwhiuk ,
we sometimes have clients that face validation problems (e.g. due to special characters) and are happy when there is a setting that can be used until the real problem is solved or cause is found.
But in general I agree with you that it is better to always send messages with correct checksum.

Thanks,
Chris.

PR #228, opened 2019-07-29
QFJ-811: Wrong field order in AllocGrp

WojciechZankowski:
Not big deal, but I fixed it. Somehow field order in AllocGrp was wrong in FIX50.xml file. In files for SP1 & SP2 it seems to be correct. 

Verified with: 

* Compared with FIX50.xml from http://www.quickfixengine.org/
* Compared with Fiximate https://fiximate.fixtrading.org/legacy/index.html


chrjohn:
Thanks a lot!

PR #229, opened 2019-07-29
QFJ-722: Optimize unnecessary trailer and header allocations

WojciechZankowski:
I am not sure about this. In fact it removes unnecessary allocations, but also touches fragile part of code generation? 

Up to you! 

BTW. It is shame that we have so many constructors in Message created just for unit tests. Ones that take String should be at most static factory method or completly removed.


philipwhiuk:
I don't see the optimisation here. As soon as the object is created the header will be. In what circumstance does it improve anything. All you're doing really is adding another function call.

The only real benefit that I can see is it allows an overriding class to replace the Header & Trailer with a custom implementation. That's pretty thin.

quickfixj-core/src/test/java/quickfix/MessageTest.java
```diff
@@ -21,69 +21,7 @@
...
-import quickfix.field.Signature;
-import quickfix.field.SignatureLength;
-import quickfix.field.Symbol;
-import quickfix.field.TargetCompID;
-import quickfix.field.TargetSubID;
-import quickfix.field.TotNoOrders;
-import quickfix.field.TransactTime;
-import quickfix.field.UnderlyingCurrency;
-import quickfix.field.UnderlyingSymbol;
+import quickfix.field.*;
```
philipwhiuk: I think we prefer single-class imports.

WojciechZankowski: Yep, sorry, changed!

PR #232, opened 2019-08-12
Fixed Issue QFJ-983, "SLF4JLog is suffixing category with session ID …

simon-mebrahtu:
…instead of prefixing it."

JIRA:
+ https://www.quickfixj.org/jira/browse/QFJ-983

Mailing list discussion about issue:
+ https://sourceforge.net/p/quickfixj/mailman/message/36714538/


chrjohn:
Great, thanks for the PR!

PR #233, opened 2019-08-13
Provide correct end sequence number sent in request in logs

philipwhiuk:
It's misleading to suggest we're sending a closed range request when instead we're sending an open ended request.


philipwhiuk:
Hmm, I need to look at `ResendRange` in some detail for that. I don't entirely understand why the range is set based on the received `msgSeqNum` not `endSeqNo`.

philipwhiuk:
@chrjohn hopefully this is good now :)

chrjohn:
Hi @philipwhiuk , yes, thank you. :) 
:+1: for the additional javadoc.

PR #234, opened 2019-08-15
QFJ-983 Revert changes made to try to optimise SLF4JLog

philipwhiuk:
I think I probably need to revert some tests too. It's a pure revert - we can probably keep some bits.
We should probably also workout how to ensure it's not broken in the future somehow.

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/SLF4JLog.java
+++ b/quickfixj-core/src/main/java/quickfix/SLF4JLog.java
@@ -47,24 +47,26 @@ public class SLF4JLog extends AbstractLog {
 
     private final Logger outgoingMsgLog;
 
+    private final String logPrefix;
+
     private final String callerFQCN;
 
     public SLF4JLog(SessionID sessionID, String eventCategory, String errorEventCategory,
-                    String incomingMsgCategory, String outgoingMsgCategory, boolean prependSessionID,
-                    boolean logHeartbeats, String inCallerFQCN) {
+            String incomingMsgCategory, String outgoingMsgCategory, boolean prependSessionID,
+            boolean logHeartbeats, String inCallerFQCN) {
         setLogHeartbeats(logHeartbeats);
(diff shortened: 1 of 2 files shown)


WojciechZankowski:
If you meant to fix defect QFJ-983 then it is already fixed here: https://github.com/quickfix-j/quickfixj/pull/232https://github.com/quickfix-j/quickfixj/pull/232

Cheers,

Wojtek

philipwhiuk:
No it isn't. The fix doesn't work. 

Previously you'd have a logging setup that looks like:

```
log4j.logger.quickfixj.event=INFO
log4j.appender.stdout.layout.ConversionPattern=%c{1} %p %d{yyyyMMdd-HH:mm:ss.SSS} - %m%m
```

And with the default configuration you'd get

`event INFO 20190102-03:04:05.678 - 8=FIX.4.4....`

With the original suffix this gives you nothing because the fully-qualified logger name is:

`quickfixj.eventFIX.4.4:SENDER->TARGET: `

With it as a prefix it gives you nothing because you have to declare something like this for every session.

`log4j.logger.FIX.4.4:SENDER->TARGET.quickfixj.event=INFO`

because the fully-qualified logger name is:

`FIX.4.4:SENDER->TARGET.quickfixj.event` and so the heirarchical structure fails.

I made a change locally to make the fully-qualified logger name:

`quickfixj.event.FIX.4.4:SENDER->TARGET: `

But even this isn't usable as it gives:

`4:SENDER->TARGET: INFO 20190102-03:04:05.678 - 8=FIX.4.4....`

The change, with or without the fix that was made breaks logging configurations and makes it harder to actually specify only logging specific stuff.

chrjohn:
Hi @philipwhiuk ,
is this good to merge or do you still want to add something?
Thanks,
Chris.

philipwhiuk:
I think I'm happy enough now.

PR #235, opened 2019-08-16
Field ordering functionality.

philipwhiuk:
`TreeMap` is the default ordering for fields in QFJ. This provides numerical tag based ordering.

Occasionally it is useful to be able to override it to enforce a specific ordering of fields based on a provided list. 

It's also useful to be able to get the ordered lists of fields from the Dictionary for a Message/Group.


philipwhiuk:
I'll have a look why this is breaking a test :)

PR #236, opened 2019-08-16
Add extra options to session class

philipwhiuk:
We have long had some internal setting that I feel like upstream will benefit from (also it makes my life easier).

**AllowPosDup**

`PossDupFlag` is set when resending messages. Currently it's automatically removed by QFJ. 
This PR makes two changes:

* It exposes a method 'sendNoChange' which never removes these settings. This is useful when resending messages from a cache or from an application store behind QFJ.
* It exposes a session setting 'AllowPosDup' that means we never remove the flag. This has been useful on occasion - primarily when a QFJ application is acting as purely a pass-through/monitoring hop.

**Fix41ResendRequestAsFix42**

* Some sessions claim to be FIX.4.1 but actually send FIX.4.2 format ResendRequests. This is useful for that. Quite common in less formalised trading networks where quite frankly the begin string is often just misleading.

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/Session.java
+++ b/quickfixj-core/src/main/java/quickfix/Session.java
@@ -362,6 +362,16 @@ public class Session implements Closeable {
 
     public static final String SETTING_VALIDATE_CHECKSUM = "ValidateChecksum";
 
+    /**
+     * Option so that the session doesn't remove PossDupFlag (43) information when sending.
+     */
+    public static final String SETTING_ALLOW_POS_DUP_MESSAGES = "AllowPosDup";
+    /**
+     * Setting telling us to treat FIX.4.1 resend requests as FIX.4.2 ones. I.e. requesting to '0' actually
+	 * requests to infinity.
+	 */
+    public static final String SETTING_FIX_41_RESEND_AS_FIX42 = "Fix41ResendRequestAsFix42";
+
     private static final ConcurrentMap<SessionID, Session> sessions = new ConcurrentHashMap<>();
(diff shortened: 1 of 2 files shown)


philipwhiuk:
I'll have to work out how to handle it if we rename them. We can probably map the old values in our code before passing them to QFJ. I'm open to renaming it and making it more general.

In terms of sending that does make sense. I'll see if we have any active usage of it these days (it's been internal for way way too long) and see whether we can try to see whether it was intentionally only done on one side.

I'll try to write some more tests for both pieces before we merge this in :)

esanchezros:
Hi @chrjohn,

What needs doing on this PR? I'm happy to help moving it forward

Thanks

philipwhiuk:
Hi @esanchezros. I'll update this tomorrow with some tests and the changes from Chris and then hopefully it can be merged :)

esanchezros:
Hi @philipwhiuk,
I thought this PR was abandoned and created a new one just for allowing PossDupFlag in send https://github.com/quickfix-j/quickfixj/pull/395
If you and @chrjohn are okay with it perhaps it can be reviewed/merged and you could create a separate one for the other feature?
Thanks 
Ed

chrjohn:
Hi @philipwhiuk @esanchezros 

personally I would prefer to have each feature in a single PR, so if it is OK with you (@philipwhiuk) then I would be happy if we could proceed with #395 for the PossDup thing and leave this PR for the FIX4.1/4.2 Resend option.

What do you think?
Thanks,
Chris.

PR #242, opened 2019-11-20
Persistence is done before sending and will fail if message cannot be stored or sequence number cannot be incremented.

nolag:
This applies a fix from quickfix's main repository https://github.com/quickfix/quickfix/commit/5fb3d4ffc44775295d5525fb30335d57e869f27e.

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/Session.java
+++ b/quickfixj-core/src/main/java/quickfix/Session.java
@@ -2538,6 +2538,16 @@ public class Session implements Closeable {
         state.setLogonSent(true);
     }
 
+    private void persist(Header header, String messageString, int num) throws IOException, FieldNotFound {
+      if (num == 0) {
+          if (persistMessages) {
+              final int msgSeqNum = header.getInt(MsgSeqNum.FIELD);
+              state.set(msgSeqNum, messageString);
+          }
+          state.incrNextSenderMsgSeqNum();
+      }
+    }
+
     /**
      * Send the message
      *
@@ -2591,6 +2601,7 @@ public class Session implements Closeable {
                 }
 
                 messageString = message.toString();
+                persist(message.getHeader(), messageString, num);
                 if (MsgType.LOGON.equals(msgType) || MsgType.LOGOUT.equals(msgType)
                         || MsgType.RESEND_REQUEST.equals(msgType)
                         || MsgType.SEQUENCE_RESET.equals(msgType) || isLoggedOn()) {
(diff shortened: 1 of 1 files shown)


chrjohn:
Thanks for the PR! I was always asking myself why the order of send and persist is not switched. :)

nolag:
> Thanks for the PR! I was always asking myself why the order of send and persist is not switched. :)

My pleasure, I just want to send a friendly ping, we're hoping to use official builds at work but this is a blocker for us.

chrjohn:
I'm pretty swamped with work at the moment so cannot really say when the next release will be done. :-/
Hopefully I'll have time in January.

quickfixj-core/src/main/java/quickfix/Session.java
```diff
@@ -2538,6 +2538,16 @@ private void generateLogon(Message otherLogon, int expectedTargetNum) throws Fie
         state.setLogonSent(true);
     }
 
+    private void persist(Message message, int num) throws IOException, FieldNotFound {
+      if (num == 0) {
+          final int msgSeqNum = message.getHeader().getInt(MsgSeqNum.FIELD);
```
chrjohn: Could you please move this line inside of the `if (persistMessages)` block since it is not used elsewhere.

quickfixj-core/src/main/java/quickfix/Session.java
```diff
@@ -2590,6 +2600,7 @@ private boolean sendRaw(Message message, int num) {
                     }
                 }
 
+                persist(message, num);
                 messageString = message.toString();
```
chrjohn: Could you please swap these lines and pass the `messageString` into `persist()`. That avoids two calls to `toString()` which should be avoided since it recalculates some fields of the message.

quickfixj-core/src/main/java/quickfix/Session.java
```diff
@@ -2605,19 +2616,12 @@ private boolean sendRaw(Message message, int num) {
                     logApplicationException("toApp()", t);
                 }
                 messageString = message.toString();
+                persist(message, num);
```
chrjohn: Could you please pass the `messageString` into `persist()`. That avoids two calls to `toString()` which should be avoided since it recalculates some fields of the message.

quickfixj-core/src/main/java/quickfix/Session.java
nolag: Sure, didn't realize it was that inefficient, sorry :).

nolag:
Sorry, Intellij got in the way of one of those commits.  I'll fix that.

nolag:
Ready again, sorry about the multiple force commits, I wanted to keep your history clean.

prerak-proof:
FYI - this causes the unit test `SessionDisconnectTest.reconnectReceivingLogonResponseBeforeLogonRequestPersisted` to be broken/obsoleted as the situation it describes can no longer occur:  
> The purpose of this test is to simulate the situation where a logon response is received before a logon request can be successfully persisted to the MessageStore.

chrjohn:
Thanks for the hint @prerak-proof, you are absolutely right.
So the complete `SessionDisconnectTest` could be deleted since it only has the mentioned test method. Do you want to open a PR for that? Otherwise I'll delete the test.

prerak-proof:
@chrjohn - PR created: https://github.com/quickfix-j/quickfixj/pull/278

PR #246, opened 2019-12-04
Fix Sending Messages Example Code in User Manual

gemivnet:
There seems to have been a side effect from the QuickFix C++ Manual. I've updated the example code which was not changed to Java.


chrjohn:
Thanks, good catch. :+1:

PR #255, opened 2020-01-07
Introduce separate thread pool for establishing Initiator connections

### LINKED ISSUE #254: Connection timeout on one Session will block timer-related tasks on other Sessions (cf. QFJ-291)
On some of our FIX nodes we noticed that sometimes Heartbeats weren't sent out when they were due. After analysis and taking some stack dumps we noticed that QFJ was trying to connect to a FIX session at that time. But the connection attempt was not refused but timed out. This will block the thread (running on a `SingleThreadedExecutor`) for a maximum of 2 seconds.

https://github.com/quickfix-j/quickfixj/blob/b40c9a05403e7acd68990b16ec280b01107c7098/quickfixj-core/src/main/java/quickfix/mina/initiator/IoSessionInitiator.java#L230-L232

There already was a JIRA issue for this (https://www.quickfixj.org/jira/browse/QFJ-291) which introduced the 2 second timeout. Before it probably waited until the network timeout triggered.

The same `SingleThreadedExecutor` is also responsible to call the `Session.next()` method (which among other things is responsible to generate Heartbeats) for **all** Sessions. 

As a first step we will introduce a separate Executor with a few threads that will deal with the Initiator connection attempts. This will
 * enable concurrent establishment of Sessions even if one or two are timing out when trying to connect.
 * not block calling `Session.next()` for other Sessions.
  (Later we might also upgrade the `SingleThreadedExecutor` to have more threads. But we need to take care that `Session.next()` gets called always by the same thread from the pool to prevent possible concurrency/visibility issues)

The latter point is also tracked in https://www.quickfixj.org/jira/browse/QFJ-555.

Will open a PR for this shortly. Just opening this issue so I don't forget about it.

chrjohn:
Fixes #254

 * introduced separate thread pool with 3 threads for connection establishment
 * changed `enabled` flag in `Session` to `volatile` and removed synchronization from `setEnabled`/`isEnabled` since I could not find any good reason why it was synchronized
 ** `volatile` should be enough to ensure that all threads see the current state now that it is checked from distinct threads
 * removed `synchronized` from `IoSessionInitiator.ConnectTask.run()` since I could not find any good reason why it was synchronized

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/mina/SessionConnector.java
+++ b/quickfixj-core/src/main/java/quickfix/mina/SessionConnector.java
@@ -71,7 +71,7 @@ public abstract class SessionConnector implements Connector {
     private final Map<SessionID, Session> sessions = new ConcurrentHashMap<>();
     private final SessionSettings settings;
     private final SessionFactory sessionFactory;
-    private final static ScheduledExecutorService scheduledExecutorService = Executors
+    private static final ScheduledExecutorService SCHEDULED_EXECUTOR = Executors
             .newSingleThreadScheduledExecutor(new QFTimerThreadFactory());
     private ScheduledFuture<?> sessionTimerFuture;
     private IoFilterChainBuilder ioFilterChainBuilder;
@@ -318,7 +318,7 @@ public abstract class SessionConnector implements Connector {
         if (shortLivedExecutor != null) {
             timerTask = new DelegatingTask(timerTask, shortLivedExecutor);
         }
-        sessionTimerFuture = scheduledExecutorService.scheduleAtFixedRate(timerTask, 0, 1000L,
+        sessionTimerFuture = SCHEDULED_EXECUTOR.scheduleAtFixedRate(timerTask, 0, 1000L,
                 TimeUnit.MILLISECONDS);
         log.info("SessionTimer started");
     }
@@ -339,10 +339,11 @@ public abstract class SessionConnector implements Connector {
     }
 
     protected ScheduledExecutorService getScheduledExecutorService() {
-        return scheduledExecutorService;
+        return SCHEDULED_EXECUTOR;
     }
 
     private class SessionTimerTask implements Runnable {
+        @Override
         public void run() {
             try {
                 for (Session session : sessions.values()) {
@@ -411,6 +412,7 @@ public abstract class SessionConnector implements Connector {
 
     private static class QFTimerThreadFactory implements ThreadFactory {
 
+        @Override
         public Thread newThread(Runnable runnable) {
             Thread thread = new Thread(runnable, "QFJ Timer");
             thread.setDaemon(true);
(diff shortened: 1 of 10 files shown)


quickfixj-core/src/main/java/quickfix/mina/initiator/AbstractSocketInitiator.java
```diff
@@ -69,9 +76,27 @@ protected AbstractSocketInitiator(Application application,
...
+    }
+
+    protected AbstractSocketInitiator(SessionSettings settings, SessionFactory sessionFactory, int numReconnectThreads)
+            throws ConfigError {
         super(settings, sessionFactory);
         IoBuffer.setAllocator(new SimpleBufferAllocator());
         IoBuffer.setUseDirectBuffer(false);
+        if (numReconnectThreads > 0) {
+            scheduledReconnectExecutor = Executors.newScheduledThreadPool(numReconnectThreads, new QFScheduledReconnectThreadFactory());
+            ((ThreadPoolExecutor) scheduledReconnectExecutor).setMaximumPoolSize(numReconnectThreads);
```
chrjohn, line 96: Setting `maximumPoolSize` to `initialPoolSize` ensures that `numReconnectThreads` threads are available to service tasks. Otherwise new threads only get created when the task queue is full.

quickfixj-core/src/main/java/quickfix/mina/SessionConnector.java
```diff
@@ -318,7 +318,7 @@ protected void startSessionTimer() {
         if (shortLivedExecutor != null) {
             timerTask = new DelegatingTask(timerTask, shortLivedExecutor);
         }
-        sessionTimerFuture = scheduledExecutorService.scheduleAtFixedRate(timerTask, 0, 1000L,
+        sessionTimerFuture = SCHEDULED_EXECUTOR.scheduleAtFixedRate(timerTask, 0, 1000L,
```
the-thing, line 321: Not directly related to this change, but we probably don't want to queue up timer tasks if they take too long for some reason.

https://github.com/quickfix-j/quickfixj/blob/e53d41bc6b7a51ae02a63b7b0222846d248dcb7c/quickfixj-core/src/main/java/quickfix/mina/initiator/IoSessionInitiator.java#L364

chrjohn: The `reconnectTasks` timed out after a maximum of 2000ms. That was the reason to have a bigger pool to service these requests, see #254 .
What do you suggest to prevent piling up these tasks?

the-thing: Similar to `IoSessionInitiator`. As a rule of thumb `scheduleWithFixedDelay` is better in majority of cases than `scheduleAtFixedRate`. It respects the delay, but it only reschedules when the tasks finishes - this does not allow bursts if for some reason task is slow.

PR #259, opened 2020-02-20
Fix variable interpolation with custom properties for SessionID

MAXbrainRUS:
There is no way to resolve variableValues from custom place (without using system properties) for SessionId fields. But they are used for FIX session creation.


quickfixj-core/src/test/java/quickfix/SessionSettingsTest.java
```diff
@@ -537,4 +599,8 @@ public void testConcurrentAccess() throws ConfigError, InterruptedException {
         return defaultSettings;
     }
 
+    protected String getConfigurationFileName() {
```
philipwhiuk: Make this private - it's not used outside this class.

quickfixj-core/src/main/java/quickfix/SessionSettings.java
```diff
@@ -86,13 +86,34 @@
     // problems with moving configuration files between *nix and Windows.
     private static final String NEWLINE = "\r\n";
 
-    private Properties variableValues = System.getProperties();
+    private Properties variableValues;
```
philipwhiuk: This is still a terrible name for a variable IMO but we could fix that in another review. 

To be completely obtuse, it's a variable property that contains contains variable properties.

`modifiableValues` would move us in the right direction. I'll take that up in another PR when this is merged.

quickfixj-core/src/test/java/quickfix/SessionSettingsTest.java
```diff
@@ -313,6 +324,57 @@ public void testVariableInterpolationWithProps() throws Exception {
                 settings.getString("VariableTest"));
     }
 
+    @Test
+    public void testVariableInterpolationWithCustomPropsForSessionIdFromInputStream() throws Exception {
+        // GIVEN
```
philipwhiuk: These comments are kind of unnecessary IMO.

quickfixj-core/src/test/java/quickfix/SessionSettingsTest.java
```diff
@@ -313,6 +324,57 @@ public void testVariableInterpolationWithProps() throws Exception {
...
+        settingsString += "BeginString=FIX.4.2\n";
+        settingsString += "TargetCompID=CLIENT3_${test.1}_${test.2}\n";
+        settingsString += "DataDictionary=../spec/FIX42.xml\n";
+
+        // WHEN
+        final SessionSettings settings = createSettingsFromString(settingsString, properties);
+
+        // THEN
+        Iterator<SessionID> sessionIDIterator = settings.sectionIterator();
+        while (sessionIDIterator.hasNext()) {
```
philipwhiuk: This method is both hard to follow and repeated.

I prefer

```java
private SessionID findSessionID(SessionSettings settings, String targetCompID) {
     while (sessionIDIterator.hasNext()) {
          if (sessionID.getTargetCompID().startsWith(sessionID)) {
                  return sessionID;
          }
     }
     return null;
}
```
then

```java
SessionID sessionID = findSessionID(settings, "CLIENT3");
assertNotNull("Settings for CLIENT3 not found", sessionID);
assertEquals("Wrong TargetCompID", "CLIENT3_FOO_BAR", sessionID.getTargetCompID());
```

This abstracts the nasty Iterator iteration out the test and makes the test closer to the arrange, act, assert model - which is what your comments are really saying.

quickfixj-core/src/test/java/quickfix/SessionSettingsTest.java
MAXbrainRUS: Fixed

quickfixj-core/src/test/java/quickfix/SessionSettingsTest.java
MAXbrainRUS: Fixed

quickfixj-core/src/test/java/quickfix/SessionSettingsTest.java
MAXbrainRUS: Fixed

MAXbrainRUS:
Shall I do something more?

kevinstembridge:
Looking forward to this one being merged and released.

Thanks for the work @MAXbrainRUS.

chrjohn:
Thanks for the PR @MAXbrainRUS and thanks for the review @philipwhiuk :+1: 
@kevinstembridge : slowly working my way through ;)

pyy07:
how can I inject environment variable into sessionSettings?

chrjohn:
@pyy07 could you please open a discussion the next time? That will keep the issues/PR clean.

I think it should work like outlined in this test case:
https://github.com/quickfix-j/quickfixj/blob/e612ee6858673e6d4e1e21999fbda105e9c283bc/quickfixj-core/src/test/java/quickfix/SessionSettingsTest.java#L361

Where `CLIENT_PLACEHOLDER1` and `CLIENT_PLACEHOLDER2` are used in the config file https://github.com/quickfix-j/quickfixj/blob/e612ee6858673e6d4e1e21999fbda105e9c283bc/quickfixj-core/src/test/resources/configWithSessionVariables.ini#L19

pyy07:
I want to know why the design chose system properties instead of environment variables. I am using quickfixj-spring-boot-starter and there is no way to inject custom properties. However, my runtime environment only provides the option of using environment variables.

PR #261, opened 2020-02-25
Only check for ApplVerID on Logon and application messages

### LINKED ISSUE #260: We incorrectly check for ApplVerID on admin messages
**Describe the bug**

We check for ApplVerID even on session level (admin) messages. We only exclude logon from this.
Per the FIXT.1.1 spec session level messages aren't versioned using the ApplVerID. So we shouldn't check them.

We don't even need the ApplVerID for the session level messages as we then use the session dictionary for validating the message.

**To Reproduce**

Apply `MessageUtils.parse` on a logout message.

**Expected behavior**

We should process the message regardless.

**Actual behavior***

We throw an `InvalidMessage` Exception in `MessageUtils.getApplVerID` 

https://github.com/quickfix-j/quickfixj/blob/master/quickfixj-core/src/main/java/quickfix/MessageUtils.java#L136

https://github.com/quickfix-j/quickfixj/blob/master/quickfixj-core/src/main/java/quickfix/MessageUtils.java#L185

**system information:**
 - OS: Linux
 - Java version: Java 8
 - QFJ Version: 2.2.0-SNAPSHOT

**Additional context**

This is commonly seen in the real world in the following scenario:

1. Logon
2. Logout with no ApplVerID due to bad sequence numbers with no default set for the session
3. `InvalidMessage` thrown

philipwhiuk:
Fixes #260

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/MessageUtils.java
+++ b/quickfixj-core/src/main/java/quickfix/MessageUtils.java
@@ -131,11 +131,11 @@ public class MessageUtils {
         final String msgType = getMessageType(messageString);
         final MessageFactory messageFactory = session.getMessageFactory();
         final DataDictionaryProvider ddProvider = session.getDataDictionaryProvider();
-        ApplVerID applVerID = null;
+        final ApplVerID applVerID;
         final DataDictionary sessionDataDictionary = ddProvider == null ? null : ddProvider
                 .getSessionDataDictionary(beginString);
-        quickfix.Message message;
-        DataDictionary payloadDictionary;
+        final quickfix.Message message;
+        final DataDictionary payloadDictionary;
 
         if (!isAdminMessage(msgType) || isLogon(messageString)) {
             if (FixVersions.BEGINSTRING_FIXT11.equals(beginString)) {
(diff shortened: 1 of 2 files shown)


quickfixj-core/src/main/java/quickfix/MessageUtils.java
```diff
@@ -131,11 +131,11 @@ public static Message parse(Session session, String messageString) throws Invali
...
         final MessageFactory messageFactory = session.getMessageFactory();
         final DataDictionaryProvider ddProvider = session.getDataDictionaryProvider();
-        ApplVerID applVerID = null;
+        final ApplVerID applVerID;
         final DataDictionary sessionDataDictionary = ddProvider == null ? null : ddProvider
                 .getSessionDataDictionary(beginString);
-        quickfix.Message message;
-        DataDictionary payloadDictionary;
+        final quickfix.Message message;
+        final DataDictionary payloadDictionary;
```
chrjohn, line 138: Added some`final` here to make it look similar to the surrounding code.

quickfixj-core/src/main/java/quickfix/MessageUtils.java
```diff
@@ -145,18 +145,18 @@ public static Message parse(Session session, String messageString) throws Invali
...
+        } else {
+            applVerID = null;
             payloadDictionary = sessionDataDictionary;
-       }
+        }
 
         final boolean doValidation = payloadDictionary != null;
         final boolean validateChecksum = session.isValidateChecksum();
 
+        message = messageFactory.create(beginString, applVerID, msgType);
```
chrjohn, line 159: Pulled this out of the if-else-block since the assignment was the same.

chrjohn:
Hi @philipwhiuk, I have added one commit. If it is OK with you I am going to merge this.

PR #264, opened 2020-03-11
QFJ-905 - adding support for MULTIPLECHARVALUE

the-thing:
**Changes**

- added support for MULTIPLECHARVALUE

There is a discrapency between wiki and actual FIXML schema "fixml-datatypes". The regex in the schema considers all white spaces to be a valid separator, while FIX wiki states that characters should be separated by single space.

I used single space since `quickfix.DataDictionary#isFieldValue` alerady expects single space character`.

Validation does not care about same character repeated more than once.

http://fixwiki.org/fixwiki/MultipleCharValueDataType
https://www.fixtrading.org/standards/fixml/


chrjohn:
Thanks for the PR! Could you please change to use specific imports instead of wildcard imports. Sorry to bug you about this. Will look at Checkstyle or some other tool to automate it.

As far as I see this change is backward compatible (as in it would be OK to go to QFJ 2.2.0)?

Cheers,
Chris.

the-thing:
Changed the imports to what they were before. Yes, this is a new feature so existing code should not be affected. 

I don't think there is a need to cherry pick it into older branches unless somebody specifically needs  it.

Thanks,
Marcin

quickfixj-core/src/main/java/quickfix/field/converter/CharArrayConverter.java
```diff
@@ -0,0 +1,40 @@
...
+
+    private static final String CHAR_ARRAY_REGEX = "[A-Za-z0-9]( [A-Za-z0-9])*";
+
+    public static String convert(char... chars) {
+        if (chars.length == 0) {
+            throw new IllegalArgumentException("empty character array");
+        }
+
+        StringBuilder builder = new StringBuilder(chars.length * 2 - 1);
+        builder.append(chars[0]);
```
chrjohn: I am asking myself if we could use `String.join()` or respectively `StringJoiner` here.  So we would not need the loop and the code would be cleaner. What do you think?

chrjohn: OK, just realized that we would need to convert the `char` to `CharSequence` or wrap it in `String.valueOf( char )'.

quickfixj-core/src/main/java/quickfix/field/converter/CharArrayConverter.java
```diff
@@ -0,0 +1,40 @@
...
+
+        for (int i = 1; i < chars.length; i++) {
+            builder.append(' ').append(chars[i]);
+        }
+
+        return builder.toString();
+    }
+
+    public static char[] convert(String value) throws FieldConvertError {
+        if (!value.matches(CHAR_ARRAY_REGEX)) {
```
chrjohn: We could use a `Matcher` here and re-use the `Pattern`. OTOH this would not be thread-safe anymore and we would need synchronization or `ThreadLocal`. Probably not worth to be changed.

the-thing: Yes. Matchers are thread safe, but patterns are not so at the end we still have to create local pattern which is essentially what String class does. Garbage creation in QuickFIX is probably one of the biggest complaint people make, but this probably requires refactoring in other places more.

quickfixj-core/src/main/java/quickfix/field/converter/CharArrayConverter.java
the-thing: I was thinking about the same, but `StringBuilder` was better.

chrjohn: Because of the needed char to String conversion? Or because of other reasons?

the-thing: Yes - because `java.util.StringJoiner#add` requires `java.lang.CharSequence`

quickfixj-core/src/main/java/quickfix/field/converter/CharArrayConverter.java
```diff
@@ -0,0 +1,40 @@
+package quickfix.field.converter;
+
+import quickfix.FieldConvertError;
+
+/**
+ * Converts between character array and string.
+ */
+public class CharArrayConverter {
+
+    private static final String CHAR_ARRAY_REGEX = "[A-Za-z0-9]( [A-Za-z0-9])*";
```
chrjohn: I think this regex needs to be changed to comply to the definition as per the FIX spec of `character` since it would not match punctuation: https://fiximate.fixtrading.org/en/FIX.5.0SP2_EP255/fix_datatypes.html

> Single character value, can include any alphanumeric character or punctuation except the delimiter. All char fields are case sensitive (i.e. m != M).

TBH there seems to be no field which has a char value which would not match that regex. However, I think we should adhere to the spec here.

charlesbr1: "Single character value, can include any alphanumeric character or punctuation except the delimiter. All char fields are case sensitive (i.e. m != M)."

It would make sens that a char is any ascii char at the end, except the | delimiter of course...

the-thing: Have a look at the latest commit guys. I added more specific test cases and I think that's what you both guys proposing.

I kept the separator to single space character because `quickfix.DataDictionary#isFieldValue` supports it.

chrjohn:
Hi Marcin, many thanks for your efforts and the additional test coverage!

PR #148, opened 2017-11-03
Required component in group

mmitya:
This PR fixes the following problem:
- Field "F" belongs to component "C"
- Field "F" is marked as req'd within the component "C"
- Component "C" belongs to repeating group "G"
- Component "C" is marked as req'd within the group "G"

Expected: field "F" is req'd within the group "G", but validation ignores it.

For my own reference: PROM-4700.

PR #266, opened 2020-03-11
QFJ-951 - test case covering JIRA description

the-thing:
I'm going through open JIRAs to see what still needs fixing.

Just added a unit test case suggested in the JIRA. The issue was fixed by https://github.com/quickfix-j/quickfixj/pull/148.

I think the JIRA ticket should be marked as fixed in QFJ 2.1.0


chrjohn:
Thanks. :+1:

PR #277, opened 2020-03-27
Add optional TLS SNI initiator support

### LINKED ISSUE #276: Add TLS SNI support
Quickfix/J now supports SSL, which is great, but to fully leverage this when running a mutli-client server in a managed (Kuberenetes/Traefik2) environment (and many others), one can't always allocate a new port for each new session, and instread 3rd+ level hostnames are used to distinguish. For this to work, a SNI header should be added to the initiator connection (JDK fully supports this since version 7 I think).
https://en.wikipedia.org/wiki/Server_Name_Indication

A SNI header should be added to the initiator connection (JDK fully supports this since version 7 I think).

Alternative is to use something like stunnel, or allocate ports per connection, neither of which is nice, given that we have the differentiating mechanism already available.

**Additional context**
SNI shoul be enabled by default when using TLS -- servers which don't support it simply ignore it.

youurayy:
Fixes #276.


the-thing:
Is there a chance to wring unit test for it? I guess doing end to end unit test is not possible without setting up virtual hosts, isn't it?

chrjohn:
I am not familiar with this, but adding a flag to disable SNI is not necessary because it will be ignored by counterparties not supporting it?? Really really sure? ;)

quickfixj-core/src/main/java/quickfix/mina/ssl/SSLFilter.java
```diff
@@ -49,4 +58,20 @@ public void setEnabledCipherSuites(String[] cipherSuites) {
...
+    @Override
+    public void onPreAdd(IoFilterChain parent, String name, NextFilter nextFilter)
+        throws SSLException {
+
+        IoSession session = parent.getSession();
+        SocketAddress remoteAddress = session.getRemoteAddress();
+
+        if(remoteAddress instanceof InetSocketAddress) {
+            // activate the SNI support in the JSSE SSLEngine
+            log.info("activating TLS SNI support");
```
the-thing: I would log the remote peer address being set.

youurayy: debug or info ?

youurayy:
> I am not familiar with this, but adding a flag to disable SNI is not necessary because it will be ignored by counterparties not supporting it?? Really really sure? ;)

I'm pretty sure, but yeah, still. Will make it opt-in then?

quickfixj-core/src/main/java/quickfix/mina/ssl/SSLFilter.java
the-thing: I would keep the current info and just added the param.

quickfixj-core/src/main/java/quickfix/mina/ssl/SSLFilter.java
```diff
@@ -49,4 +59,26 @@ public void setEnabledCipherSuites(String[] cipherSuites) {
...
+    public void onPreAdd(IoFilterChain parent, String name, NextFilter nextFilter)
+        throws SSLException {
+
+        if (useSNI) {
+            IoSession session = parent.getSession();
+            SocketAddress remoteAddress = session.getRemoteAddress();
+
+            if (remoteAddress instanceof InetSocketAddress) {
+                // activate the SNI support in the JSSE SSLEngine
+                log.info("activating TLS SNI support for peer address: " + remoteAddress);
```
the-thing: No need for +, you can use the argument

quickfixj-core/src/main/java/quickfix/mina/ssl/SSLFilter.java
```diff
@@ -49,4 +59,26 @@ public void setEnabledCipherSuites(String[] cipherSuites) {
     public void setCipherSuites(String[] cipherSuites) {
         super.setEnabledCipherSuites(cipherSuites);
     }
+
+    @Override
+    public void onPreAdd(IoFilterChain parent, String name, NextFilter nextFilter)
+        throws SSLException {
+
+        if (useSNI) {
```
the-thing: This can be combined with  below if statement.

youurayy: Yes it can, but it will diminish the clarity of intent.

quickfixj-core/src/main/java/quickfix/mina/ssl/SSLFilter.java
youurayy: That doesn't seem to be the style established in the project (or, I don't understand what you mean by "argument")
image

the-thing: I think it is mixed, depends which class you look at e.g. `quickfix.mina.SingleThreadedEventHandlingStrategy#blockInThread`. If you keep + I will not complain.

quickfixj-core/src/main/java/quickfix/mina/ssl/SSLFilter.java
youurayy: I mean, the check on the enabling parameter is at a different "logical level" than the safety `instanceof` check, and mixing them may confuse code readers due to this. I can change if you insist.

youurayy: also, the `remoteAddress` local variable is introduced is introduced within the `if (useSNI) {` block, so to have it participate in the top-most condition would require a hacky style, e.g. `if (useSNI && ((remoteAddress = session.getRemoteAddress()) instanceof InetSocketAddress)) {`.
Similarly for the `session` local variable.

quickfixj-core/src/main/java/quickfix/mina/ssl/SSLFilter.java
youurayy: Changed.

quickfixj-core/src/main/java/quickfix/mina/ssl/SSLFilter.java
the-thing: I see the address. Keep it then.

youurayy:
Not sure what to do with the failing LGTM for JavaScript, the [log](https://lgtm.com/projects/g/quickfix-j/quickfixj/logs/rev/pr-f01d2ad0e71cb551a5e4932c72dacee7db2e21a5/lang:javascript/stage:Build%20merge_1d307fde52f335f95c3137443d19df75ec01a887) says:

image

I assume this is because `configuration.html` was edited, and LGTM cannot extract any JavaScript from it to perform checks on.

One way to solve this would be to add a `./lgtm.yaml` and `exclude` the `doc`  dir, since no files there use any JavaScript:
```yaml
extraction:
  javascript:
    index:
      exclude:
        - quickfixj-core/scr/main/doc
      filters:
        - exclude: "**/*.html"
```
as per https://lgtm.com/help/lgtm/javascript-extraction.

Not sure if that should be part of this PR, let me know pls.

chrjohn:
Please add the LGTM config. IMHO you could also disable JavaScript checking altogether, but don't know if that is possible.

chrjohn:
JavaScript analysis failed again. But for me this should not block the PR from merging. If you don't have any idea how to fix it let me know.

youurayy:
@chrjohn I'm in the process of trying to figure it out -- don't want to be guessing so I'm reading their docs now. There's lot of `.js` files under `/dist/tools/..` (they seem correctly ignored), so I think simply excluding all *.html won't do it. I'll try to commit a fix soon.

youurayy:
It seems that the LGTM JavaScript extractor will always complain if it fails to extract at least one file: https://github.com/Semmle/ql/blob/master/javascript/extractor/src/com/semmle/js/extractor/AutoBuild.java#L1088

chrjohn:
We don't have any lgtm.yml in place yet so it seems that by default JavaScript is tested. Could we only specify Java to skip JS extraction? Cf here: https://github.com/elastic/apm-agent-java/blob/master/.lgtm.yml

chrjohn:
Dang..

youurayy:
Okay, giving up.

Somebody should check if it's possible to turn off the LGTM JavaScript integration in the repo settings.

Alternatively, the `lgtm.yml` could perhaps be tweaked using the `path_classifiers:` section, but this change will not be applied from a PR, must be applied on the repo directly:
image


The last `lgtm.yml` I tried was:
```yaml
extraction:
  javascript:
    before_index:
      - export LGTM_TRAP_CACHE=
    index:
      exclude:
        - quickfixj-core
        - quickfixj-codegenerator
      filters:
        - exclude: "*.html"
        - exclude: "*.js"
        - exclude: "*.yml"
        - exclude: "**/*.html"
        - exclude: "**/*.js"
        - exclude: "**/*.yml"
      typescript: none
```
(the `LGTM_TRAP_CACHE` var reset was an attempt to [disable the externs extraction](https://github.com/Semmle/ql/blob/master/javascript/extractor/src/com/semmle/js/extractor/AutoBuild.java#L499))

What was weird, the extraction was failing even when a phony `<script>..</script>` block was inserted into one of the files which were 100% processed by the extractor.

chrjohn:
I'll check with LGTM to disable the JavaScript support since I cannot disable it in the project settings. Seems to be a known problem or missing feature: https://discuss.lgtm.com/t/how-to-turn-off-language-python-support-for-the-c-project/1388 and https://discuss.lgtm.com/t/disable-javascript-analysis/2150/15

Apart from that I think we are ready to merge this. Thanks for the contribution @youurayy :+1: 
Thank you @the-thing for your review and comments here and on #276 :+1:

PR #278, opened 2020-03-29
Remove unit test obsoleted by change in the order of 'persist' and 'send'

prerak-proof:
The unit test was obsoleted due to this PR: https://github.com/quickfix-j/quickfixj/pull/242

The condition described in the test is no longer possible:
>The purpose of this test is to simulate the situation where a logon response is received before a logon request can be successfully persisted to the MessageStore.


chrjohn:
Thanks for the PR! Good catch :+1:

PR #280, opened 2020-04-08
Add method to DataDictionary to get the value for a given enum value name

chrjohn:
- this is useful for mapping from/to the XML representation obtained
from `Message.toXML()`
 - since the DataDictionary is kind of a memory hog we do not store the
values in yet another HashMap (which would cause Node objects to be created) but iterate over the entries on each call to `getValue(int, String)`
  -- it is up to the user's application to cache the returned values if desired


chrjohn:
@the-thing , thanks for the review.

PR #283, opened 2020-04-27
Optimise session reset to reduce Calendar creation

### LINKED ISSUE #282: There's considerable idle memory churn due to the construction of `Calendar` objects for reset calculations
**Describe the bug**
An idle QuickFIXJ application produces a stream of garbage to be collected in the form of throwaway (Gregorian)Calendar objects

**To Reproduce**
* Create a QFJ application that has a Session attached
* Observe the memory behaviour and see what objects are being allocated and garbage collected.

**Expected behaviour**
* Minimal (ideally 0) use of additional memory without traffic being passed

**System information:**
 - OS independent
 - Java version 8
 - QFJ Version - 2.2.0-SNAPSHOT

**Additional context**
The source of this is the DefaultSessionSchedule class which creates (and invokes creation of) Calendar objects every time it's called.

For example, the `MessageStore` implementations all retain a `Calendar` object for the creationTime. However the API exposes a `getCreationTime()` method - and so the store constructs a `DateTime` from the calendar. But `DefaultSessionSchedule` uses a `Calendar,` so it then reconstructs a `Calendar` object from that date time to then do the comparison.

This all happens every time `isSessionTime` is called - which (at least) is every time `Session.next()` is called - which is up to once a second.

philipwhiuk:
Fixes #282

Attempts to improve Calendar-based memory churn seen in #282 by accessing original calendars and keeping a local cache of temporary ones.

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/DefaultSessionSchedule.java
+++ b/quickfixj-core/src/main/java/quickfix/DefaultSessionSchedule.java
@@ -41,10 +41,17 @@ public class DefaultSessionSchedule implements SessionSchedule {
     private final int[] weekdayOffsets;
     protected static final Logger LOG = LoggerFactory.getLogger(DefaultSessionSchedule.class);
 
+    //Cache recent time data to reduce creation of calendar objects
+    private ThreadLocal<Calendar> threadLocalCalendar;
+    private ThreadLocal<TimeInterval> threadLocalRecentTimeInterval;
+
     public DefaultSessionSchedule(SessionSettings settings, SessionID sessionID) throws ConfigError,
             FieldConvertError {
+        threadLocalCalendar = ThreadLocal.withInitial(SystemTime::getUtcCalendar);
+        threadLocalRecentTimeInterval = new ThreadLocal<>();
+        isNonStopSession = settings.isSetting(sessionID, Session.SETTING_NON_STOP_SESSION)
+            && settings.getBool(sessionID, Session.SETTING_NON_STOP_SESSION);
 
-        isNonStopSession = settings.isSetting(sessionID, Session.SETTING_NON_STOP_SESSION) && settings.getBool(sessionID, Session.SETTING_NON_STOP_SESSION);
         TimeZone defaultTimeZone = getDefaultTimeZone(settings, sessionID);
         if (isNonStopSession) {
             isWeekdaySession = false;
@@ -104,7 +111,7 @@ public class DefaultSessionSchedule implements SessionSchedule {
     }
 
     private TimeZone getDefaultTimeZone(SessionSettings settings, SessionID sessionID)
-            throws ConfigError, FieldConvertError {
+            throws ConfigError {
         TimeZone sessionTimeZone;
         if (settings.isSetting(sessionID, Session.SETTING_TIMEZONE)) {
             String sessionTimeZoneID = settings.getString(sessionID, Session.SETTING_TIMEZONE);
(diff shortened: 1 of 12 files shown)


quickfixj-core/src/main/java/quickfix/DefaultSessionSchedule.java
```diff
@@ -41,10 +41,17 @@
     private final int[] weekdayOffsets;
     protected static final Logger LOG = LoggerFactory.getLogger(DefaultSessionSchedule.class);
 
+    //Cache recent time data to reduce creation of calendar objects
+    private ThreadLocal<Calendar> threadLocalCalendar;
```
philipwhiuk, line 45: ThreadLocal because I'm fairly sure Calendar isn't thread-safe.
This keeps a calendar which we update with the new time we're checking against.

quickfixj-core/src/main/java/quickfix/DefaultSessionSchedule.java
```diff
@@ -300,9 +307,16 @@ public boolean isSessionTime() {
         if(isNonStopSession()) {
             return true;
         }
-        Calendar now = SystemTime.getUtcCalendar();
-        TimeInterval interval = theMostRecentIntervalBefore(now);
-        return interval.isContainingTime(now);
+        Calendar now = threadLocalCalendar.get();
+        now.setTimeInMillis(SystemTime.currentTimeMillis());
+        TimeInterval mostRecentInterval = threadLocalRecentTimeInterval.get();
+        if (mostRecentInterval != null && mostRecentInterval.isContainingTime(now)) {
```
philipwhiuk, line 313: This optimisation should mean that during session time we don't have to keep reconstructing time intervals and instead just re-use one. Constructing a `TimeInterval` involves a bunch of throwaway Calendar usage.

quickfixj-core/src/main/java/quickfix/Session.java
```diff
@@ -584,7 +584,7 @@ public String getRemoteAddress() {
     private boolean isCurrentSession(final long time)
             throws IOException {
         return sessionSchedule == null || sessionSchedule.isSameSession(
-                SystemTime.getUtcCalendar(time), SystemTime.getUtcCalendar(state.getCreationTime()));
+                SystemTime.getUtcCalendar(time), state.getCreationTimeCalendar());
```
philipwhiuk, line 587: This avoids creating a calendar from a date-time that's itself created from a calendar (which is the behaviour of all the implementations of `getCreationTime()`).

quickfixj-core/src/main/java/quickfix/SessionState.java
```diff
@@ -20,10 +20,7 @@
 package quickfix;
 
 import java.io.IOException;
-import java.util.Collection;
-import java.util.Date;
-import java.util.LinkedHashMap;
-import java.util.Map;
+import java.util.*;
```
chrjohn, line 23: Wildcard import... you know... ;)

chrjohn: Sorry, did not manage to find some code style tool which checks this automatically.

chrjohn:
I've put this to 3.0.0 milestone since a new interface method was added to `MessageStore`.

chrjohn:
Closing and reopening to trigger Java CI checks.

quickfixj-core/src/main/java/quickfix/DefaultSessionSchedule.java
```diff
@@ -41,10 +41,17 @@
     private final int[] weekdayOffsets;
     protected static final Logger LOG = LoggerFactory.getLogger(DefaultSessionSchedule.class);
 
+    //Cache recent time data to reduce creation of calendar objects
+    private ThreadLocal<Calendar> threadLocalCalendar;
+    private ThreadLocal<TimeInterval> threadLocalRecentTimeInterval;
```
chrjohn, line 46:
```suggestion
    private final ThreadLocal<Calendar> threadLocalCalendar;
    private final ThreadLocal<TimeInterval> threadLocalRecentTimeInterval;
```
Could we add a `final` here to a) make it consistent with the rest of the fields and b) give a hint to the reader that these fields are initialized in the constructor?

PR #285, opened 2020-05-02
Add custom tags to Logon message.

chrjohn:
Adds possibility to specify tags via setting `LogonTag` or `LogonTag<n>`.
Example:
```
LogonTag=553=user
LogonTag1=554=password 
```

It is checked whether a data dictionary is present to determine whether the specified tag is a header field. Otherwise it is put into the body of the message.


chrjohn:
LGTM JavaScript analysis failing. This is somewhat expected. See https://github.com/quickfix-j/quickfixj/pull/277#issuecomment-607183447

chrjohn:
@the-thing , thanks for the review.
You are right, I feel like one of the most asked questions is "how do I set username and password on the Logon message". ;)

the-thing:
BTW. I think you can close QFJ-177.

PR #286, opened 2020-05-04
Support setting RefreshOnLogon for Initiators

### LINKED ISSUE #262: Setting RefreshOnLogon should also work for Initiators
Up to now, the setting RefreshOnLogon will only refresh the message store for Acceptors.

The .NET and C++ implementations do not have this restriction.

Use case: when adding a dynamic session and changing the sequence number via JMX this apparently works but when Logon is initiated it is done with sequence number 1 again.

NB: probably documentation needs to be adapted also.

chrjohn:
Fixes #262

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ QuickFIX/J
 This is the official QuickFIX/J project repository.
 
 ## intro
-QuickFIX/J is a full featured messaging engine for the FIX protocol.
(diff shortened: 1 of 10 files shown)


the-thing:
This is slightly confusing to review as the changes from other PR are here. https://github.com/quickfix-j/quickfixj/pull/285

chrjohn:
@the-thing , thanks, I have updated the PR.

PR #289, opened 2020-05-20
QFJ-684 - added charset support to `BytesField`

the-thing:
**Changes**
- `quickfix.BytesField` uses `org.quickfixj.CharsetSupport`

This is the last remaining item from QFJ-789, so possibly this could be closed as well.

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/BytesField.java
+++ b/quickfixj-core/src/main/java/quickfix/BytesField.java
@@ -19,9 +19,7 @@
 
 package quickfix;
 
-import java.io.UnsupportedEncodingException;
-
-import org.quickfixj.QFJException;
+import org.quickfixj.CharsetSupport;
 
 /**
  * BytesField enables better handling of binary data. With BytesFields binary data can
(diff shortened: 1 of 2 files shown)


chrjohn:
Closing and reopening to trigger Java CI checks.

chrjohn:
> This is the last remaining item from QFJ-789, so possibly this could be closed as well.

If I understand QFJ-789 correctly then it proposes to use a per-message encoding which is specified by tag `347/MessageEncoding` on the message. I think this is a different thing.
https://fiximate.fixtrading.org/legacy/en/FIX.5.0SP2/tag347.html

the-thing:
I haven't seen it in practice, but according to https://www.fixtrading.org/online-specification/introduction/ there is datatype "data" that should be using this field.

> Datatype data is a union of opaque data and character data in a different encoding. It is recommended that the latter case be distinguished by naming such a field as EncodedXXX, e.g. EncodedSecurityDesc(351).

This means that that 347/MessageEncoding only applies to "data" fields e.g. 446/EncodedListStatusText in FIX42 and the other field values should be ASCII encoded.

chrjohn:
> This means that that 347/MessageEncoding only applies to "data" fields

Yes, that's also what it says in the link I posted: "Type of message encoding (non-ASCII (non-English) characters) used in a message's "Encoded" fields."

I only wanted to imply that QFJ-789 cannot be closed (as opposed to what you suggested) because this functionality is not yet covered by this or other PRs.

Issue #295, opened 2020-06-04
Connecting over an authenticating HTTP proxy fails


curdreinert:
We are trying to connect over an HTTP proxy that enforces authentication, i.e. user and password need to be set. But even though we put them into the dictionary (and set the appropriate proxy type and the correct proxy host and port) the connection fails.

Snooping into the network traffic with wireshark reveals that, at first, a CONNECT request is sent by our application to the proxy. As usual, this initial request contains no authentication data, and is answered by the proxy with an 407 Authentication Required. The expected behaviour would be that a new CONNECT request is send, this time with a Basic Proxy-Authorization header containing the user and password. Instead, the connection is closed.

I ran the code in a debugger and found that the problem seems to be that in quickfix.mina.ProtocolFactory (line 145 to 150) a prefered order of the authentication methods is set into the proxyIoSession. This order, however, has NO_AUTH at the first position. So that, after receiving the 407 and when going through the possible authentication methods, Mina's HttpSmartProxyHandler first encounters a NO_AUTH and sees this as a sign that every further search for an authentication method is futile. And therefore closes the connection.

The prefered order of the authentication methods in ProtocolFactory seems to be rather obviously wrong. But instead of correcting it, you might want to just _not_ set it, in which case Mina would go with their own idea of which method to prefer, which seems to be a good idea.

I forced the prefered order list to be null in my debugger, and was able to succesfully connect through the proxy.

curdreinert:
I just noticed that at the end of the discussion of https://github.com/quickfix-j/quickfixj/pull/92 the same issue with pretty much the same explanation was mentioned. I can confirm that the current prefered order prevents the authentication with http proxies from working, and that setting no order at all would be better.

chrjohn:
Sounds like a simple fix to me. Maybe someone with more proxy experience than me can comment whether this change will break something else.
Maybe we should make it configurable?

chrjohn:
Will removing the prefered order also work with a proxy without username and password?

curdreinert:
The whole method autoSelectAuthHandler in HttpSmartProxyHandler.java is only ever reached after the 407, i.e. only if the proxy needs authentication.
So yes, removing the prefered order also works with a proxy without username and password.

the-thing:
Are you guys working on it or planning to or can I pick it up?

chrjohn:
@the-thing , feel free to do it. Thanks in advance. :)

the-thing:
[...]

PR #296, opened 2020-06-05
Preferred proxy authentication methods removed

the-thing:
Fixes https://github.com/quickfix-j/quickfixj/issues/295

Opening as a DRAFT before we gather the test evidence.

@curdreinert

Would you be able to build or test directly from branch? At the moment I'm using CCProxy for Windows, but I struggle to get either SOCKS or HTTP method to work (most likely a setup issue). Also, what OS and proxy software are you using? I will have a look again in the evening/tomorrow.

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/mina/ProtocolFactory.java
+++ b/quickfixj-core/src/main/java/quickfix/mina/ProtocolFactory.java
@@ -141,13 +141,6 @@ public class ProtocolFactory {
         }
 
         ProxyIoSession proxyIoSession = new ProxyIoSession(proxyAddress, req);
-
-        List<HttpAuthenticationMethods> l = new ArrayList<>();
-        l.add(HttpAuthenticationMethods.NO_AUTH);
-        l.add(HttpAuthenticationMethods.DIGEST);
-        l.add(HttpAuthenticationMethods.BASIC);
-
-        proxyIoSession.setPreferedOrder(l);
         connector.setProxyIoSession(proxyIoSession);
 
         return connector;
--- a/quickfixj-core/src/test/java/quickfix/mina/ProtocolFactoryTest.java
+++ b/quickfixj-core/src/test/java/quickfix/mina/ProtocolFactoryTest.java
@@ -1,0 +1,49 @@
+/*******************************************************************************
+ * Copyright (c) quickfixengine.org  All rights reserved.
+ *
+ * This file is part of the QuickFIX FIX Engine
+ *
+ * This file may be distributed under the terms of the quickfixengine.org
+ * license as defined by quickfixengine.org and appearing in the file
+ * LICENSE included in the packaging of this file.
+ *
+ * This file is provided AS IS with NO WARRANTY OF ANY KIND, INCLUDING
+ * THE WARRANTY OF DESIGN, MERCHANTABILITY AND FITNESS FOR A
+ * PARTICULAR PURPOSE.
+ *
+ * See http://www.quickfixengine.org/LICENSE for licensing information.
+ *
+ * Contact ask@quickfixengine.org if any conditions of this licensing
+ * are not clear to you.
+ ******************************************************************************/
+
+package quickfix.mina;
+
+import org.apache.mina.core.service.IoConnector;
+import org.apache.mina.proxy.ProxyConnector;
+import org.apache.mina.proxy.session.ProxyIoSession;
+import org.apache.mina.transport.socket.SocketConnector;
+import org.apache.mina.util.AvailablePortFinder;
+import org.junit.Test;
+import quickfix.ConfigError;
+
+import java.net.InetSocketAddress;
+
+import static org.junit.Assert.assertNull;
+
+public class ProtocolFactoryTest {
+
+    @Test
+    public void shouldCreateProxyConnectorWithoutPreferredAuthOrder() throws ConfigError {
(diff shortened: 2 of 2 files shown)


curdreinert:
Took me some time, but managed to build and test directly from branch. And learned some things on the way :-)
I can confirm that with this fix
- I could succesfully connect through an authenticating HTTP proxy. Tested with a Squid (3.5.20) running on Linux (CentOS Linux release 7.8.2003 (Core)), the QuickFIX/J application was running on Windows 10
- I could also connect through an non-authenticating HTTP proxy. Tested with the same setup (Squid, Linux, Windows as above), only with the authentication turned off in the proxy.

I checked the traffic with Wireshark to confirm that in both cases actually happened what was expected (i.e. CONNECT -> 407 -> CONNECT with BASIC credentials -> Connection esatblished in the first case and CONNECT -> Connection established in the second case).

curdreinert:
As the Mina documentation of ProxyIOSession.getPreferedOrder() points out, the order is only used for http proxies (and you can confirm that by looking into their source code). However, for good measure, I also tried against an authenticating Socks 5 Proxy (some badly misused variant of jsocks), and this also works.
I'm lacking the imagination why this fix should yield problems on other operating systems or with other Http Proxies. So from my point of view, I think we are fine.
The only thing I can think of is that the new test in the ProtocolFactoryTest sets the proxy type to "socks" which might suggest to the casual reader that the necessity of not setting the preffered order has something to do with socks, while it only comes into play for http proxies. But this is rather cosmetically.

chrjohn:
@the-thing  @curdreinert 

Great, thanks guys!

the-thing:
@curdreinert 

Good work and thanks for the suggestion with the unit test - I've changed it.

SOCKS proxy handshake seems to be different. According to RFC https://tools.ietf.org/html/rfc1928 for SOCKS proxy client advertises available methods and the server chooses which one to use. Current MINA version advertises NO_AUTH, GSSAPI_AUTH, BASIC_AUTH statically, order does not matter for the client (but I assume it might be configured withing the proxy server itself). For HTTP proxy server returns methods and client chooses how to authenticate.

org.apache.mina.proxy.handlers.socks.Socks5LogicHandler#writeRequest

sachin681:
Can you please provide more details with some sample of configurations you tried with Squid ? What the values used at QFJ end? Can we do same with Apache proxy ?

sachin681:
> Took me some time, but managed to build and test directly from branch. And learned some things on the way :-)
> I can confirm that with this fix
> 
> * I could succesfully connect through an authenticating HTTP proxy. Tested with a Squid (3.5.20) running on Linux (CentOS Linux release 7.8.2003 (Core)), the QuickFIX/J application was running on Windows 10
> * I could also connect through an non-authenticating HTTP proxy. Tested with the same setup (Squid, Linux, Windows as above), only with the authentication turned off in the proxy.
> 
> I checked the traffic with Wireshark to confirm that in both cases actually happened what was expected (i.e. CONNECT -> 407 -> CONNECT with BASIC credentials -> Connection esatblished in the first case and CONNECT -> Connection established in the second case).

Can you please provide more details with some sample of configurations you tried with Squid ? What the values used at QFJ end? Can we do same with Apache proxy ?

chrjohn:
@sachin681 Please don't use the bug tracker for help requests. See https://github.com/quickfix-j/quickfixj#questions

sachin681:
> https://github.com/quickfix-j/quickfixj#questions

Sorry for posting it here. I will post the question on the right forum.

curdreinert:
Hi,

coming from a working squid configuration, we added more or less the 
following lines:

auth_param basic program /usr/lib64/squid/basic_ncsa_auth 
/etc/squid/passwd
auth_param basic children 5
auth_param basic realm Squid proxy-caching web server
auth_param basic credentialsttl 2 hours
auth_param basic casesensitive on
acl ncsa_users proxy_auth REQUIRED
http_access allow ncsa_users

So we set up an access control list (acl) called ncsa_users for which 
proxy authentication was required, and allowed access only for that acl. 
The authentication parameters say that we use a basic authentication, that 
the user id and password are stored in a file called /etc/quid/passwd and 
that the tool basic_ncsa_auth (see 
https://manpages.debian.org/testing/squid/basic_ncsa_auth.8.en.html) is 
used to verify against that file. I'm not sure what the "children" and 
"realm" parameters are for, I guess the guy who set it up copied it from 
some online how to. TTL for the credentials and casesensitiveness should 
be obvious.

On the Quickfix side, we set ProxyType to http, ProxyHost and ProxyPort to 
point to the Squid proxy, and ProxyUser and ProxyPassword as needed by 
Squid.

My guess would be that you set up an authenticating web proxy with Apaches 
mod_proxy module, but I've never done it and can't tell you how to do it.

Best regards

Curd




Von:    "sachin681" <notifications@github.com>
An:     "quickfix-j/quickfixj" <quickfixj@noreply.github.com>
Kopie:  "curdreinert" <curd.reinert@ppi.de>, "Mention" 
<mention@noreply.github.com>
Datum:  29.09.2020 19:17
Betreff:        Re: [quickfix-j/quickfixj] Preferred proxy authentication 
methods removed (#296)



Can you please provide more details with some sample of configurations you 
tried with Squid ? What the values used at QFJ end? Can we do same with 
Apache proxy ?
?

sachin681:
Thanks a lot for your detailed response.

We will try out following steps.

PR #302, opened 2020-06-15
QFJ-741 - missing value in AllocCancReplaceReason in FIX44.xml

the-thing:
Fixes QFJ-741

Seems like 99 - Other should be available since FIX 4.4

http://fixwiki.org/fixwiki/AllocCancReplaceReason/99_Other


chrjohn:
Hi @the-thing , thanks for the PR. Do you have something else that could go into 2.2.0? I want to build the release in a few hours. But of course it could only be small changes like this one. Everything else needs to go to the next release.

the-thing:
No, I'm good. I don't mind waiting if anything else comes along. Thanks.

PR #303, opened 2020-06-24
Integration with JaCoCo and Coveralls

the-thing:
I was able to integrate with JaCoCo code coverage and Coveralls plugins. The configuration excludes auto generated code so it covers only quickfixj-core logic (as it should). The build takes 30 seconds more on Travis - no impact on local builds as the goals are not executed.

Since the project is public there are no tokens required etc. The only post requisite is to login to Coveralls and add quickfix-j/quickfixj repo via QuickFIX/J organization from one of the members. Coverage status badge already points to it.

At the moment coverage is 78% which will make the badge red (> 80% orange, > 90% green), but I will be happy to improve the coverage later on.

Please have a look at my branch view in Coveralls.
https://coveralls.io/github/the-thing/quickfixj

There are also other code coverage projects such as Code Climate, but they do more than that. Coveralls seems to the most feasible for pure code coverage.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -87,6 +87,8 @@
 		<maven-deploy-plugin-version>2.8.2</maven-deploy-plugin-version>
 		<nexus-staging-maven-plugin-version>1.6.8</nexus-staging-maven-plugin-version>
 		<build-helper-maven-plugin-version>3.0.0</build-helper-maven-plugin-version>
+		<coveralls-maven-plugin-version>4.3.0</coveralls-maven-plugin-version>
+		<jacoco-maven-plugin-version>0.8.2</jacoco-maven-plugin-version>
 	</properties>
 
 	<build>
@@ -222,6 +224,24 @@
 							</goals>
 						</execution>
 					</executions>
+				</plugin>
+				<plugin>
+					<groupId>org.jacoco</groupId>
+					<artifactId>jacoco-maven-plugin</artifactId>
+					<version>${jacoco-maven-plugin-version}</version>
+					<executions>
+						<execution>
+							<id>prepare-agent</id>
+							<goals>
+								<goal>prepare-agent</goal>
+							</goals>
+						</execution>
+					</executions>
+				</plugin>
+				<plugin>
+					<groupId>org.eluder.coveralls</groupId>
+					<artifactId>coveralls-maven-plugin</artifactId>
+					<version>${coveralls-maven-plugin-version}</version>
 				</plugin>
 			</plugins>
 		</pluginManagement>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -232,6 +234,30 @@
...
+					<execution>
+						<id>prepare-agent</id>
+						<goals>
+							<goal>prepare-agent</goal>
+						</goals>
+					</execution>
+				</executions>
+				<configuration>
+					<excludes>
+						<exclude>quickfix/field/**</exclude>
```
chrjohn, line 251: There are converters in quickfix/field/converter, does it make sense to test them, too?

chrjohn:
Thanks for the PR. I've authorized Coverall for this repo.

pom.xml
the-thing: You are right. They have logic, so they should be covered. I will fix that.

the-thing:
Cool.

I'm not able to see it on Coveralls though. Is this the right link?
https://coveralls.io/github/quickfix-j/quickfixj?branch=master

chrjohn:
@the-thing , should work now. Had to add it manually on Coveralls.

the-thing:
Seems to wok. Uploaded successfully to Coveralls. https://coveralls.io/github/quickfix-j/quickfixj

There is a problem with coverage for `quickfix.field.converter` classes. It seems that JaCoCo is not able to detect the common inclusion/exclusion for `quickfix.field.*` and `quickfix.field.converter.*`. I'm getting the same reports locally and I think this is because of JaCoCo plugin filtering being different.

We can either exclude fields and converters from the report or include them both. The way to fix it would be to move out converters to a different package e.g. `quickfix.converter` (package changes + pom exclusion pattern change). However, this will be a breaking change if somebody is using converters in their code directly.

the-thing:
Local coverage after moving converters.
coverage with converters

chrjohn:
@the-thing  :+1: for `quickfix.converter` package. The change should be OK for 3.0.0 milestone. However, we probably need to decouple this from the Coveralls change.

the-thing:
Ok. Leaving this PR as is.

chrjohn:
Sorry for the long delay, still busy with other stuff. :-( Will hopefully be able to get back to this soon...

the-thing:
No problem, I was on holidays myself. I'm actually more keen on merging https://github.com/quickfix-j/quickfixj/pull/305.

chrjohn:
@the-thing I see that you did this change for Travis. Would it be much work to integrate this into the Java CI jobs?

the-thing:
Not sure, but I will have a look sometime this week. Yes, this seems to be obsolete change atm.

PR #309, opened 2020-08-29
QFJ-456 - NetworkingOptions should be applied before the socket is connected

the-thing:
As mentioned in the original JIRA certain socket options set after `org.apache.mina.core.service.IoHandlerAdapter#sessionOpened` event might not be configured properly as they require to be set before call to listen() or connect() e.g. SO_REUSEADDR, SO_RCVBUF.

**Changes**

- `quickfix.mina.NetworkingOptions` will be applied before sockets around bound. 

This chagne makes "SocketReceiveBufferSize" configuration TCP handshake advertisement correct and it will allow finer tweaking based on network capacity, system requirements etc.

Increasing the "SocketReceiveBufferSize" above 64KB was working correct for **initiators** already, but was fixed as part of https://issues.apache.org/jira/projects/DIRMINA/issues/DIRMINA-1123 for **acceptors**.

Please be aware that there is still an issue for **initiators** when setting "SocketReceiveBufferSize" below 65536 which will still use default os values.

https://github.com/apache/mina/blob/faf9e5cdaf40da3a816f789bd27c95e2f3b4d354/mina-core/src/main/java/org/apache/mina/transport/socket/nio/NioSocketConnector.java#L245

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/mina/AbstractIoHandler.java
+++ b/quickfixj-core/src/main/java/quickfix/mina/AbstractIoHandler.java
@@ -102,13 +102,7 @@ public abstract class AbstractIoHandler extends IoHandlerAdapter {
     }
 
     @Override
-    public void sessionCreated(IoSession ioSession) throws Exception {
-        super.sessionCreated(ioSession);
-        networkingOptions.apply(ioSession);
-    }
-
-    @Override
-    public void sessionClosed(IoSession ioSession) throws Exception {
+    public void sessionClosed(IoSession ioSession) {
         try {
             Session quickFixSession = findQFSession(ioSession);
             if (quickFixSession != null) {
--- a/quickfixj-core/src/main/java/quickfix/mina/NetworkingOptions.java
+++ b/quickfixj-core/src/main/java/quickfix/mina/NetworkingOptions.java
@@ -19,7 +19,7 @@
 
 package quickfix.mina;
 
-import org.apache.mina.core.session.IoSession;
+import org.apache.mina.core.service.IoService;
 import org.apache.mina.core.session.IoSessionConfig;
 import org.apache.mina.transport.socket.SocketSessionConfig;
 import org.slf4j.Logger;
@@ -28,7 +28,6 @@ import quickfix.FieldConvertError;
 import quickfix.field.converter.BooleanConverter;
 import quickfix.field.converter.IntConverter;
 
-import java.net.SocketException;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Properties;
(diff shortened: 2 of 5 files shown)


the-thing:
Test evidence with packet capture.

Scenario 1

Acceptor receive buffer size: 64KB
Initiator receive buffer size: 64KB

s1

Scenario 2

Acceptor receive buffer size: 5MB
Initiator receive buffer size: 5MB

s2

Scenario 3
Acceptor receive buffer size: 2KB
Initiator receive buffer size: 2KB (**this is still not correct in MINA**)

s3

chrjohn:
Thanks for the PR.
If I understand correctly, this applies to all socket options, so also to `TCP_NODELAY` I assume?
`TCP_NODELAY` is enabled by default but due to this bug it might be that it isn't enabled?

https://github.com/quickfix-j/quickfixj/blob/82a16ce3ea7ea0cb259ab8112f3bfa918b31b558/quickfixj-core/src/main/java/quickfix/mina/NetworkingOptions.java#L83

the-thing:
According to various documents this seems to be a concern for `SO_REUSEADDR` and `SO_RCVBUF` only (documents are often unclear) . Yes, the remaining socket options can be set after the socket has been created such as `TCP_NODELAY`.

Example. Before this PR you could not disable SO_REUSEADDR for acceptors as this was explicitly set to true before system socket listen() call.

https://github.com/quickfix-j/quickfixj/blob/e24b744979263826e4785300bf3505bc2ef1c04f/quickfixj-core/src/main/java/quickfix/mina/ProtocolFactory.java#L110

quickfixj-core/src/main/java/quickfix/mina/AbstractIoHandler.java
```diff
@@ -102,13 +102,7 @@ public void exceptionCaught(IoSession ioSession, Throwable cause) throws Excepti
...
 
     @Override
-    public void sessionCreated(IoSession ioSession) throws Exception {
-        super.sessionCreated(ioSession);
-        networkingOptions.apply(ioSession);
-    }
-
-    @Override
-    public void sessionClosed(IoSession ioSession) throws Exception {
+    public void sessionClosed(IoSession ioSession) {
```
chrjohn, line 105: Is there a specific reason to remove the thrown `Exception` from the method signature? Just asking because it remains on the other methods.

the-thing: Unrelated to the change, but it was highlighted by the IDE as unused for this method. All other methods in this class that declare `throws Exception` actually do throw some sort of checked exception.

chrjohn (approved):

PR #311, opened 2020-08-31
Set UtcTimestampPrecision in codegenerator

### LINKED ISSUE #310: Support setting a default UtcTimestampPrecision in the quickfixj-codegenerator Maven plugin
When using the `quickfixj-codegenerator` Maven plugin to generate QuickFIX/J field classes for a FIX system that uses nano seconds as precision for all `UTCTIME`, `UTCTIMEONLY` and `UTCTIMESTAMP` fields, it would have bee nice to be able to set nano seconds as a default precision for all the generated `UTCTIME`, `UTCTIMEONLY` and `UTCTIMESTAMP` fields.

If this was supported, we could do `newOrderSingle.set(TransactTime(transactTime))` instead of `newOrderSingle.setUtcTimeStamp(TransactTime.FIELD, transactTime, UtcTimestampPrecision.NANOS)`.

Add a parameter, `utcTimestampPrecision`, to the `quickfixj-codegenerator` Maven plugin that takes a value from `UtcTimestampPrecision` and using that as a default in the classes `UtcTimeField`, `UtcTimeOnlyField`, `UtcTimeStampField` and their generated sub classes.

Using the `FieldMap.setUtcTimeStamp` method etc. like described above.

**Additional context**
None.

simenflatby:
Fixes #310

Added support for setting a default UtcTimestampPrecision in the
quickfixj-codegenerator Maven plugin. The default applies to the
following classes and their generated sub classes.
- UtcTimeField
- UtcTimeOnlyField
- UtcTimeStampField

### DIFF
--- a/quickfixj-codegenerator/src/main/java/org/quickfixj/codegenerator/MessageCodeGenerator.java
+++ b/quickfixj-codegenerator/src/main/java/org/quickfixj/codegenerator/MessageCodeGenerator.java
@@ -61,6 +61,7 @@ public class MessageCodeGenerator {
     private static final String BIGDECIMAL_TYPE_OPTION = "generator.decimal";
     private static final String ORDERED_FIELDS_OPTION = "generator.orderedFields";
     private static final String OVERWRITE_OPTION = "generator.overwrite";
+    private static final String UTC_TIMESTAMP_PRECISION_OPTION = "generator.utcTimestampPrecision";
 
     // An arbitrary serial UID which will have to be changed when messages and fields won't be compatible with next versions in terms
     // of java serialization.
@@ -461,6 +462,7 @@ public class MessageCodeGenerator {
                 return;
             }
 
+            String utcTimestampPrecision = getOption(UTC_TIMESTAMP_PRECISION_OPTION, null);
             boolean overwrite = getOption(OVERWRITE_OPTION, true);
             boolean orderedFields = getOption(ORDERED_FIELDS_OPTION, false);
             boolean useDecimal = getOption(BIGDECIMAL_TYPE_OPTION, false);
@@ -477,7 +479,7 @@ public class MessageCodeGenerator {
                 task.setMessagePackage("quickfix." + version.toLowerCase());
                 task.setOutputBaseDirectory(new File(args[2]));
                 task.setFieldPackage("quickfix.field");
-                task.setUtcTimestampPrecision(task.utcTimestampPrecision);
+                task.setUtcTimestampPrecision(utcTimestampPrecision);
                 task.setOverwrite(overwrite);
                 task.setOrderedFields(orderedFields);
                 task.setDecimalGenerated(useDecimal);
@@ -491,6 +493,11 @@ public class MessageCodeGenerator {
             codeGenerator.logError("error during code generation", e);
             System.exit(1);
         }
+    }
+
+    @SuppressWarnings("SameParameterValue")
+    private static String getOption(String key, String defaultValue) {
+        return System.getProperties().getProperty(key, defaultValue);
     }
 
     private static boolean getOption(String key, boolean defaultValue) {
--- a/quickfixj-codegenerator/src/main/java/org/quickfixj/codegenerator/MessageCodeGenerator.java
+++ b/quickfixj-codegenerator/src/main/java/org/quickfixj/codegenerator/MessageCodeGenerator.java
@@ -147,7 +147,7 @@ public class MessageCodeGenerator {
                         String utcTimestampPrecisionParameterName = "utcTimestampPrecision";
                         if (!UTC_TIMESTAMP_PRECISION_ALLOWED_VALUES.contains(utcTimestampPrecision)) {
                             throw new CodeGenerationException(new IllegalArgumentException(String.format(
-                                "Allowed values for parameter %s is %s. Supplied value: \"%s\".",
+                                "Allowed values for parameter %s are %s. Supplied value: \"%s\".",
                                 utcTimestampPrecisionParameterName,
                                 String.join(", ", UTC_TIMESTAMP_PRECISION_ALLOWED_VALUES),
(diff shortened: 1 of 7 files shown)


chrjohn:
Thanks for the PR. Since the changes seem to be backward-compatible I'm going to add this to the next minor release.

simenflatby:
Awesome @chrjohn.

quickfixj-codegenerator/src/main/java/org/quickfixj/codegenerator/MessageCodeGenerator.java
```diff
@@ -134,6 +142,19 @@ private void generateFieldClasses(Task task) throws ParserConfigurationException
                     Map<String, String> parameters = new HashMap<>();
                     parameters.put("fieldName", fieldName);
                     parameters.put("fieldPackage", task.getFieldPackage());
+                    String utcTimestampPrecision = task.getUtcTimestampPrecision();
+                    if (utcTimestampPrecision != null) {
+                        String utcTimestampPrecisionParameterName = "utcTimestampPrecision";
+                        if (!UTC_TIMESTAMP_PRECISION_ALLOWED_VALUES.contains(utcTimestampPrecision)) {
+                            throw new CodeGenerationException(new IllegalArgumentException(String.format(
+                                "Allowed values for parameter %s is %s. Supplied value: \"%s\".",
```
the-thing, line 150: Trivial: I think it should be "are".

quickfixj-codegenerator/src/main/java/org/quickfixj/codegenerator/MessageCodeGenerator.java
```diff
@@ -447,6 +477,7 @@ public static void main(String[] args) {
                 task.setMessagePackage("quickfix." + version.toLowerCase());
                 task.setOutputBaseDirectory(new File(args[2]));
                 task.setFieldPackage("quickfix.field");
+                task.setUtcTimestampPrecision(task.utcTimestampPrecision);
```
the-thing, line 480: Not sure how often this method is used, but this does not seem correct.

quickfixj-codegenerator/src/main/resources/org/quickfixj/codegenerator/Fields.xsl
```diff
@@ -87,6 +95,17 @@ public class <xsl:value-of select="@name"/> extends <xsl:call-template name="get
...
 	</xsl:if>
+
+    <xsl:choose>
+        <xsl:when test="$utcTimestampPrecision">
+            <xsl:if test="@type='UTCTIMESTAMP' or @type='UTCTIME' or @type='UTCTIMEONLY'">
+    @Override
+    protected UtcTimestampPrecision getDefaultUtcTimestampPrecision() {
+        return UtcTimestampPrecision.<xsl:value-of select="$utcTimestampPrecision"/>;
+    }
+            </xsl:if>
```
the-thing, line 106: Trivial. In the generated classes there is an empty line here. Not sure which xsl part generates this.

the-thing (changes requested):

quickfixj-codegenerator/src/main/java/org/quickfixj/codegenerator/MessageCodeGenerator.java
simenflatby: Fixed in ce5f595346c622671dc741ea09c6242ef5a7e81b.

quickfixj-codegenerator/src/main/resources/org/quickfixj/codegenerator/Fields.xsl
simenflatby: Fixed in 95c1373e1968d968e3002751c931d501c3d9c0f3.

quickfixj-codegenerator/src/main/java/org/quickfixj/codegenerator/MessageCodeGenerator.java
simenflatby: Hi @the-thing. Can you please elaborate? I did not understand what is wrong here.

simenflatby:
@chrjohn @the-thing Made an extra improvement: e1bb5fc45a007aac7f21d6ebdbf0b3254361228c. Please review.

quickfixj-codegenerator/src/main/java/org/quickfixj/codegenerator/MessageCodeGenerator.java
simenflatby: Ah, I see now that this looks stupid. Sets the value to it self... I'm not familiar with the usage og this main method, so maybe one of you could suggest a solution?

the-thing: I believe this methods is used when you want to run code generation locally from IDE etc.. I would follow the same pattern as e.g. overwrite/orderedFields/useDecimal. Create an option, get an option or default value, set it.

simenflatby: OK. Thank you. I will have a look at it.

simenflatby: Suggestion in e1b5bee92c8aeb4893e5f5a77936499dc78a7a71. Created the `getOption(String key, String defaultValue)` in case someone want to change the default value to something other than null later. I will check back to see if you have any feedback tomorrow.

the-thing (approved):
LGTM. I run it once via Maven plugin with NANO config and it worked. Also the `org.quickfixj.codegenerator.MessageCodeGenerator#main` seem to work with default values after the suggested changes.

chrjohn (approved):

Issue #300, opened 2020-06-09
Graceful disconnect functionality


philipwhiuk:
There's no straightforward way to tell a session "Logout and then disconnect"

If you call 'disconnect()' we just disconnect - we don't try to send a logout let alone wait for a response - we take a reason but we don't send it to the client. This is seen in #244 and #101 

Ideally there would be a method similar to:

Future<LogoutResponse> logoutAndDisconnect(int waitForResponseTimeout) {}

which:

* sent a logout
* optionally waited for a response (if timeout > 0)
* disconnected

This would be a high level call which abstracted away all the interaction waiting for the logout response. The future would provide a way of monitoring it if desired. 

Alternatives/work-arounds:

* User implementation of behaviour in app-code
* Just calling `disconnect` and hoping the other side behaves reasonably

chrjohn:
Also see #312

PR #312, opened 2020-09-20
Wait for Logout response or timeout before disconnect on session reset

### LINKED ISSUE #244: QFJ does not wait for Logout reply when EndTime is reached
Good afternoon,

We are using QuickFix 2.2.0 and FIX4.2 messages version.

We configured our session.cnf with StartTime and EndTime. Our problem is when time is equal to EndTime, QuickFix sends logout message but doesn't wait to response and generates a reset event: Disconnecting: Session reset.

Could you help us?

Thank you, regards.

simenflatby:
Fixes #244

_This is my first attempt to contribute to the core QuickFIX/J codebase. Please read the described changes bellow and review the changes thoroughly._

### Relevant changes

**quickfix.Session#reset()**
* When there is a need to logout: Instead of calling `quickfix.Session#generateLogout()` followed by `quickfix.Session#disconnect(String, boolean)`, set `quickfix.Session#isResetStatePending` flag and call `quickfix.Session#logout(String)` followed by `quickfix.Session#generateLogout(String)`. `quickfix.Session#logout(String)` will set `quickfix.Session#enabled` to `false` and `quickfix.Session#generateLogout(String)` will send a `Logout` message. The flag `quickfix.Session#isResetStatePending` is checked when `Logout` response is received or when timing out waiting for a `Logout` response.
* Only call `quickfix.Session#resetState()` if there is no need to logout.

**quickfix.Session#nextLogout(Message)**
* Reset state also if `quickfix.Session#isResetStatePending` flag is set.

**quickfix.Session#next()**
* Do not call `quickfix.Session#reset()` when outside of session time and `quickfix.Session#isResetStatePending` flag is set.
* When outside of session time and reset is not needed and reset state is pending, call `quickfix.Session#disconnect(String, boolean)`.
* Check for `Logout` timeout before `HeartBeat` timeout.

**quickfix.Session#disconnect(String, boolean)**
* Reset state also if `quickfix.Session#isResetStatePending` flag is set.

**quickfix.Session#resetState()**
* Clear `quickfix.Session#isResetStatePending` flag.

**quickfix.SessionTest#testLogonLogoutOnAcceptor()**
* Call `quickfix.SessionTest#logoutFrom(Session, int)`. This is needed because we need to receive a Logout response in order to trigger the session reset.

**quickfix.SessionTest#testStartOfInitiatorOutsideOfSessionTime()**
* Call `quickfix.SessionTest#logoutFrom(Session, int)`. This is needed because we need to receive a Logout response in order to trigger the session reset. If we would like to simulate a `Logout` timeout instead since we are outside of session time, we could replace the call to `quickfix.SessionTest#logoutFrom(Session, int)` with `systemTimeSource.increment(TimeUnit.HOURS.toMillis(state.getLogoutTimeout()) * 2); session.next();`.

### DIFF
--- a/quickfixj-core/src/main/java/org/quickfixj/jmx/mbean/session/SessionAdmin.java
+++ b/quickfixj-core/src/main/java/org/quickfixj/jmx/mbean/session/SessionAdmin.java
@@ -135,7 +135,7 @@ public class SessionAdmin extends NotificationBroadcasterSupport implements Sess
     /* (non-Javadoc)
      * @see quickfix.jmx.SessionMBean#reset()
      */
-    public void reset() throws IOException {
+    public void reset() {
         logInvocation("reset");
         session.reset();
     }
--- a/quickfixj-core/src/main/java/quickfix/Session.java
+++ b/quickfixj-core/src/main/java/quickfix/Session.java
@@ -428,6 +428,7 @@ public class Session implements Closeable {
 
     private final AtomicBoolean isResetting = new AtomicBoolean();
     private final AtomicBoolean isResettingState = new AtomicBoolean();
+    private final AtomicBoolean isResetStatePending = new AtomicBoolean();
 
     private final ListenerSupport stateListeners = new ListenerSupport(SessionStateListener.class);
     private final SessionStateListener stateListener = (SessionStateListener) stateListeners
@@ -867,12 +868,11 @@ public class Session implements Closeable {
     }
 
     /**
-     * Logs out and disconnects session (if logged on) and then resets session state.
+     * Logs out session (if logged on) and then resets session state.
      *
-     * @throws IOException IO error
      * @see SessionState#reset()
      */
-    public void reset() throws IOException {
+    public void reset() {
         if (!isResetting.compareAndSet(false, true)) {
             return;
         }
@@ -881,10 +881,13 @@ public class Session implements Closeable {
                 if (application instanceof ApplicationExtended) {
                     ((ApplicationExtended) application).onBeforeSessionReset(sessionID);
                 }
-                generateLogout();
-                disconnect("Session reset", false);
+                isResetStatePending.set(true);
+                logout("Session reset");
+                getLog().onEvent("Initiated logout request");
+                generateLogout(state.getLogoutReason());
+            } else {
+                resetState();
             }
-            resetState();
         } finally {
             isResetting.set(false);
         }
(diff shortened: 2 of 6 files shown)


chrjohn:
Thanks for the PR. I did not check the code yet but have one question: in `SessionTest` you are calling `next()` twice, does this mean one has to wait for two calls to `next()` at runtime also, i.e. max 2 seconds?

simenflatby:
> Thanks for the PR. I did not check the code yet but have one question: in `SessionTest` you are calling `next()` twice, does this mean one has to wait for two calls to `next()` at runtime also, i.e. max 2 seconds?

@chrjohn Looking at how `quickfix.Session#next()` i called now, I think you might be right. I will look at a solution tonight or one of the next days.

simenflatby:
@chrjohn I have now pushed an improved version. I squashed the new commit with the previous one so that you only have to review one. I have also edited the description of the pr. to reflect the new changes. Please review thoroughly when you can find the time.

philipwhiuk:
This looks okay to me - it doesn't fix #300 but we can probably build that on top...

chrjohn:
@simenflatby thanks for this and :+1: for the extensive test coverage.
@philipwhiuk thanks for your review.

Currently not sure if I want to put this to the next patch release (to be released during the next weeks I think) since the state flags (i.e. the already existing ones) are hard to follow already and there were some problems around them during the last years.

simenflatby:
@chrjohn My pleasure. I'm no longer with the company that I did the FIX integration that required this for, so I will not be able to test it in a real life project. If there is any questions, feel free to let me know and I will try to answer.

@philipwhiuk There has been some months since i looked at this code, so bear with me if I'm wrong: Would not calling `quickfix.Session#generateLogout()` solve #300? I would think that this will send a Logout message and that when the Logout message response is received in `quickfix.Session#nextLogout(Message)` it would trigger a disconnect?

chrjohn (approved):

chrjohn:
@simenflatby One question though: is there a specific reason why you used an `AtomicBoolean` for this? With the other `AtomicBooleans` that we use we do an atomic operation `compareAndSet` which you don't do with your `isResetStatePending` flag.

simenflatby:
@chrjohn I can not remember, but from looking at it now, it looks like a sloppy copy-paste of the other state flags. I should probably be changed. I'm not in front of a dev computer for a while. Will you change it please? Otherwise I can do it after the weekend. Cheers!

chrjohn:
Please leave it for the time being. Need to think about if it needs to go into the SessionState class...

chrjohn:
@simenflatby , I think this state flag needs to be moved to the `SessionState` class and protected by its lock. See the other flags for reference, e.g. `logonSent`, `logoutSent` etc.
Please let me know if you still have the time to change this, otherwise I'll do it.
Thanks in advance :+1:

simenflatby:
@chrjohn Unfortunately I have a lot on my schedule in January and February and I do not think I can find any time for OpenSource work. If it can wait until after that, I can definitely give it a shot. Just let me know!

chrjohn:
@simenflatby , no worries, I'll plan to do it then. Thanks for your work on this. :+1:

chrjohn (approved):

PR #316, opened 2020-09-23
ResetOnLogon/ResetOnLogout/ResetOnDisconnect fails when PersistMessages is set to 'N' 

### LINKED ISSUE #314: ResetOnLogon/ResetOnLogout/ResetOnDisconnect fails when PersistMessages is set to 'N'
**Describe the bug**
ResetOnLogon/ResetOnLogout/ResetOnDisconnect fails when PersistMessages is set to 'N'. It throws an IOException "Invalid object name `messages`"

at quickfix.JdbcStore.reset(JdbcStore.java:195)

**To Reproduce**
Set `ResetOnLogon/ResetOnLogout/ResetOnDisconnect` to `Y` and `PersistMessages` to `N`. When Logon/Logout/Disconnect event occurs, the session tries to reset state, which in turn tries to reset the message store. On JdbcStore, this results in trying to delete from `messages` table, before updating the `sessions` table. If `PersistMessages` is set to `N`, no `messages` table is created.

**Expected behavior**
If `PersistMessages` is set to `N`, no attempt should be made to delete from `messages` table.

**system information:**
 - OS: Linux
 - Java version JDK8
 - QFJ Version 2.2.0

**Additional context**
Either update https://www.quickfixj.org/usermanual/2.1.0/usage/configuration.html to clearly mention which settings can be used in conjunction with another. Or pull delete messages logic from lines 182-185 in `JdbcStore.java` (QFJ Version 2.2.0) into a separate method `deleteMessages` and call it from line 2661 of `Session.java` with a check if `persistMessages` is `true` around it.

dparui:
Fixes #314 

ResetOnLogon/ResetOnLogout/ResetOnDisconnect fails when PersistMessages is set to 'N'. It throws an IOException "Invalid object name messages"

More details at https://github.com/quickfix-j/quickfixj/issues/314

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/JdbcStore.java
+++ b/quickfixj-core/src/main/java/quickfix/JdbcStore.java
@@ -173,17 +173,28 @@ class JdbcStore implements MessageStore {
         setNextTargetMsgSeqNum(cache.getNextTargetMsgSeqNum());
     }
 
-    public void reset() throws IOException {
-        cache.reset();
+    public void clearMessages() throws IOException {
         Connection connection = null;
         PreparedStatement deleteMessages = null;
-        PreparedStatement updateTime = null;
         try {
             connection = dataSource.getConnection();
             deleteMessages = connection.prepareStatement(SQL_DELETE_MESSAGES);
             setSessionIdParameters(deleteMessages, 1);
             deleteMessages.execute();
+        } catch (SQLException e) {
+            throw new IOException(e.getMessage(), e);
+        } finally {
+            JdbcUtil.close(sessionID, deleteMessages);
+            JdbcUtil.close(sessionID, connection);
+        }
+    }
 
+    public void reset() throws IOException {
+        cache.reset();
+        Connection connection = null;
+        PreparedStatement updateTime = null;
+        try {
+            connection = dataSource.getConnection();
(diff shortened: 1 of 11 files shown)


chrjohn:
I thought about this. IMHO there are quite some changes to solve a problem in only one of the stores In my opinion it would be easier to query the `PersistMessages` setting in the `JdbcStore` and then either try to delete the `messages` table based on that setting.
What do you think?

dparui:
Restricted changes to `JdbcStore` as suggested. Please review.

quickfixj-core/src/main/java/quickfix/JdbcStore.java
```diff
@@ -194,7 +200,9 @@ public void reset() throws IOException {
         } catch (SQLException e) {
             throw new IOException(e.getMessage(), e);
         } finally {
-            JdbcUtil.close(sessionID, deleteMessages);
+            if(persistMessages) {
```
chrjohn, line 203: I think this is not necessary since `JdbcUtil` will check the statement for `null` anyway.
https://github.com/quickfix-j/quickfixj/blob/9df0fe1334b56449a448a985b80df77acae3813b/quickfixj-core/src/main/java/quickfix/JdbcUtil.java#L129-L132

dparui: Removed check.

PR #319, opened 2020-10-19
Created maven.yml for GitHub CI workflow

chrjohn:


chrjohn:
Did not check thoroughly yet but is this just a test error caused by comparison?

the-thing:
Yes, this is a comparison error, but it happens, because there is a small difference in behaviour in `javax.xml.transform.Transformer` between Java 8 & 11. There will be an additional one line code change to fix this for all versions.

Issue #317, opened 2020-10-12
Support for FIX Latest


david-gibbs-ig:
It  would be good for QFJ to support FIX Latest from FTC. If this is done by means of a dependency on the published FIX Latest Repository (Orchestration).

I would suggest that FIXLatest be added as an additional module to preserve backward compatibility of the existing FIX50sp2 module.

This would also provide a ready means for users to depend on their customized repositories instead of the base-line FIX Standard.

The FIX Latest repo includes some extension packs that break the QFJ build so code generated for messages in the new FIX50sp2Latest needs to be slightly different. The problem is where components and fields have the same name. As Java does not overload based on return type the method name for components must change. For backward compatibility this should not be applied to messages for prior versions of the standard.

I will provide an example build. See the branch here : https://github.com/david-gibbs-ig/quickfixj/tree/fix-orchestra
This above is a work in progress,

david-gibbs-ig:
To see how this might work, you can check this branch https://github.com/david-gibbs-ig/quickfixj/tree/fix-orchestra. I have added some readmes in an attempt to explain the changes. I can't get it to build on Travis as the "travis_wait" directive does not seem to have any effect and the build times out.

david-gibbs-ig:
This link provided for context : https://www.fixtrading.org/fix-online-specification-updated-to-fix-latest/

david-gibbs-ig:
Hi the build for the branch https://github.com/david-gibbs-ig/quickfixj/tree/fix-orchestra works for Java 8, it just will never work in Travis because it takes 1.5 hours.  Are you in the process of migrating to git actions ?

JThoennes:
@chrjohn Do you have any plans to release this? Cheers, Jörg

chrjohn:
Not in the next version, will be 3.0.0 most probably.

JThoennes:
@chrjohn 
> Not in the next version, will be 3.0.0 most probably.
Do you have any concrete plans for version 3? What it will contain and when you expect it to arrive?

chrjohn:
No concrete plans at the moment. Currently trying to get 2.2.1 (as I think about it, it should probably be called 2.3.0) out of the door.

What 3.0.0 will contain: up to now this: https://github.com/quickfix-j/quickfixj/milestone/10
It is loosely planned that 3.0.0 will also contain support for FIX Orchestra as far as validation is concerned, i.e. directly validate against an Orchestra file.

chrjohn:
[...]

PR #325, opened 2020-10-21
Fix orchestra

david-gibbs-ig:
Here is a pull request for Support for FIX Latest #317. It is for discussion. 

It generates the FIX Dictionary from the FIX Latest orchestration - and applies some transformations to work around the small number of incompatibilities with QFJ. Some of these will in due course be corrected by FTC.

Please see the readmes.

I think that supporting FIX Latest may require incrementing the major release number of QFJ because of the way that QFJ shares the generation of classes for Fields between FIX Versions.

FIX Latest uses more terse descriptions.
example :
newSingle.setField(new HandlInst(HandlInst.MANUAL_ORDER_BEST_EXECUTION)); becomes
newSingle.setField(new HandlInst(HandlInst.MANUAL_ORDER));

These are very easy code changes to make but they are an incompatibility.
Please note that Memory requirements and build time for FIX Latest javadoc are both significant.


chrjohn:
Hi @david-gibbs-ig , thanks for the PR. This will definitely be useful.  :+1: 
Will take some time to review, though. Just a question: was there a specific reason to deviate from the `quickfixj-` naming scheme for the module? No big thing, just asking. 
Thanks again.

david-gibbs-ig:
Hi Christophe, it might be better fit if the module was named "quickfixj-from-orchestra", I just did not think of it.  I'm sure this needs discussion and many refinements. It takes a long time to build which I think is a real concern, I have trouble with the tests in my local build environment which makes it is very hard to verify the build. It looks like Travis still won't build it, maybe there can be better luck with the github build tools.

chrjohn:
Will there be a follow up PR to this? Anything I can help you with?

david-gibbs-ig:
Thanks I have been working on it. I may have figured out what was making the build fail tonight. I will certainly ask for help and it would be good to discuss the potential changes with you if possible.

david-gibbs-ig:
Hi figured out how to get the build further but then hit this limit at 50 minutes. "The job exceeded the maximum time limit for jobs, and has been terminated." I may need a little help getting the builds running in github actions instead.

david-gibbs-ig:
> Will there be a follow up PR to this? Anything I can help you with?

Thanks, here is new PR that is pretty clean, see also comments on the issue and PR. Maybe we could have a call to discuss. PS have you considered using gitter, https://gitter.im/home . I only just found out about it.

https://github.com/quickfix-j/quickfixj/pull/333

thanks

PR #331, opened 2020-10-24
#330 Do not log messageString - it likely contains cleartext password

### LINKED ISSUE #330: Do not log cleartext password on logon failure
**Describe the bug**
Full FIX message with cleartext password is logged on logon failure.

**Expected behavior**
Either do not log the FIX message, or remove cleartext password from the message.

fillreject:
Closes #330

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/Session.java
+++ b/quickfixj-core/src/main/java/quickfix/Session.java
@@ -364,6 +364,12 @@ public class Session implements Closeable {
     public static final String SETTING_ALLOWED_REMOTE_ADDRESSES = "AllowedRemoteAddresses";
 
     /**
+     * Log the entire message when the corresponding session can not be found.
+     * Valid values are "Y" or "N". Default is "Y".
+     */
+    public static final String SETTING_LOG_MESSAGE_WHEN_SESSION_NOT_FOUND = "LogMessageWhenSessionNotFound";
+
+    /**
      * Setting to limit the size of a resend request in case of missing messages.
      * This is useful when the remote FIX engine does not allow to ask for more than n message for a ResendRequest
      */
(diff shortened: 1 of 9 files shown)


the-thing (commented):
I don't think this is a correct fix for the problem. This path is taken when session cannot be found for any message received, regardless of message type and if it contains password field. We might not want to disable logging for these messages.

fillreject:
@the-thing thanks for the prompt response! I can submit a PR to either remove password field from messageString or to make it possible to toggle message logging on/off. What would be the preferred option?

chrjohn:
Can't do a complete source code analysis at the moment but since there is no associated session I am asking myself where to store that configuration to toggle the behaviour on/off.

Moreover, what about other fields with sensitive data like 925/NewPassword?

I could only imagine that we log only the SessionID of the unknown session and not the rest of the message.

Could you maybe elaborate why this is a problem for unknown sessions? For known sessions you suppress it in your log implementation?

fillreject:
> Moreover, what about other fields with sensitive data like 925/NewPassword?

That is why my initial attempt was to remove messageString from the log output entirely - removing all fields, that may contain cleartext secrets, is relatively unreliable.

> Could you maybe elaborate why this is a problem for unknown sessions?

The problem is that this will log some messages for known sessions, that are sent by misconfigured clients. For example, this code will log error and include messageString for a "correct" logon message with a version that is not expected by the server (FIX.4.3 instead of FIX.4.4).

Another possible solution might be to disable detailed logging for messages with a FIX version that is not configured on the server.

chrjohn:
Just for information, the complete FIX message will also be logged by the `FIXMessageDecoder` if you have enabled DEBUG logging.

Apart from that: maybe it would be sufficient to log the SessionID or header of the wrong FIX message by default. I am open to suggestions. Personally I only used the SessionID in cases where this error turned up. That is usually all you need to look at because the validation is about the SessionID.

lyncmi07:
My new PR https://github.com/quickfix-j/quickfixj/pull/338
Has the changes suggested for this issue

chrjohn:
Closing, superseded by #338

PR #333, opened 2020-10-29
Fix orchestra

david-gibbs-ig:
Here is a pull request building QFJ from the FIX Latest orchestration.  re :  https://github.com/quickfix-j/quickfixj/issues/317. The build succeeds for Java 8 in git actions but Travis cannot build it as the complete build now takes about 1.5 hours and Travis has a hard limit of around 50 minutes. Perhaps some optimization is possible. Compilation and Javadoc now require more memory , presumably because of the size of FIX Latest.


.github/workflows/maven.yml
```diff
@@ -24,5 +24,5 @@ jobs:
           java-version: ${{ matrix.java }}
       - name: Test with Maven
         env:
-          MAVEN_OPTS: "-Xms3g -Xmx3g"
+          MAVEN_OPTS: "-Xms3g -Xmx9g"
```
chrjohn, line 27: 9GB??? woah

david-gibbs-ig: Yes it takes a lot of memory for compiler and for javadoc. It's a beast

david-gibbs-ig:
Closing this pull request due to issue FIXTradingCommunity/fix-orchestra#104 .

david-gibbs-ig:
@chrjohn Hi Christophe, have you got Java 11 plus build working ? If so, how ?

chrjohn:
Hi @david-gibbs-ig , I guess you are referring to the following Exception when building:

```
javax/xml/bind/JAXBContext
    at io.fixprotocol.orchestra.quickfix.DataDictionaryGenerator.unmarshal (DataDictionaryGenerator.java:208)
    at io.fixprotocol.orchestra.quickfix.DataDictionaryGenerator.generate (DataDictionaryGenerator.java:103)
    at io.fixprotocol.orchestra.quickfix.DataDictionaryGenerator.main (DataDictionaryGenerator.java:90)
    at org.codehaus.mojo.exec.ExecJavaMojo$1.run (ExecJavaMojo.java:254)
    at java.lang.Thread.run (Thread.java:834)
Caused by: java.lang.ClassNotFoundException: javax.xml.bind.JAXBContext
    at java.net.URLClassLoader.findClass (URLClassLoader.java:471)
    at java.lang.ClassLoader.loadClass (ClassLoader.java:589)
    at java.lang.ClassLoader.loadClass (ClassLoader.java:522)
    at io.fixprotocol.orchestra.quickfix.DataDictionaryGenerator.unmarshal (DataDictionaryGenerator.java:208)
    at io.fixprotocol.orchestra.quickfix.DataDictionaryGenerator.generate (DataDictionaryGenerator.java:103)
    at io.fixprotocol.orchestra.quickfix.DataDictionaryGenerator.main (DataDictionaryGenerator.java:90)
    at org.codehaus.mojo.exec.ExecJavaMojo$1.run (ExecJavaMojo.java:254)
    at java.lang.Thread.run (Thread.java:834)

```

This can be corrected by using the following dependencies as per https://stackoverflow.com/a/48204154/4962355

```
<!-- API -->
<dependency>
    <groupId>jakarta.xml.bind</groupId>
    <artifactId>jakarta.xml.bind-api</artifactId>
    <version>2.3.3</version>
</dependency>

<!-- Runtime -->
<dependency>
    <groupId>com.sun.xml.bind</groupId>
    <artifactId>jaxb-impl</artifactId>
    <version>2.3.3</version>
    <scope>runtime</scope>
</dependency>
```

chrjohn:
Hmm, I am asking myself why the message generation runs so slow for FIX Latest. Comparing to FIX5.0SP2 there are

- three times as much `components`
- about 50% more `message` types
- about 10 times as much `enum? definitions (2700 compared to 25.000 for FIX Latest)

FIX5.0SP2 generation takes only about a minute on my machine. FIX Latest takes about an hour...

I am not an XSLT expert, but it seems there are some inefficiencies in the transformation. At least that is my gut feeling, which might of course be wrong.

david-gibbs-ig:
Thanks, I think that makes sense. When I was losing sleep about it the other night I imagined that it might be that this test <xsl:when test="//fix/@fixLatest ='true'"> is done for each component.  

I am not experienced with XSLT but I imagine that setting an <xsl:variable> one time and testing the variable rather than the expression would be better.  In fact this is the only change to the XSLT. Also I was wondering if "//fix/@fixLatest" vs "//fix/@fixLatest" would make a difference. 

If you are still supportive in general of adding FIX Latest then I will spend a little more time on it. I have raised issue FIXTradingCommunity/fix-orchestra#104 with the FIX Orchestra project which affects this work. If you would like to join the FIX Orchestra working group then you will see some of my other comments and questions and we can discuss the issues such as what the Major and Minor FIX versions should look like for FIXLatest. There are definitely some wrinkles to iron out. I think that having QFJ  (and QuickFIX ) in the pipeline for FIX Orchestra is really important to prove that Orchestra is viable.

I don't know how to use a XSLT variable but I have found a clue that I will  put here so I can find it later : https://stackoverflow.com/questions/4204996/xslt-can-i-declare-a-variable-globally-and-later-assign-a-value-to-it

<xsl:choose>
	    
	  	**<xsl:when test="//fix/@fixLatest ='true'">**
	public <xsl:value-of select="$type"/> get<xsl:value-of select="@name"/>Component() throws FieldNotFound {
		return get(new <xsl:value-of select="$type"/>());
	}
		</xsl:when>
		<xsl:otherwise>
	public <xsl:value-of select="$type"/> get<xsl:value-of select="@name"/>() throws FieldNotFound {
		return get(new <xsl:value-of select="$type"/>());
	}	
		</xsl:otherwise>
	</xsl:choose>

chrjohn:
Hi @david-gibbs-ig , I think the portion that you mentioned is not the problem because I also removed it to test if it made a difference. It didn't. :-/

david-gibbs-ig:
Here is a question I asked in Orchestra working group. Please tell me what you think makes sense.

"Hi, I know it must have been discussed, but what is the intention for identifying FIX Latest in terms of ApplVerID, DefaultApplVerID etc ?

"repository-quickfix" outputs <fix major="5" minor="0"> and FIXimate displays "FIX.5.0SP2_EP258".

These seems to suggest that FIX Latest should be treated in FIXT1.1 transport as being FIX 5.0. I think that is confusing.

The maintainer of QuickFIX/J sees FIX 5.0SP2 as an immutable version for the purpose of maintaining backward compatibility.

Could it be helpful to specify "Latest" as a new major version in FIXT messages and headers ?

"ApplExtID" and "DefaultApplExtID" would of course serve to identify the EP and "DefaultCstmApplVerID" the custom application version.

P.S. In order to build QuickFIX/J it is neccesary for code generation to distinguish if the dictionary is for FIXLatest. This is because of name collision in components and fields that needs to be worked-around. At present I add an attribute to the generated dictionary but it would be ideal to standardise a solution for this. Whatever is decided about the above major/minor version, adding an attribute for the EP to the dictionary would be sufficent to distinguish FIX Latest as the legacy dictionaries do not contain this.

(see https://github.com/david-gibbs-ig/quickfixj/blob/fix-orchestra/quickfixj-codegenerator/src/main/resources/org/quickfixj/codegenerator/MessageSubclass.xsl)"

chrjohn:
Thanks, I will join the Orchestra working group. Was planning to do this earlier anyway. I am open to suggestions about the FIX version. On one hand users might expect FIX5.0SP2 dictionary to contain just that, others are probably OK with that dictionary containing the latest changes...

david-gibbs-ig:
> Hi @david-gibbs-ig , I think the portion that you mentioned is not the problem because I also removed it to test if it made a difference. It didn't. :-/

Do you mean you removed the change to the XSL to test ? If so then I assume that potential optimizations would be needed in the original XSL :-(

david-gibbs-ig:
> Thanks, I will join the Orchestra working group. Was planning to do this earlier anyway. I am open to suggestions about the FIX version. On one hand users might expect FIX5.0SP2 dictionary to contain just that, others are probably OK with that dictionary containing the latest EP changes, too...

I kind of liked your approach of it being an (the last) immutable version , especially if it will be hard to reduce build times for latest. Latest could be commented out by forkers that do not want to build latest. The approach I wanted to be able to take is to simply point the forked build at the customised (cut down) orchestration. 
This works - https://github.com/IG-Group/ig-orchestrations/tree/master/ig-us-rfed/orchestration

chrjohn:
Hi @david-gibbs-ig ,

> Do you mean you removed the change to the XSL to test ? If so then I assume that potential optimizations would be needed in the original XSL :-(

Yes, I also meant the original XSLT. But as I am not an expert I don't know where to start.

chrjohn:
Re XSLT: I saw that XMLSpy has an XSLT optimizer. You can try out the software for 30 days. However it seems only to be available for Windows so I cannot use it.

david-gibbs-ig:
> Hi @david-gibbs-ig , I guess you are referring to the following Exception when building:
> 
> ```
> javax/xml/bind/JAXBContext
>     at io.fixprotocol.orchestra.quickfix.DataDictionaryGenerator.unmarshal (DataDictionaryGenerator.java:208)
>     at io.fixprotocol.orchestra.quickfix.DataDictionaryGenerator.generate (DataDictionaryGenerator.java:103)
>...
> Edit: should also work with Java 8.

Thanks I realise now that I saw the issue on my branch and not on master branch because of the use of exec-mvn-plugin in the branch. I updated the config for the plugin and the build now works on Java 8 and 11.

david-gibbs-ig:
@chrjohn  I now think that the way to implement FIX Latest will be to use the code generator from the orchestra project - it does run much more quickly. I will experiment with this for FIX Latest  to see how its compatibility looks. If we were to use it, it could be only for FIX Latest to help assure greater stability for legacy versions.

chrjohn:
@david-gibbs-ig , yes, I saw the code generator from Orchestra. It directly converts the Orchestra files to Java classes.

PR #335, opened 2020-10-31
Updated berkeleydb / sleepycat dependency

### LINKED ISSUE #334: Relying on ancient version of Berkeley DB JE
Hi Quickfix folk,

I just wanted to note that the https://github.com/quickfix-j/quickfixj/blob/master/quickfixj-core/pom.xml file is depending on a 2006 version of Berkeley DB JE, with strong copyleft licensing. 

That project was relicensed to Apache-2.0 a few years ago, and presumably has bugfixes/security fixes that would be useful for Quickfix users.

Might I recommend that you update the dependency?

chrjohn:
Fixes #334

### DIFF
--- a/quickfixj-core/pom.xml
+++ b/quickfixj-core/pom.xml
@@ -112,9 +112,9 @@
             <optional>true</optional>
         </dependency>
         <dependency>
-            <groupId>berkeleydb</groupId>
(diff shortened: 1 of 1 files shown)


chrjohn:
Reopen to test if Travis is disabled now.

PR #338, opened 2020-11-18
Log sessionId for rejected message - message may contain password

### LINKED ISSUE #330: Do not log cleartext password on logon failure

lyncmi07:
Closes #330 
Following up on comments on previous PR for this issue: https://github.com/quickfix-j/quickfixj/pull/331

I'm logging just the session id now as suggested. This should be enough information from our perspective too.
Less worried about the issue raised regarding it appearing in DEBUG enabled logging given we don't run that in prod.


chrjohn:
Closing due to #352

PR #340, opened 2020-12-08
#330 Logging settings for messageString

fillreject:
New setting to disable messageString logging.

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/mina/LoggingSettings.java
+++ b/quickfixj-core/src/main/java/quickfix/mina/LoggingSettings.java
@@ -1,0 +1,36 @@
+/*******************************************************************************
+ * Copyright (c) quickfixengine.org  All rights reserved.
+ *
+ * This file is part of the QuickFIX FIX Engine
+ *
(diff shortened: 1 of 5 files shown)


chrjohn (changes requested):
Actually this setting should originate from the general `SessionSettings` object that already is available and needs to be handed to the `AcceptorIoHandler`.

quickfixj-core/src/main/java/quickfix/mina/LoggingSettings.java
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * This class holds the QFJ settings information related to logging while session is unavailable.
+ */
+public class LoggingSettings {
+
+    private final boolean logGarbledMessage;
+
+    public LoggingSettings(final boolean logGarbledMessage) {
+        this.logGarbledMessage = logGarbledMessage;
```
chrjohn, line 30: Just as a side note: a garbled message is something completely different.
https://www.fixtrading.org/standards/fix-session-layer-online/#garbled-message-processing

PR #347, opened 2020-12-30
Add methods to `FieldMap` API to make it usable in for-each loop

AndreyNudko:
FieldMap already provides iterator() to inspect all fields, however does not implement Iterable - and so can't be used in for-each loop. This PR adds Iterable interface, as well as groupKeys() which exposes group keys as Iterable.
The motivation for the change is that in some cases it is useful to traverse message as a tree of key-value pairs - e.g. when visualising message structure in UI, or transforming it to another intermediate representation. Being able to iterate in a more idiomatic Collection-like way improves readability of such code.

I was thinking about marking iterator() and groupKeyIterator() as deprecated - I don't see any advantages of those methods over Iterable. In the end decided to not do this for the first cut, but they can be eventually retired.

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/DataDictionary.java
+++ b/quickfixj-core/src/main/java/quickfix/DataDictionary.java
@@ -43,7 +43,6 @@ import java.io.InputStream;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


chrjohn:
Hi @AndreyNudko , thanks for the PR. Please mark methods as deprecated as you see fit.

AndreyNudko:
@chrjohn I've realised iterator() needs to stay as it is part of `Iterable`;  it may be used implicitly in for-each, but there's nothing wrong with the method. I've deprecated `groupKeyIterator()` in favour of `groupKeys()` though.

PR #368, opened 2021-02-09
Add SessionStateListener.onConnectException()

### LINKED ISSUE #342: SessionStateListener not notified about initial connection failures
**Describe the bug**
If a Session wholly fails to initialise (never gets as far as a TCP connection) nothing is notified to its SessionStateListener

**To Reproduce**
* Add a SessionStateListener to to a Session configured such that it will never connect
* Observe the event log has something like 'java.net.ConnectException during connection to /localhost:80: java.net.ConnectException: Connection timed out (Next retry in 30000 milliseconds)'

**Expected behavior**
Something to be notified to your listener (`onDisconnect` ?)

**system information:**
 - QFJ 2.1.1

**Additional context**
The exception shown above is sent to the qfj `Log` framework then swallowed by `quickfix.mina.initiator.IoSessionInitiator.ConnectTask#handleConnectException`

Listener `onDisconnect` is the only viable existing method that would be suitable although I don't know if it's necessarily the right choice. That rather depends on how it used for other categories of first-time initialisation error (i.e. an SSL error or FW issue) that manifest _after_ TCP is connected.

chrjohn:
Fixes #342 

Added `SessionStateListener.onConnectException()` which gets called from `IoSessionInitiator.handleConnectException()`

Added `Session.getStateListener()`.

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/mina/initiator/IoSessionInitiator.java
+++ b/quickfixj-core/src/main/java/quickfix/mina/initiator/IoSessionInitiator.java
@@ -261,8 +261,10 @@ public class IoSessionInitiator {
             final String nextRetryMsg = " (Next retry in " + computeNextRetryConnectDelay() + " milliseconds)";
             if (e instanceof IOException) {
                 fixSession.getLog().onErrorEvent(e.getClass().getName() + " during connection to " + socketAddress + ": " + e + nextRetryMsg);
+                fixSession.getStateListener().onConnectException((IOException) e);
             } else {
                 LogUtil.logThrowable(fixSession.getLog(), "Exception during connection to " + socketAddress + nextRetryMsg, e);
+                fixSession.getStateListener().onConnectException(new Exception(e));
             }
             connectFuture = null;
         }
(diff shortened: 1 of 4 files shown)


quickfixj-core/src/main/java/quickfix/mina/initiator/IoSessionInitiator.java
```diff
@@ -261,8 +261,10 @@ private void handleConnectException(Throwable e) {
             final String nextRetryMsg = " (Next retry in " + computeNextRetryConnectDelay() + " milliseconds)";
             if (e instanceof IOException) {
                 fixSession.getLog().onErrorEvent(e.getClass().getName() + " during connection to " + socketAddress + ": " + e + nextRetryMsg);
+                fixSession.getStateListener().onConnectException((IOException) e);
```
drekbour, line 264: Little smelly to expose and call another classes listeners (Law of Demeter) but not egregiously so.

chrjohn: I agree. What's the alternative? Pass the listener through the Initiator implementations or create an all-new Initiator exception listener?

drekbour: The line above already sets precedent for calling through `Session.getLog()` and the usecase appears (since I reported it after many years) "edge" so I wouldn't over-engineer it more than I see here. I can't dig in enough to volunteer an alternative solution.

PR #370, opened 2021-02-10
Add SessionID to SessionStateListener methods

esanchezros:


chrjohn (commented):
In my opinion you can remove the `@Deprecated` methods. Since adding a new parameter to the methods is a breaking change anyway (and only the new methods are called in the code), we can simply change the method signature without deprecating the old methods. What do you think?
Edit: so if we left the `@Deprecated` methods in the code then everyone who implemented these would be curious why the listener methods are not called anymore.

esanchezros:
My intention was to allow users to notice the `@Deprecated` methods and let them change their implementations at their discretion, before removing them. But on the other, you're right, their listeners won't get called at all. I will update the PR accordingly.

chrjohn:
Note to self: once #368 is merged, there will be another method signature to change.

PR #374, opened 2021-03-02
Update configuration.html

lschin:
Change/correct logging setting from ScreenLogEvents to ScreenLogShowEvents


chrjohn:
Thanks for the PR! Good catch :+1:

PR #380, opened 2021-04-15
[QFJ-968] Allow provision of custom Message Queue implementation

korzha:
We have ran into similar issue described here https://www.quickfixj.org/jira/browse/QFJ-968

We have a huge fix message flow and if we restart our service after a while (after fixing some errors), 
our resend request becomes really big and takes a few minutes to fulfill.
Live messages go into the SessionState's messageQueue and fill our whole memory (tens of gigs) in just a few minutes very quickly and the process crashes OOM.

My proposal is to allow to provide a custom implementation of the message queue like you do with MessageStore for example. This will let us store have our implementation of the queue to flush queue on disk.
I've added new constructor to Session to keep old api in place.


wangvic:
This is a very useful change. I want this featue/improvement to be included in the next release version.

chrjohn (commented):
@korzha thanks for the PR. Could you please take a look? About 750 tests are failing.

chrjohn:
@wangvic apart from the things that still need to be looked at this PR could probably be part of 3.0.0 or the minor release after the next.

chrjohn (changes requested):
Unit tests need fixing. Looks like a general problem around Logon process.

korzha:
@chrjohn Yes, sorry for the noise.
I think I've fixed them now, let's wait for ci results.

And thanks for suggestion about using watermark queue. Unfortunately we need to keep all the messages we receive without blocking upstream.

quickfixj-core/src/main/java/quickfix/SessionState.java
```diff
@@ -260,6 +259,10 @@ public MessageStore getMessageStore() {
         return messageStore;
     }
 
+    public MessageQueue getMessageQueue() {
+        return messageQueue;
```
chrjohn, line 263: Is there a specific reason why you have a getter for the `messageQueue` and then do `state.getMessageQueue().xyz` all the time?

korzha: I wanted to make usage the same as you do with message store:

Sessions does `state.getMessageStore().xyz` calls

Would you prefer to keep those methods?

chrjohn: I see, let me think about it...

chrjohn:
Thanks so far, I will review in due course. Looking good at first sight.

chrjohn:
Closing and reopening to trigger new build jobs.

chrjohn (approved):
Thanks for the PR. Finally gotten around to merge it. Had to do two minor changes in the tests.

PR #384, opened 2021-04-20
Always re-resolve `SocketAddress` on reconnect.

jimbogithub:
QFJ-266 and QFJ-822 previously made changes around the clearing of cached DNS resolution.  These resolved issues relating to a cached non-resolution, and enforcing re-resolution after a failure to connect.  However there are other cases that this does not cover such as a counter-party changing their DNS entries but leaving something listening at the old location.  In this case you can still connect successfully but not get a response to your Logon message.  As this is not a connection error, the QFJ-822 enhancement is not triggered.

All cases can be covered by following the QFJ-266 solution but ignoring whether the address was previously resolved or unresolved.  Simply clear the cached DNS info on every connect.

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/mina/initiator/IoSessionInitiator.java
+++ b/quickfixj-core/src/main/java/quickfix/mina/initiator/IoSessionInitiator.java
@@ -254,7 +254,6 @@ public class IoSessionInitiator {
         private void handleConnectException(Throwable e) {
             ++connectionFailureCount;
             SocketAddress socketAddress = socketAddresses[getCurrentSocketAddressIndex()];
-            unresolveCurrentSocketAddress(socketAddress);
             while (e.getCause() != null) {
                 e = e.getCause();
             }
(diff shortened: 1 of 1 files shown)


chrjohn:
@jimbogithub might this have any performance implications? E.g. could there be cases where recreation of the `SocketAddress` might be undesired because it takes some time to resolve the address? But I assume unless there is a network timeout or DNS problems the resolution should be fast.

Could you also please check the whitespace around your change? It seems it is still a bit off in lines 277-279.

jimbogithub:
> @jimbogithub thanks for the PR!
> Might this have any performance implications? E.g. could there be cases where recreation of the `SocketAddress` might be undesired because it takes some time to resolve the address? But I assume unless there is a network timeout or DNS problems the resolution should be fast.

I'm not an expert on this but I assume DNS resolution is normally fairly fast and is cached in many places and invalidated automatically for the proper distributed DNS participants.  I guess it's possible that a system could get stuck with no DNS capability (it does happen once in a while with my ISP/Modem) and wish that the cached value was still being used.  Maybe this needs a vote or an expert to consider.

A lighter weight alternative would be to set a flag similar to the way that `SessionConnector.QFJ_RESET_IO_CONNECTOR` is used.  My case is `quickfix.Session` `disconnect("Timed out waiting for logon response", true);`.  Setting a `SessionConnector.QFJ_INVALIDATE_DNS` flag there and checking for it in `quickfix.mina.initiator.IoSessionInitiator.getNextSocketAddress` could also work.  That would only cover this one extra case though - are there others?

> Could you also please check the whitespace around your change? It seems it is still a bit off in lines 277-279.

Whitespace done.

chrjohn:
@jimbogithub thank you.
How important is this change for you? Would it be OK if it is part of the release after next?

jimbogithub:
> @jimbogithub thank you.
> How important is this change for you? Would it be OK if it is part of the release after next? This would allow for more time to analyse/clarify.

That could depend on how long till that next release.

Would it help now if we had an opt-in or opt-out flag?  i.e. restore the original code but have AbstractSocketInitiator check for a new Initiator.SETTING_xxx and pass a flag to the IoSessionInitiator constructor which tunes it to either legacy or non-cached behaviour.  That way we can get it into the wild now but allow users to choose / tune.

chrjohn:
After thinking more about it I believe this change is OK. If there are other opinions there is some time now (months) until the next major release. ;) If it is really required faster then we could cherry-pick this into the next minor/patch release.

@the-thing and/or @philipwhiuk any opinion on this? I think it is OK, just did not want to put it into the release at short notice.

the-thing:
I'm not familiar with the DNS case described, but the fix looks good enough.

philipwhiuk:
This feels kind of like a work-around for a proper failover/retry strategy. 
It would be nice if it were a little more fine-grained than just "flush every time"

I would caution that this sort of issue is really an endemic issue of relying on DNS to handle DR and it's why it's not a good approach. DNS is designed around caching, everywhere.

It would better to just configure both IPs/entries for both and then have an appropriate failover strategy.

Currently we just have a very basic "try them in order, immediately" strategy.

What we really want is "Try the first one X-times over Y seconds, then failover to the second one and stay connected to that until failing X-times over Y seconds"

I might have the time to dedicate to this sort of work, but not imminently.

warnesey333:
Hey, out of interest did this get released? Seeing the same issue when connecting to an AWS acceptor.

chrjohn:
@warnesey333 it has been merged to master branch but there was no release done since then. Currently cannot tell when this will be done.

PR #395, opened 2021-05-17
allow sending unchanged message in `Session#send`

### LINKED ISSUE #394: Be able to send a raw message without removing any field
I have a use case where one of our services need to propagate `PossDupFlag` and `OrigSendingTime` to a third party. Using `Session.send` makes this impossible as these fields are removed from the message header.

It will be great to have an overloaded method that will allow sending the message as-is without removing any field. Something like:

    public boolean send(Message message, boolean sendRaw);

Alternatively, `sendRaw` method can be made public in `Session`. At the moment I have to use reflection to invoke `sendRaw` which is pretty hacky:

    Method method = targetSession.getClass().getDeclaredMethod("sendRaw", Message.class, int.class);
    method.setAccessible(true);
    if (!(boolean) method.invoke(targetSession, message, 0)) {  

Or fork the project and maintain a patch version in our dependency management server.

esanchezros:
Fixes #394

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/Session.java
+++ b/quickfixj-core/src/main/java/quickfix/Session.java
@@ -2685,8 +2685,29 @@ public class Session implements Closeable {
      * @return a status flag indicating whether the write to the network layer was successful.
      */
     public boolean send(Message message) {
-        message.getHeader().removeField(PossDupFlag.FIELD);
-        message.getHeader().removeField(OrigSendingTime.FIELD);
+        return send(message, false);
+    }
+
+    /**
+     * Send a message to a counterparty. Sequence numbers and information about the sender
+     * and target identification will be added automatically (or overwritten if that
+     * information already is present).
+     *
+     * The returned status flag is included for
+     * compatibility with the JNI API but it's usefulness is questionable.
+     * In QuickFIX/J, the message is transmitted using asynchronous network I/O so the boolean
+     * only indicates the message was successfully queued for transmission. An error could still
+     * occur before the message data is actually sent.
+     *
+     * @param message       the message to send
+     * @param allowPosDup   whether to allow PossDupFlag and OrigSendingTime in the message
+     * @return a status flag indicating whether the write to the network layer was successful.
+     */
+    public boolean send(Message message, boolean allowPosDup) {
+        if (!allowPosDup) {
+            message.getHeader().removeField(PossDupFlag.FIELD);
+            message.getHeader().removeField(OrigSendingTime.FIELD);
+        }
         return sendRaw(message, 0);
     }
 
--- a/quickfixj-core/src/main/java/quickfix/Session.java
+++ b/quickfixj-core/src/main/java/quickfix/Session.java
@@ -374,7 +374,7 @@ public class Session implements Closeable {
     public static final String SETTING_VALIDATE_CHECKSUM = "ValidateChecksum";
 
     /**
-     * Option so that the session doesn't remove PossDupFlag (43) information when sending.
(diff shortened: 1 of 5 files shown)


esanchezros:
This PR is based on https://github.com/quickfix-j/quickfixj/pull/236, which seems to be abandoned, and related to https://github.com/quickfix-j/quickfixj/issues/394

chrjohn:
Hi @esanchezros , thanks for the PR.
The original PR also has a setting that permanently turns this feature on or off, although AFAICT it is not assigned in the `Session` constructor (maybe was forgotten). I think it would be beneficial to also have this setting. This then also needs to be put into `configuration.html'. If you have questions about this, just ask. :)

Cheers,
Chris.

quickfixj-core/src/main/java/quickfix/Session.java
```diff
@@ -2685,8 +2685,29 @@ private void resetState() {
...
+        return send(message, false);
+    }
+
+    /**
+     * Send a message to a counterparty. Sequence numbers and information about the sender
+     * and target identification will be added automatically (or overwritten if that
+     * information already is present).
+     *
+     * The returned status flag is included for
+     * compatibility with the JNI API but it's usefulness is questionable.
```
chrjohn, line 2697:
```suggestion
     * compatibility with the JNI API but its usefulness is questionable.
```

esanchezros:
Hi @chrjohn, I've updated the PR including your the configuration setting you mentioned. Let me know if it needs more work
Thanks
Ed

quickfixj-core/src/main/java/quickfix/Session.java
```diff
@@ -373,6 +373,11 @@
 
     public static final String SETTING_VALIDATE_CHECKSUM = "ValidateChecksum";
 
+    /**
+     * Option so that the session doesn't remove PossDupFlag (43) information when sending.
```
chrjohn, line 377:
```suggestion
     * Option so that the session does not remove PossDupFlag (43) and OrigSendingTime (122) information when sending.
```

quickfixj-core/src/main/java/quickfix/Session.java
```diff
@@ -423,6 +428,7 @@
     private boolean enableNextExpectedMsgSeqNum = false;
     private boolean enableLastMsgSeqNumProcessed = false;
     private boolean validateChecksum = true;
+    private boolean allowPosDup;
```
chrjohn, line 431:
```suggestion
    private boolean allowPosDup = false;
```
I know `false` is the default but let's make it explicit as it is with the other flags.

quickfixj-core/src/main/java/quickfix/Session.java
```diff
@@ -2685,6 +2693,31 @@ private void resetState() {
...
+     * @param allowPosDup   whether to allow PossDupFlag and OrigSendingTime in the message
+     * @return a status flag indicating whether the write to the network layer was successful.
+     */
+    public boolean send(Message message, boolean allowPosDup) {
+        // Send message as is if allowPosDup flag is set
+        if (allowPosDup) {
+            return sendRaw(message, 0);
+        }
+
+        // Else remove PossDupFlag and OrigSendingTime
```
chrjohn, line 2720:
```suggestion
```
I think this comment is not necessary.

esanchezros:
Gentle reminder for this PR to get merged if possible thanks

chrjohn:
Hi Ed, just back from vacation and slowly starting to pick things up where I left off... :)

esanchezros:
No worries, thanks @chrjohn

chrjohn:
Closing and reopening to trigger complete Java CI checks. When this PR has been created these weren't active yet.

PR #414, opened 2021-09-16
Remove group entry entirely when removing a group

### LINKED ISSUE #413: removeGroup leaves entry in groups
https://github.com/quickfix-j/quickfixj/blob/9df0fe1334b56449a448a985b80df77acae3813b/quickfixj-core/src/main/java/quickfix/FieldMap.java#L672

Why do we clear the list of groups rather than just do:

```
this.groups.remove(field);
```

Is there some purpose behind this?

philipwhiuk:
Fixes #413

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/FieldMap.java
+++ b/quickfixj-core/src/main/java/quickfix/FieldMap.java
@@ -669,7 +669,7 @@ public abstract class FieldMap implements Serializable {
     }
 
     public void removeGroup(int field) {
-        getGroups(field).clear();
+        this.groups.remove(field);
         removeField(field);
     }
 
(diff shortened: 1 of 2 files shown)


chrjohn:
Thanks @philipwhiuk , looks good to me.

quickfixj-core/src/main/java/quickfix/FieldMap.java
```diff
@@ -669,7 +669,7 @@ public void replaceGroup(int num, Group group) {
     }
 
     public void removeGroup(int field) {
-        getGroups(field).clear();
+        this.groups.remove(field);
```
chrjohn, line 672: Just a minor nitpick really: what about `getGroups().remove(field)`? Or at least get rid of the `this` because no other place in that class has it when accessing the `groups` field.

Issue #450, opened 2021-12-06
Add affected `socketAddress` to `SessionStateListener#onConnectException`


stasal:
Please, consider adding `socketAddress` to exception passed as argument for `quickfix.SessionStateListener#onConnectException`

This information is available in `quickfix.mina.initiator.IoSessionInitiator.ConnectTask#handleConnectException` method and it can be very useful for troubleshooting outside the library

chrjohn:
sounds like a sensible enhancement. Are you able to submit a PR for this?

stasal:
@chrjohn Sure! Opened a pull request, take a look. Maybe I need to implement some tests for new code?

chrjohn:
Thanks for the PR, let's continue the discussion there.

PR #451, opened 2021-12-07
Wrap connection exception and add socketAddress to details #450

stasal:
Implementation for the following issue: https://github.com/quickfix-j/quickfixj/issues/450

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/mina/initiator/ConnectionException.java
+++ b/quickfixj-core/src/main/java/quickfix/mina/initiator/ConnectionException.java
@@ -1,0 +1,48 @@
+/*******************************************************************************
+ * Copyright (c) quickfixengine.org  All rights reserved.
+ *
+ * This file is part of the QuickFIX FIX Engine
+ *
+ * This file may be distributed under the terms of the quickfixengine.org
+ * license as defined by quickfixengine.org and appearing in the file
+ * LICENSE included in the packaging of this file.
+ *
+ * This file is provided AS IS with NO WARRANTY OF ANY KIND, INCLUDING
+ * THE WARRANTY OF DESIGN, MERCHANTABILITY AND FITNESS FOR A
+ * PARTICULAR PURPOSE.
+ *
+ * See http://www.quickfixengine.org/LICENSE for licensing information.
+ *
+ * Contact ask@quickfixengine.org if any conditions of this licensing
+ * are not clear to you.
+ ******************************************************************************/
+
+package quickfix.mina.initiator;
+
+import java.io.IOException;
+import java.net.SocketAddress;
+
+public class ConnectionException extends IOException {
+
+    private final SocketAddress socketAddress;
+
+    public ConnectionException(SocketAddress socketAddress) {
+        this.socketAddress = socketAddress;
+    }
+
+    public ConnectionException(String message, Throwable cause, SocketAddress socketAddress) {
+        super(message, cause);
+        this.socketAddress = socketAddress;
+    }
+
+    public ConnectionException(String message, SocketAddress socketAddress) {
+        super(message);
+        this.socketAddress = socketAddress;
+    }
+
+    public ConnectionException(Throwable cause, SocketAddress socketAddress) {
+        super(cause.getMessage(), cause);
+        this.socketAddress = socketAddress;
+    }
+
+}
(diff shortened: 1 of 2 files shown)


quickfixj-core/src/main/java/quickfix/mina/initiator/IoSessionInitiator.java
```diff
@@ -261,12 +261,13 @@ private void handleConnectException(Throwable e) {
                 e = e.getCause();
             }
             final String nextRetryMsg = " (Next retry in " + computeNextRetryConnectDelay() + " milliseconds)";
```
chrjohn, line 263: Was there a specific reason why you decided against your first idea of adding the `socketAddress` as argument?

quickfixj-core/src/test/java/quickfix/SocketInitiatorTest.java
```diff
@@ -237,18 +241,21 @@ public void testInitiatorConnectionException() throws Exception {
...
         AtomicBoolean onConnectExceptionWasCalled = new AtomicBoolean(false);
+        AtomicReference<SocketAddress> socketAddress = new AtomicReference<>(null);
         try {
             SessionID clientSessionID = new SessionID(FixVersions.BEGINSTRING_FIX42, "TW", "ISLD");
             SessionSettings settings = getClientSessionSettings(clientSessionID, freePort);
             settings.setString(clientSessionID, "ReconnectInterval", "1");
-            settings.setString(clientSessionID, "SocketConnectHost", "localhost");
-            settings.setString(clientSessionID, "SocketConnectProtocol", ProtocolFactory.getTypeString(ProtocolFactory.VM_PIPE));
+            settings.setString(clientSessionID, "SocketConnectHost", host);
+            settings.setString(clientSessionID, "SocketConnectProtocol", ProtocolFactory.getTypeString(ProtocolFactory.SOCKET));
```
chrjohn, line 252: Does this not work with `VM_PIPE` addresses or what was the reason for the change?

stasal: It does work with `VM_PIPE`, but I'm not sure if this is a sufficient test, because I didn't find a lot about this protocol.
If I understand it correctly, it doesn't work over network but uses some other means to communicate between local peers located in the same environment or on the same server

If needed, I can switch it back to `VM_PIPE` (and reduce the sleep delay as well), the assertion in this case will spread only to the port value

quickfixj-core/src/main/java/quickfix/mina/initiator/IoSessionInitiator.java
stasal: Do you mean to add `socketAddress` as `quickfix.SessionStateListener#onConnectException` method argument?
It was not my firs idea actually, I proposed to wrap the exception and add some details to this exception

I'm not sure if this is a good approach to change the interface method and its signature, because it will force all the clients to modify their code when it may be not necessary

That's why I just extended the exception details, so after update to new version nobody will need to modify their existing code

quickfixj-core/src/main/java/quickfix/mina/initiator/ConnectionException.java
```diff
@@ -0,0 +1,51 @@
...
+ * Contact ask@quickfixengine.org if any conditions of this licensing
+ * are not clear to you.
+ ******************************************************************************/
+
+package quickfix.mina.initiator;
+
+import java.io.IOException;
+import java.net.SocketAddress;
+
+public class ConnectionException extends IOException {
```
chrjohn, line 25: Sorry for the delay. I have one minor remark: could we rename the class to `ConnectException` so that it matches the name of the callback `onConnectException`. Thank you.

jvirtanen: There's [`java.net.ConnectException`](https://docs.oracle.com/javase/8/docs/api/java/net/ConnectException.html), but I guess that is fine as classes in most applications wouldn't have both this and that in the namespace anyway.

chrjohn: Thanks, that's a valid point. It even also extends `IOException`. But as you said it should be OK.

quickfixj-core/src/main/java/quickfix/mina/initiator/IoSessionInitiator.java
```diff
@@ -261,12 +261,13 @@ private void handleConnectException(Throwable e) {
...
+            ConnectException wrappedException = new ConnectException(e, socketAddress);
             if (e instanceof IOException) {
                 fixSession.getLog().onErrorEvent(e.getClass().getName() + " during connection to " + socketAddress + ": " + e + nextRetryMsg);
-                fixSession.getStateListener().onConnectException(fixSession.getSessionID(), (IOException) e);
+                fixSession.getStateListener().onConnectException(fixSession.getSessionID(), wrappedException);
```
chrjohn, line 267: `fixSession.getStateListener().onConnectException...` could be moved outside of the if block.

Issue #418, opened 2021-10-07
`UTCTimestampConverter` can't handle leapseconds


philipwhiuk:
**Describe the bug**
In the event of a message with a leap-second QFJ will fail to convert it:

**To Reproduce**

Taking the sample leap second value from FIXimate: "19981231-23:59:60"

`UtcTimestampConverter.convertToLocalDateTime("19981231-23:59:60")`

**Expected behavior**

* A LocalDateTime object is returned

**Actual behavior**

`quickfix.FieldConvertError: invalid UTC timestamp value: 19981231-23:59:60`

This happens because LocalDateTime won't parse a leap second.

`Text '19981231-23:59:60' could not be parsed: Invalid value for SecondOfMinute (valid values 0 - 59): 60`

**Additional context**

Given UTC dates are used in sending times, it's probably not sufficient to just assume exchanges won't be open at midnight (which is the usual insertion point).

philipwhiuk:
We probably have to look at http://time4j.net/ or roll our own.

QwertGold:
You could use Date via SimpleDateFormat, but it has some caveats, look at this code
```
SimpleDateFormat format = new SimpleDateFormat("yyyyMMdd-HH:mm:ss");
format.setTimeZone(TimeZone.getTimeZone("UTC"));
Instant instant = format.parse("19981231-23:59:60").toInstant();
System.out.println("instant = " + instant);
```
it outputs `instant = 1999-01-01T00:00:00Z`

This could throw off things related to date handling. 

The catch block solution sounds reasonable, just to add `replace("60", "59")` and re-parse if we get a message during a leap second.

chrjohn:
Hi guys, thanks for your comments so far.

I don't think we should go back to the `java.util.Date` API and use `Date` or `SimpleDateFormat` (not threadsafe). There still might be some places in the code which use stuff from the old API but we definitely should not re-introduce it.

What about an approach similar to this: https://gist.github.com/dehora/9b2ce8bac22d1b37393d8b4f7c9eb7aa
It uses the new API via `DateTimeFormatter` (threadsafe) which we already use in `UtcTimestampConverter`.
Might be sensible to do a small benchmark first which approach is faster. Either to find and replace the leap second on the `String` and use the current approach or the way it is done in the gist above.
Or even time4j, but I don't know if it is feasible to introduce it only for a single corner case. Although I haven't looked at time4j and don't know if it has other useful stuff.

Cheers,
Chris.

[...]

PR #456, opened 2021-12-29
Optimize `convertToLocalDateTime` method

leonchen83:
speed up convertToLocalDateTime method more details see benchmark following.

Signed-off-by: Baoyi Chen <chen.bao.yi@qq.com>

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/field/converter/UtcTimestampConverter.java
+++ b/quickfixj-core/src/main/java/quickfix/field/converter/UtcTimestampConverter.java
@@ -24,10 +24,10 @@ import quickfix.FieldConvertError;
 import quickfix.SystemTime;
 
 import java.text.DateFormat;
+import java.time.DateTimeException;
 import java.time.LocalDateTime;
 import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
-import java.time.format.DateTimeParseException;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.GregorianCalendar;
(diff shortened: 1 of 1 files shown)


leonchen83:
Benchmark
```
import java.time.LocalDateTime;
import java.time.format.DateTimeFormatter;

import org.openjdk.jmh.annotations.Benchmark;
import org.openjdk.jmh.annotations.Fork;
import org.openjdk.jmh.annotations.Measurement;
import org.openjdk.jmh.annotations.Scope;
import org.openjdk.jmh.annotations.State;
import org.openjdk.jmh.annotations.Warmup;

/**
 * @author Baoyi Chen
 */
@Fork(1)
@State(Scope.Benchmark)
@Warmup(iterations = 3, time = 1, timeUnit = TimeUnit.SECONDS)
@Measurement(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
public class ConvertLocalDateTimeBenchmark {
	
	private static final DateTimeFormatter FORMATTER_SECONDS = DateTimeFormatter.ofPattern("yyyyMMdd-HH:mm:ss");
	private static final DateTimeFormatter FORMATTER_MILLIS  = DateTimeFormatter.ofPattern("yyyyMMdd-HH:mm:ss.SSS");
	private static final DateTimeFormatter FORMATTER_MICROS  = DateTimeFormatter.ofPattern("yyyyMMdd-HH:mm:ss.SSSSSS");
	private static final DateTimeFormatter FORMATTER_NANOS   = DateTimeFormatter.ofPattern("yyyyMMdd-HH:mm:ss.SSSSSSSSS");
	
	public static void main(String[] args) throws Exception {
		Options opt = new OptionsBuilder()
				.shouldDoGC(true)
				.include(ConvertLocalDateTimeBenchmark.class.getSimpleName())
				.build();
		new Runner(opt).run();
	}
	
	@Benchmark
	public LocalDateTime benchmarkRaw() {
		return convertToLocalDateTimeRaw("20120922-12:34:56.123456789");
	}
	
	@Benchmark
	public LocalDateTime benchmarkNew() {
		return convertToLocalDateTimeNew("20120922-12:34:56.123456789");
	}
	
	public static LocalDateTime convertToLocalDateTimeNew(String value) {
		int length = value.length();
		int ns = 0;
		if (length >= 27) {
			ns = parseInt(value, 18, 9);
		} else if (length == 24) {
			ns = parseInt(value, 18, 6) * 1000;
		} else if (length == 21) {
			ns = parseInt(value, 18, 3) * 1000000;
		}
		
		int yy = parseInt(value, 0, 4);
		int mm = parseInt(value, 4, 2);
		int dd = parseInt(value, 6, 2);
		int h = parseInt(value, 9, 2);
		int m = parseInt(value, 12, 2);
		int s = parseInt(value, 15, 2);
		return LocalDateTime.of(yy, mm, dd, h, m, s, ns);
	}
	
	private static int parseInt(String value, int off, int len) {
		int num = 0;
		for (int index = 0; index < len; index++) {
			num = (num * 10) + value.charAt(off + index) - '0';
		}
		return num;
	}
	
	public static LocalDateTime convertToLocalDateTimeRaw(String value) {
		int length = value.length();
		switch (length) {
			case 17:
				return LocalDateTime.parse(value, FORMATTER_SECONDS);
			case 21:
				return LocalDateTime.parse(value, FORMATTER_MILLIS);
			case 24:
				return LocalDateTime.parse(value, FORMATTER_MICROS);
			case 27:
			case 30:
				return LocalDateTime.parse(value.substring(0, 27), FORMATTER_NANOS);
			default:
				throw new UnsupportedOperationException();
		}
	}
}

```

Benchmark result
```
Benchmark                                    Mode  Cnt         Score         Error  Units
ConvertLocalDateTimeBenchmark.benchmarkNew  thrpt    5  34913770.478 ± 3381097.452  ops/s
ConvertLocalDateTimeBenchmark.benchmarkRaw  thrpt    5   1335947.490 ±   72909.323  ops/s
```

30 times faster than before

chrjohn:
Thanks @leonchen83 , I'll have a look.

raipc:
@chrjohn @leonchen83 I think that decrementing the leap second is not the best solution since we'll have time going downward. Would moving to the next second with zero nanos be sufficient?

chrjohn:
Hi @raipc @leonchen83 @philipwhiuk 

sorry for the delay. So it seems we have two options if we don't want time to go backwards (related to the previous message):

Either store the last nanos as outlined here: https://github.com/quickfix-j/quickfixj/issues/418#issuecomment-1005331199

Or can't we (for the sake of simplicity) assume that nanos are 999999? E.g. convert `19981231-23:59:60.444444` to `19981231-23:59:59.999999`

Did I miss something?

leonchen83:
hi @chrjohn 
or we could just optimize convertToLocalDateTime method and ignore leap second problem. one PR fix one thing?
WDYT？

chrjohn:
Yes a separate PR would be good in any case. Thanks

chrjohn:
Could you please open a new (or change this) PR to base this on `master`? If needed I could cherry pick it to a minor version later on.
Thanks

leonchen83:
hi
refer to [https://github.com/quickfix-j/quickfixj/pull/529](PR529)

chrjohn:
Great, thanks. Could you open a separate PR against master with the original changes, i.e. also the leap second stuff? It should be no problem to merge it after #529 and after we finished the discussion regarding the nanos.

leonchen83:
@chrjohn 
2 commits on one PR, is that OK?

chrjohn:
Sorry for not being clear. I meant one PR with your optimization and another one with your optimization including the leap second stuff (both against master branch). So I would be able to merge the first one separately. 
Thank you again

chrjohn:
Superseded by #529

PR #457, opened 2022-01-04
Support for FIX Latest 

david-gibbs-ig:
Support for FIX Latest #317

This is not ready to merge and is submitted to attract comments.

Use of the  FTC Code Generator copied from the https://github.com/FIXTradingCommunity/fix-orchestra-quickfix is needed as the XSL based QFJ code generation takes too long for FIXLatest and the C.I. builds all fail. 

The FTC Code Generator is copied into the  repository-quickfix module in this repo. This is because:  the "reason for change" is changes to QFJ and also to avoid a cyclic dependency between the fix-orchestra-quickfix build and the QFJ build. Hopefully this will be agreeable with the FIX Orchestra working group, if not then the  FTC Code Generator can be moved to a different new repo in the FTC organization to break the cyclic dependency.  The code in this repository has changes that are not in the original repo.

The convention used in Orchestra for codeset enumerations can lead to incorrect static fields names so we would want to review every single one, you can find the work around in XSL transforms in the orchestration module. Going forward we would need to review changes to codesets and new codesets.

This project now also publishes a QuickFIXJ orchestration, updated to result in a correct QFJ build.  This can then be used to "easily" derive and build QFJ custom orchestrations, without needing to understand the work required to conform the fix standard orchestration to QFJ. This should be more maintainable than managing forks of QFJ to cope with ROE differences and enable more succinct code, document and test generation.

I've tested this approach in  branch "use-orchestration-from-qfj-project" of ig-orchestrations. 
see 
https://github.com/IG-Group/ig-orchestrations/tree/use-orchestration-from-qfj-project. 
https://github.com/IG-Group/ig-orchestrations/tree/use-orchestration-from-qfj-project/ig-us-rfed/orchestration
https://github.com/IG-Group/ig-orchestrations/blob/use-orchestration-from-qfj-project/ig-us-rfed/quickfixj-messages-fixLatest/pom.xml
I have tested the resulting customized message package end to end with private QFJ FIX applications.

Also note : 

* core module is now only core, excluding fields, this will prevent duplication/collisions/conflict as they are present in message packages

* I apologize for too much whitespace change, my Eclipse just keeps crashing so I was unable to apply the formatter. If this work is otherwise acceptable I'm sure we can format the artefacts.

* MessageTest.shouldConvertToXMLWithIndent and MessageTest.shouldConvertToXMLWithoutIndent were failing, not sure why yet.

* build with profile "-PskipBundlePlugin" speeds things up a little

* I've commented out javadoc in a few places, <skip>true</skip> did not appear to work. If we close in on a merge candidate we can turn javadoc back on.

I hope this is interesting, Happy New Year.


david-gibbs-ig:
@chrjohn The CodeQL step fails - with OutOfMemoryError. Is the heap size something that you can configure ? 

  [2022-01-07 09:32:33] [autobuild] compiler message file broken: key=compiler.misc.msg.bug arguments=11.0.11, {1}, {2}, {3}, {4}, {5}, {6}, {7}
  [2022-01-07 09:32:33] [autobuild] java.lang.OutOfMemoryError: Java heap space

david-gibbs-ig:
Please see the new pull request which has some structural changes _intended_ to make it quicker to work on core.

PR #460, opened 2022-01-13
Fix orchestra with updated project structure

### LINKED ISSUE #317: Support for FIX Latest

david-gibbs-ig:
Fixes #317

This branch has structural changes that are _intended_ to make it quicker to work with core. 

Core no longer has a mutual dependency with **ALL** the source code for fields, components and messages for all the FIX versions. Core depends on a new module "Base" as do the generated Message, Component and Field classes. Base does depend on a small number of generated classes that are used in Standard Header and Footer. These classes are generated with and packaged with Base, but only these classes. This is to help decouple the messages resources from the core distribution and facilitate the build and maintenance of custom distributions. Core now has only test dependencies on the message packages.  This is described at length in project markdown files.

There is a benefit to this structure;  "core"  can be developed without rebuilding the message packages. Build times for the message packages are long. The FIX Orchestration now has nearly 6000 fields. One only needs to rebuild the message packages when code generation is altered, when a new FIX Standard Orchestration is incorporated or when Base is changed. As a further convenience there is a minimal profile to reduce the size -P"minimal-fix-latest". This This makes fix latest much smaller but includes resources that are needed to test core.

Use of the FTC Code Generator copied from the https://github.com/FIXTradingCommunity/fix-orchestra-quickfix is introduced via a dependency on https://github.com/quickfix-j/quickfixj-orchestra. The XSL based QFJ code generation takes too long for FIXLatest and this was causing C.I. builds to fail. This FTC Code Generator  is used for fixt11 and fixlatest to help with future compatibility. Other protocol versions are built with the old code generator to all options to build custom packages with very high backward compatibility but benefitting from the new core (and base). 

The convention used in Orchestra for codeset enumerations can lead to sub-optimal constant names. Work is in progress with the FTC to work around this. Where FIX latest is part of the build some constant names will be different from legacy builds. As mentioned above, there are work arounds for this.

You'll see that the new code generation for fixlatest does not de-normalize components. I had extended the code generation to do so and more closely mirror the QFJ code generator but this resulted in enormous classes with extreme build times for compilation and javadoc. I suggest that we adopt normalised components for fixlatest as the convention for this qfj distribution. If you build from orchestra you will get normalised components in the fixlatest package. This could be an optional setting but I think it's a bad idea. There are no actual "components" in fixt11, only groups.

This project now also publishes a QuickFIXJ orchestration,  modified to help with a correct QFJ build. This can then be used to "easily" derive and customise QFJ builds, without needing to understand the work required to conform the standard orchestration to QFJ. 
Customised builds should now be more maintainable, there is no need to manage forks of QFJ to cope with ROE differences and this will enable more succinct code, documents and test generation.

I've tested this approach in branch "use-orchestration-from-qfj-project" of ig-orchestrations.
see:
https://github.com/quickfix-j/quickfixj-orchestra
https://github.com/IG-Group/ig-orchestrations/tree/use-orchestration-from-qfj-project.
https://github.com/IG-Group/ig-orchestrations/tree/use-orchestration-from-qfj-project/ig-us-rfed/orchestration
https://github.com/IG-Group/ig-orchestrations/blob/use-orchestration-from-qfj-project/ig-us-rfed/quickfixj-messages-fixLatest/pom.xml
I have also tested the resulting customized message package end to end with private QFJ FIX applications.


david-gibbs-ig:
@chrjohn  Is there a build setting to allow snapshots to be obtained from a build-local or  snapshot repo ? This build tries to depend on artefacts in a few places rather than reading across the local modules (which I think is less ideal).

david-gibbs-ig:
All the tests are failing as I imagined (ok assumed) that a "build-local" maven repo would be available for snapshots. Can we discuss ?

https://stackoverflow.com/questions/57695362/github-actions-cache-repo-to-speed-up-maven-builds

david-gibbs-ig:
@chrjohn Hi, I would really welcome some feedback on this, as there are some structural changes that will make this really hard to maintain in line with master as time goes on. I think that the best way to work with this is to first check out the branch and build it,  and to look at the readmes. There are some things I don't like about this structure but it works, it builds fast and can help with productive maintenance of QFJ. It is intended to make it quite easy for users to maintain their custom orchestration with type safety without needing manipulate the base/core QFJ distribution.

chrjohn:
Reopened to trigger build jobs.

README.md
```diff
@@ -300,18 +300,23 @@ This project builds artefacts for the standard published FIX specification versi
...
 
-If you have no need to specialise a FIX integration then you can use the ```org.quickfixj``` artefacts built by this project. Simply include them as dependencies of your application.
+If you have no need to customise a FIX integration then you can use the ```org.quickfixj``` artefacts built by this project. Simply include them as dependencies of your application.
 
-Artefacts for unused FIX specification versions can be omitted from your runtime. Many integrations will not require ```quickfixj-messages-all``` and need only depend on artefacts for a subset of the FIX standard versions. Please note that FIX Protocol verisions 5.0 and later depend on ```quickfixj-messages-fixt11``` to provide the implementation for the FIXT1.1 transport messages.
+Artefacts for unused FIX specification versions can be omitted from your runtime. 
+Many integrations will not require ```quickfixj-messages-all``` and need only depend on artefacts for a subset of the FIX standard versions. Please note that FIX Protocol verisions 5.0 and later depend on ```quickfixj-messages-fixt11``` which provides the implementation for the FIXT1.1 transport messages.
```
chrjohn, line 308:
```suggestion
Many integrations will not require ```quickfixj-messages-all``` and need only depend on artefacts for a subset of the FIX standard versions. Please note that FIX Protocol versions 5.0 and later depend on ```quickfixj-messages-fixt11``` which provides the implementation for the FIXT1.1 transport messages.
```

chrjohn:
Just tried to resolve the conflicts, hopefully I didn't mess something up...

david-gibbs-ig:
> Just tried to resolve the conflicts, hopefully I didn't mess something up...

No, I think it's exactly the same as I did. Will look at the errors when I get a chance.

david-gibbs-ig:
This is a test failure I have seen, I don't understand why the output seems different, possibly changes I made to XML dependencies. I'll look to see how the test expectations compare to master. I might have gone with what was in master when I merged master in...
[INFO] Tests run: 70, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 3.119 s - in quickfix.SessionTest
[INFO] 
[INFO] Results:
[INFO] 
Error:  Failures: 
Error:    MessageTest.shouldConvertToXMLWithIndent:1979 expected:<...ncoding="ISO-8859-1"[?>
<message>
   <header/>
   <body>
      <field tag="1">test-account</field>
   </body>
]   <trailer/>
</mess...> but was:<...ncoding="ISO-8859-1"[ standalone="no"?>
<message>
    <header/>
    <body>
        <field tag="1"><![CDATA[test-account]]></field>
    </body>
 ]   <trailer/>
</mess...>
Error:    MessageTest.shouldConvertToXMLWithoutIndent:1966 expected:<...ncoding="ISO-8859-1"[?><message><header/><body><field tag="1">test-account]</field></body><trai...> but was:<...ncoding="ISO-8859-1"[ standalone="no"?><message><header/><body><field tag="1"><![CDATA[test-account]]>]</field></body><trai...>
Error:    MessageTest.shouldConvertToXmlWhenDataDictionaryLoadedWithExternalDTD:1959 expected:<...ncoding="ISO-8859-1"[?><message><header/><body><field name="Account" tag="1">test-account]</field></body><trai...> but was:<...ncoding="ISO-8859-1"[ standalone="no"?><message><header/><body><field name="Account" tag="1"><![CDATA[test-account]]>]</field></body><trai...>
[INFO]

david-gibbs-ig:
> > possibly changes I made to XML dependencies
> 
> I'd suspect the same.

XML errors resolved.

chrjohn:
> > > possibly changes I made to XML dependencies
> > 
> > 
> > I'd suspect the same.
> 
> XML errors resolved.

What was the problem? It seems an additional attribute somehow was added?

david-gibbs-ig:
> What was the problem? It seems an additional attribute somehow was added?

The output of Message.toXML(...) must express the field values as CDATA 
i.e. <field tag="1"><![CDATA[test-account]]></field></field> . 

The "toXML" code in QFJ deliberately treats the fields as CDATA, this make good sense. See line 408 of Message (in quickfixj-base in this branch, or quickfixj-core in master); 
408 : "final CDATASection value = document.createCDATASection(field.getObject().toString());"

At some point on this branch the field data was not being out as CDATA, the test expectations are now the same as on master.

david-gibbs-ig:
Just making a note here that I need to look a changes to SerialisationTest in quickfixj-core.

david-gibbs-ig:
> I tried to build with `-Pminimal-fix-latest` but got `The requested profile "minimal-fix-latest" could not be activated because it does not exist.`

Will have a look asap

chrjohn:
I think it was a problem on my end. I did a fresh clone of your repo and branch and it worked.

chrjohn:
But I get a lot of JavaDoc errors and the build fails because of that (when I omit the `-Dmaven.javadoc.skip=true` option).
Just a question: by which criterion did you decide what goes into ?quickfixj-base`? Just wondering whether it would be good to have that as a separate PR. I could do that if you like.

david-gibbs-ig:
> But I get a lot of JavaDoc errors and the build fails because of that (when I omit the `-Dmaven.javadoc.skip=true` option). Just a question: by which criterion did you decide what goes into `quickfixj-base`? Just wondering whether it would be good to have that as a separate PR. I could do that if you like.

 `quickfixj-base` is meant to be only the immediate dependencies of the generated code and there are some other important and stable things that do not depend on the generated messages. Of course the way it is written, it still depends on some fields that are in std headers so these alone are generated for this module. One of the things that is suboptimal is that many tests must remain in  `quickfixj-core`, for example it may be possible to pull some tests out of MessageTest in core, ones that don't require all the message builds. As you know this is all so that core->messages->base. For a QFJ developer, this means you can do mvn clean install `quickfixj-base`; mvn clean install `quickfixj-messages`; and then iterate more quickly over builds of core without needing to rebuild those modules. This is motivated by "Stable dependencies principal". Similarly `quickfixj-core` now only has test dependencies on  `quickfixj-messages`. This is so that the message generation (and FIX specs) can be verified by the "home project" and prove that these "home" artefacts are demonstrably safe to use, but allows for much "easier" generation of custom message libraries by allowing them to be dropped in at run time . Custom libs can even published on maven central without conflicts with qfj namespace.
 
I don't think we can refactor base independently from much of the rest of the PR. It would still depend on the fields and messages in the existing way of generating code. Once we see that a lot of those changes are just moving things to base, there is not as much left to understand as it would first appear. Quite a lot of it is documents.

There are a number of changes related to preferring the enumeration names and similar from FIX Latest. I have tried to design  the solution so that existing QFJ users could even quite easily work around that by omitting FIX Latest from their custom message builds. This PR proposed though that the home project though, prefers FIX Latest field names etc...

I did clean up some mvn errors in the existing code generation plugin, that can be independent. The generation of the QuickFIX/J Orchestration can be added independently but is of no real use.

chrjohn:
I hope you don't get me wrong. I am all for this PR, I just need to look at it from the maintainer's perspective. :) I am just thinking how we could get this (or a new) PR a little more to a state which could be comprehended more easily (w.r.t. the changes to the structure of the project).

E.g. the history contains old changes (e.g. related to Travis) which are outdated now. I know this is my fault since it took me a long time to look more thoroughly at this PR. :-/ This (and generally the large number of commits) makes it very hard for me to make sure whether all changes done to the classes in `core` in the last months are also part of this PR.

david-gibbs-ig:
Understood, I have tried to keep up but even with git it is not always easy. Core should be quite straightforward but I did struggle with the last merge. I think that we need to have an extra step to compare checked out branches on the file system, just to be certain some things have not gone backwards.

david-gibbs-ig:
@chrjohn I think we should build some examples of Client Applications that build cut down/simple examples of custom message builds as discussed in the docs. I proved the concept with work on FIX APIs I was doing for my former employer and  but we can make it explicit and compact and visible for users to explore. It can have integration tests of one kind or another. The examples could be in a multi-module project in a new repo that should be separate from the quickfix/j repo. 
It can look like : 
* Here is a custom orchestration (derived from the published quickfix/j orchestration)
* Here is one way to generate and package your messages 
* Here is another way to generate and package the messages
* Here is a build that uses the latest QuickFIX/J but builds using only using the FIX Dictionary and without FIX Latest, to provide an upgrade pathway for QuickFIX/J 2.n users. They can test QuickFIX/J 3.n against existing code with minimal changes.

I would suggest removing the existing examples from the home quickfixj repo and moving them to this new repo.

The tests provide a way to verify that new QFJ releases don't break this scheme.

quickfixj-examples/executor/src/main/java/quickfix/examples/executor/Application.java
```diff
@@ -385,7 +385,7 @@ public void onMessage(quickfix.fix50.NewOrderSingle order, SessionID sessionID)
             Price price = getPrice(order);
 
             quickfix.fix50.ExecutionReport accept = new quickfix.fix50.ExecutionReport(
-                    genOrderID(), genExecID(), new ExecType(ExecType.NEW), new OrdStatus(
+                    genOrderID(), genExecID(), new ExecType(ExecType.TRADE), new OrdStatus(
```
chrjohn, line 388: Shouldn't this be `NEW` still? Cumulated qty is 0.

quickfixj-examples/executor/src/main/java/quickfix/examples/executor/Application.java
```diff
@@ -350,7 +350,7 @@ public void onMessage(quickfix.fix44.NewOrderSingle order, SessionID sessionID)
         Price price = getPrice(order);
 
         quickfix.fix44.ExecutionReport accept = new quickfix.fix44.ExecutionReport(
-                    genOrderID(), genExecID(), new ExecType(ExecType.NEW), new OrdStatus(
+                    genOrderID(), genExecID(), new ExecType(ExecType.TRADE), new OrdStatus(
```
chrjohn, line 353: Shouldn't this be `NEW` still? Cumulated qty is 0.

quickfixj-examples/executor/src/main/java/quickfix/examples/executor/Application.java
```diff
@@ -315,7 +315,7 @@ public void onMessage(quickfix.fix43.NewOrderSingle order, SessionID sessionID)
         Price price = getPrice(order);
 
         quickfix.fix43.ExecutionReport accept = new quickfix.fix43.ExecutionReport(
-                    genOrderID(), genExecID(), new ExecType(ExecType.NEW), new OrdStatus(
+                    genOrderID(), genExecID(), new ExecType(ExecType.TRADE), new OrdStatus(
```
chrjohn, line 318: Shouldn't this be `NEW` still? Cumulated qty is 0.

quickfixj-examples/banzai/src/main/java/quickfix/examples/banzai/BanzaiApplication.java
```diff
@@ -1,5 +1,4 @@
 /*******************************************************************************
- * Copyright (c) quickfixengine.org  All rights reserved.
```
chrjohn, line 2: Probably deleted by accident...

pom.xml
```diff
@@ -223,18 +373,118 @@
 						</execution>
 					</executions>
 				</plugin>
+				<plugin>
+					<groupId>org.codehaus.mojo</groupId>
+					<artifactId>xml-maven-plugin</artifactId>
+					<version>${xml-maven-plugin-version}</version>
+				</plugin>
+				<plugin>
+					<groupId>org.jboss.tattletale</groupId>
```
chrjohn, line 382: IIRC you removed that dependency from the other PR #493

chrjohn:
@david-gibbs-ig I think those examples would help a lot! There are often questions dealing with customising data dictionaries and the build respectively.

> I would suggest removing the existing examples from the home quickfixj repo and moving them to this new repo.

That is probably a good idea, too. I could do that. I think it shouldn't interfere with your changes (apart from minor changes you did that I just commented).

david-gibbs-ig:
> In general I still have headaches about tests being separate from the tested class (e.g. `Message` in `base` but `MessageTest` in `core`. Maybe this needs to be split up in two separate tests then?

Yes -  I agree 100% . As a hasty check I commented out the Message package imports in MessageTest and I think there are a number of message tests that can be moved to a MessageTest(Base) in  quickfixj-base.

quickfixj-examples/executor/src/main/java/quickfix/examples/executor/Application.java
david-gibbs-ig: Yes, well spotted.

quickfixj-examples/banzai/src/main/java/quickfix/examples/banzai/BanzaiApplication.java
david-gibbs-ig: yes, good spot

quickfixj-examples/executor/src/main/java/quickfix/examples/executor/Application.java
david-gibbs-ig: yes

quickfixj-examples/executor/src/main/java/quickfix/examples/executor/Application.java
david-gibbs-ig: yes

pom.xml
david-gibbs-ig: I have update PR for the above comments, excepting the javadoc.

david-gibbs-ig:
@chrjohn I've moved lots of classes back to core to reduce the number of changes vs master, and to conform to the principle that quickfixj-base should contain, as far as possible, only the things that are dependencies of the generated code.

customising-quickfixj.md
```diff
@@ -0,0 +1,77 @@
...
+```
+	<properties>
+		<generator.decimal>true</generator.decimal>
+	</properties>
+```
+See [QuickFIX/J Messages](./quickfixj-messages/readme.md) for details of the build and recommendation for  **how to implement custom builds.**
+
+### **Incompatible Data Types**
+
+Some incompatable changes have occurred in the evolution of the FIX protocol. For example see below changes to the type of **OrderQty (38)** :
```
chrjohn, line 37:
```suggestion
Some incompatible changes have occurred in the evolution of the FIX protocol. For example see below changes to the type of **OrderQty (38)** :
```

customising-quickfixj.md
```diff
@@ -0,0 +1,77 @@
...
+|FIX Version|Field Name|FIX Datatype|Base Type|QuickFIX/J Datatype|
+|---|---|---|---|---|
+|4.0|OrderQty|int|int|```int```|
+|4.2|OrderQty|Qty|float|```Double``` or ```BigDecimal```|
+
+Only one ```quickfix.Field``` class with the same name may be loaded by the Java classloader so only one version of this Field should be in the classpath. QuickFix/J also verifies the data type using the supplied QuickFIX "Dictionary". 
+
+Code generation using ```BigDecimal``` is incompatible at runtime with ```int``` for **OrderQty**. In this case, ```double``` is compatible with ```int``` at run time due to [widening primitive conversion](http://titanium.cs.berkeley.edu/doc/java-langspec-1.0/5.doc.html). 
+
+Runtime incompatabilities can be resolved by:
```
chrjohn, line 48:
```suggestion
Runtime incompatibilities can be resolved by:
```

customising-quickfixj.md
```diff
@@ -0,0 +1,77 @@
...
+
+See [QuickFIX/J Messages](./quickfixj-messages/readme.md) for details of the build and recommendation for  **how to implement custom builds.**
+
+### **Managing incompatibility with Prior Versions of QuickFIX/J**
+
+From QuickFIX/J 3.0.0 the code generation for ```quickfix.Field``` prefers the FIX Orchestra Standard. This results in incompatible changes to the names of constants. 
+
+For example : ```SettlType.REGULAR_FX_SPOT_SETTLEMENT``` becomes ```SettlType.REGULAR```. 
+
+The required code changes may be trvial in most cases but they are elective. To use ```quickfix-core``` from QuickFIX/J  3.0.0 or later without needing to implement code changes:
```
chrjohn, line 61:
```suggestion
The required code changes may be trivial in most cases but they are elective. To use ```quickfix-core``` from QuickFIX/J  3.0.0 or later without needing to implement code changes:
```

customising-quickfixj.md
```diff
@@ -0,0 +1,77 @@
...
+* build the required Message artefacts without the FIX Latest code generation. The Fields will then be generated only from legacy FIX Protocol Versions as they were prior to QuickFIX/J 3.0.0 - **or**
+* if you want to use Messages, Components and/or Fields from FIX Latest while preferring legacy constants, manipulate the order of code generation and/or the over-write behaviour to prefer earlier versions of FIX. For example, prevent FIX Latest code generation from over-writing existing generated sources (Fields).
+
+See [QuickFIX/J Messages](./quickfixj-messages/readme.md) for details of the build and recommendation for  **how to implement custom builds.**
+
+### **Customising the FIX Protocol for specialised Rules of Engagement**
+
+A Rules of Engagement can include customisation Messages, Components and Fields, including User Defined elements.
+
+This can be accomplised by editing the QuickFIX Dictionary or FIX Orchestra Repository (Orchestration) for the specific Rules of Engagement.
```
chrjohn, line 71:
```suggestion
This can be accomplished by editing the QuickFIX Dictionary or FIX Orchestra Repository (Orchestration) for the specific Rules of Engagement.
```

pom.xml
```diff
@@ -89,8 +90,151 @@
 		<maven-deploy-plugin-version>3.0.0</maven-deploy-plugin-version>
 		<nexus-staging-maven-plugin-version>1.6.13</nexus-staging-maven-plugin-version>
 		<build-helper-maven-plugin-version>3.3.0</build-helper-maven-plugin-version>
+		<maven-plugin-api.version>3.8.7</maven-plugin-api.version>
+		<maven-shared-utils.version>3.3.4</maven-shared-utils.version>
+		<file-management.version>3.0.0</file-management.version>
+		<maven.version>3.3.9</maven.version>
```
chrjohn, line 96: Duplicate to line 78

class-pruner-maven-plugin/src/it/settings.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+
+<settings>
+  <profiles>
+    <profile>
+      <id>it-repo</id>
+      <activation>
+        <activeByDefault>true</activeByDefault>
```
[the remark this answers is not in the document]
david-gibbs-ig: This is all generated by the Maven Pugin development archetype. https://maven.apache.org/archetypes/maven-archetype-plugin/ It is meant to provide an "integration test". We don't use that profile so I will the profile and "IT".

class-pruner-maven-plugin/pom.xml
```diff
@@ -0,0 +1,143 @@
...
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.apache.maven.plugin-testing</groupId>
+			<artifactId>maven-plugin-testing-harness</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>com.google.guava</groupId>
+			<artifactId>guava</artifactId>
```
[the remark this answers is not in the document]
david-gibbs-ig: This was managing a transitive dependency from maven-core dependency, but now this does not matter with mvn dependencies set to provided, so removed dependency management.

pom.xml
david-gibbs-ig: All above resolved - except javadoc

david-gibbs-ig:
@chrjohn Hi, I can't see the snyk report to see what the problem is, and I can't get snyk working on my forked repo. Can you tell me what the issue is please ?

chrjohn:
Screenshot_20230204_184602_Samsung Internet

Sorry for screenshot, currently on mobile.

david-gibbs-ig:
> 

Haha, OK that was what the Gauva dependency management was for :-)

chrjohn:
Do you know which plugin triggers this?

david-gibbs-ig:
> Do you know which plugin triggers this?

It is a transitive dependency of "org.apache.maven:maven-core". scope of guava is now "provided" rather than "test" to match the expected scope of maven. At some point I changed the scope of maven things to "provided" as advised by warning messages but I guess I left "guava" as test. Reported security vulnerability should be resolved by managing the transitive dependency.  Later I will conform if this is resolved in later versions of maven.

chrjohn:
Please excuse the dumb question, am not in reach of my dev machine now. What is maven core plugin used for? Don't recall that we had that dependency before and also have a maven mojo for the code generator.

david-gibbs-ig:
The maven-core dependency is generated when the standard maven plugin development archetype is used.  
https://maven.apache.org/guides/plugin/guide-java-plugin-development.html#Mojo_archetype 
It's basically maven's core java lib,  for example see https://maven.apache.org/ref/3.8.7/maven-core/

I think that they include it to assure the explicit control over the built artefact dependencies. The scope is however "provided" which means it is in effect coming from the maven runtime. You can see the benefit actually of doing this in that the security scanning was able to pick up on the "vulnerability" which is actually in the maven 3.5.0 distribution. As you see this is not of course a vulnerability in QuickFIX/J, it is a vulnerability in maven. 

I'm just working through a build with maven 3.8.7, which has an updated guava dependency and several other vulnerability fixes. I'm thinking that if we are publishing maven plugins, it's probably worth these formalities, including the enforcement of the maven version (which was added 5 years ago). N.B. I remember replacing "<prerequisites>" in the parent pom with the maven-enforcer-plugin as specifically advised by maven warning messages.

image

image

david-gibbs-ig:
@chrjohn I tried removing the **provided** maven dependencies but I did change more than one thing at a time. Could you tell me please what the current snyc objection is ? Is it still guava ?

chrjohn:
Yes, still guava. Can you access snyk now?
If everything else fails I think we can make that check ignored on snyk.

david-gibbs-ig:
> Yes, still guava. Can you access snyk now? If everything else fails I think we can make that check ignored on snyk.

Yes I can thank you,

david-gibbs-ig:
@chrjohn Here is what I landed on : the **provided** maven dependencies are still in the maven plugin pom, this way no information is lost and it is explicit that these dependencies are provided by the maven runtime, for those that are. I bumped back up to maven 3.8.7 to get the latest fixes. Guava is still reported as an issue so I added this to the dependency management section only. The guava dependency management has no effect on the rest  of the  QuickFIX/J build as it is not used and is also **provided** . Neither does not affect what is actually **provided** but it's good information. When this is fixed in a later maven release it can be removed from dependency management, until then it does no harm and it satisfies Snyc. Phew ! P.S. a similar approach would work with maven 3.5.0 if this is your preference.

david-gibbs-ig:
@chrjohn I think I have worked through and resolved the javadoc errors. I have also corrected javadoc warnings in other parts of the project. I have updated documents and added readmes in the message modules.

chrjohn:
Thanks @david-gibbs-ig , I'll work through it.

chrjohn:
One thing I noticed (don't know if that is related to your changes but it worked on 2.3.1): when building I can see that javadocs are generated but they are not added to the `distribution` zip file under `doc/api`. Do you have any idea what could cause this?

david-gibbs-ig:
> One thing I noticed (don't know if that is related to your changes but it worked on 2.3.1): when building I can see that javadocs are generated but they are not added to the `distribution` zip file under `doc/api`. Do you have any idea what could cause this?

Oh does distribution work for you locally ? I thought distribution not working was a known issue.  I can have a look tonight and see if I can get dist working in my build.

 Anyway when I look at the PR I can see that I changed something in the javadoc plugin config : `<includeDependencySources>false</includeDependencySources>`

This was probably while I was wrestling javadoc and when I did not have the fix latest minimal profile. 

Some people still seem to be using the distribution, I'm not sure why. If you want to keep building the dist, then at some point we can think about fixlatest full vs minimal. That's something to discuss separately to not overcomplicate this thread. I guess traditionally the dist would contain fixlatest full, fixlatest is big that's all :-)

image

chrjohn:
The distribution is also uploaded here: https://github.com/quickfix-j/quickfixj/releases
But don't know how much people use it...

david-gibbs-ig:
@chrjohn  something I did not expect is happening in the packaging of quickfixj-core, will have a look
[INFO] --- maven-shade-plugin:3.4.1:shade (create-self-executing-jar) @ quickfixj-examples-executor ---
[INFO] Including org.quickfixj:quickfixj-core:jar:3.0.0-SNAPSHOT in the shaded jar.
[INFO] Including org.quickfixj:quickfixj-base:jar:3.0.0-SNAPSHOT in the shaded jar.
[INFO] Including org.apache.mina:mina-core:jar:2.1.6 in the shaded jar.
[INFO] Including org.quickfixj:quickfixj-messages-all:jar:3.0.0-SNAPSHOT in the shaded jar.
[INFO] Including org.slf4j:slf4j-api:jar:2.0.6 in the shaded jar.
[INFO] Including org.slf4j:slf4j-jdk14:jar:2.0.6 in the shaded jar.
[WARNING] quickfixj-base-3.0.0-SNAPSHOT.jar, quickfixj-core-3.0.0-SNAPSHOT.jar define 156 overlapping classes:

chrjohn:
I'm pretty sure that was there before with the old release.

david-gibbs-ig:
> I'm pretty sure that was there before with the old release.

Quite possibly, very mysterious things are happening, at least some of them due to the bundle plugin.  I can see where someone tried to work around similar in the configuration of the jar plugin. The bundle plugin tramples all those configs though :-) 
I will follow up on this more tomorrow.

david-gibbs-ig:
Quite odd, as well as javadoc "out of memory", now seeing this intermittently on the builds.

image

chrjohn:
@david-gibbs-ig , regarding javadoc. IMHO it would be OK if the build on Github CI has JavaDoc turned off. I think you could configure the javadoc memory so that it runs on a local system. It does not make sense to build it on Github CI. What do you think?

quickfixj-base/src/main/java/quickfix/Dictionary.java
```diff
@@ -25,7 +25,7 @@
 /**
  * Name/value pairs used for specifying groups of settings.
  *
- * @see SessionSettings
+ * see SessionSettings
```
chrjohn, line 28: Is there a reason why you removed that `@see` javadoc tag?

quickfixj-base/src/main/java/quickfix/MessageFactory.java
```diff
@@ -24,7 +24,7 @@
 /**
  * Used by a Session to create a Message.
  *
- * @see quickfix.Session
+ * quickfix.Session
```
chrjohn, line 27: Same here `@see`

quickfixj-base/src/main/java/quickfix/MessageFactory.java
```diff
@@ -53,9 +53,9 @@ default Message create(String beginString, ApplVerID applVerID, String msgType)
      * Creates a group for the specified parent message type and
      * for the fields with the corresponding field ID
      *
-     * Example: to create a {@link quickfix.fix42.MarketDataRequest.NoMDEntryTypes}
+     * Example: to create a {quickfix.fix42.MarketDataRequest.NoMDEntryTypes}
      * you need to call
```
chrjohn, line 57: Same here but with `@link`

quickfixj-base/src/main/java/quickfix/MessageUtils.java
```diff
@@ -118,79 +117,6 @@ public static Message parse(MessageFactory messageFactory, DataDictionary dataDi
         return message;
     }
 
-    /**
```
chrjohn, line 121: This removed stuff went to `MessageSessionUtils`, right?

quickfixj-base/src/main/java/quickfix/UtcDateOnlyField.java
```diff
@@ -34,7 +35,11 @@ public UtcDateOnlyField(int field) {
     protected UtcDateOnlyField(int field, LocalDate data) {
         super(field, data);
     }
-        
+    
+    protected UtcDateOnlyField(int field, String data) {
```
chrjohn, line 39: Is this correct? If I am not mistaken then ISO dates could also bear a time zone offset which UTC dates do not have.

chrjohn (commented):
Worked through some more files... :)

.github/workflows/maven.yml
```diff
@@ -28,5 +28,5 @@ jobs:
           java-version: ${{ matrix.java }}
       - name: Test with Maven
         env:
-          MAVEN_OPTS: "-Xms3g -Xmx3g"
-        run: mvn test -B -V -D"java.util.logging.config.file"="logging.properties" -D"http.keepAlive"="false" -D"maven.wagon.http.pool"="false" -D"maven.wagon.httpconnectionManager.ttlSeconds"="120"
+          MAVEN_OPTS: "-Xms3g -Xmx5g -Djdk.xml.xpathExprGrpLimit=500 -Djdk.xml.xpathExprOpLimit=500"
+        run: mvn install  -B -V -Pminimal-fix-latest -D"java.util.logging.config.file"="logging.properties" -D"http.keepAlive"="false" -D"maven.wagon.http.pool"="false" -D"maven.wagon.httpconnectionManager.ttlSeconds"="120"
```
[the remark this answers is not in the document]
david-gibbs-ig: I can try putting it back , I think that I changed it because of a problem with dependencies

david-gibbs-ig:
> @david-gibbs-ig , regarding javadoc. IMHO it would be OK if the build on Github CI has JavaDoc turned off. I think you could configure the javadoc memory so that it runs on a local system. It does not make sense to build it on Github CI. What do you think?

I am personally not a great advocate of javadoc but on projects like this I do think it has an important role. If we are building javadoc, unfortunately I think it should be in the build for PRs just to make sure it's working. I'm really frustrated with all the problems though. Having said that I think javadoc should NOT be in the full (reference) build for Fix Latest, it's just too big.

quickfixj-base/src/main/java/quickfix/Dictionary.java
david-gibbs-ig: The reference is to a class that depends on the Dictionary, Dictionary does not depend on SessionSettings, so it's the wrong way round. That class is in quickfixj-core so the @see tag causes Javadoc warnings or errors, core depends on base not visa versa. I did not want to be aggressive and lose the information that the author intended so I just took away the "@" so that no errors result.

quickfixj-base/src/main/java/quickfix/MessageFactory.java
david-gibbs-ig: That class is in quickfixj-core so the @see tag causes Javadoc warnings or errors, core depends on base not visa versa. I did not want to be aggressive and lose the information that the author intended so I just took away the "@" so that no javadoc errors result.

quickfixj-base/src/main/java/quickfix/MessageFactory.java
david-gibbs-ig: The @link tag causes Javadoc warnings or errors, as the base does not depend on the message jars. I did not want to be aggressive and lose the information that the author intended so I just took away the "@link" so that no errors result.

quickfixj-base/src/main/java/quickfix/MessageUtils.java
david-gibbs-ig: Yes , confirmed

quickfixj-base/src/main/java/quickfix/UtcDateOnlyField.java
david-gibbs-ig: The doc is a bit confusing isn't it/
https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#BASIC_ISO_DATE

Here is the code from https://github.com/frohoff/jdk8u-dev-jdk/blob/master/src/share/classes/java/time/format/DateTimeFormatter.java

```
    public static final DateTimeFormatter BASIC_ISO_DATE;
    static {
        BASIC_ISO_DATE = new DateTimeFormatterBuilder()
                .parseCaseInsensitive()
                .appendValue(YEAR, 4)
                .appendValue(MONTH_OF_YEAR, 2)
                .appendValue(DAY_OF_MONTH, 2)
                .optionalStart()
                .appendOffset("+HHMMss", "Z")
                .toFormatter(ResolverStyle.STRICT, IsoChronology.INSTANCE);
    }
```
So it looks like the input could have TZ offset and then the date would not be UTC.  I suppose a custom formatter is required.

.github/workflows/maven.yml
```diff
@@ -29,4 +29,4 @@ jobs:
       - name: Test with Maven
         env:
           MAVEN_OPTS: "-Xms3g -Xmx5g -Djdk.xml.xpathExprGrpLimit=500 -Djdk.xml.xpathExprOpLimit=500"
-        run: mvn install  -B -V -Pminimal-fix-latest -D"java.util.logging.config.file"="logging.properties" -D"http.keepAlive"="false" -D"maven.wagon.http.pool"="false" -D"maven.wagon.httpconnectionManager.ttlSeconds"="120"
+        run: mvn test -B -V -Pminimal-fix-latest -D"java.util.logging.config.file"="logging.properties" -D"http.keepAlive"="false" -D"maven.wagon.http.pool"="false" -D"maven.wagon.httpconnectionManager.ttlSeconds"="120"
```
chrjohn, line 32: `Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:2.8:unpack (unpack-dependencies) on project quickfixj-messages-fixt11: Artifact has not been packaged yet. When used on reactor artifact, unpack should be executed after packaging: see MDEP-98`

Never seen that one before.

david-gibbs-ig: OK I see it now, It's because of unpacking the QFJ orchestration to be used in code generation.

This takes us into a philosophical territory. I think it's better to use packaged artefacts than to copy from relative locations in a multi-module build. This approach here is also didactic as a QFJ user customising messages would import and unpack the dependency in just the same way ( without having to check out an build the whole QFJ project ). If we do things the same way as we may expect a QFJ user to do them, this should help with maintainability.

If it is OK with you I think it's good to use mvn install, there are not any negative side effects I can think of as the build starts with a fresh local maven repo.

```
			<plugin>
				<groupId>org.apache.maven.plugins</groupId>
				<artifactId>maven-dependency-plugin</artifactId>
				<executions>
					<execution>
						<id>unpack-dependencies</id>
						<phase>initialize</phase>
						<goals>
							<goal>unpack</goal>
						</goals>
						<configuration>
							<artifactItems>
								<artifactItem>
									<groupId>org.quickfixj</groupId>
									<artifactId>quickfixj-orchestration</artifactId>
									<version>${project.version}</version>
									<includes>**/FIX*.xml,**/Orchestra*.xml</includes>
									<outputDirectory>${project.build.directory}/dependency/repository</outputDirectory>
								</artifactItem>
							</artifactItems>
						</configuration>
					</execution>
				</executions>
			</plugin>
```

The following shows that the build starts with a fresh local maven repo, artefacts are not left lying around from earlier builds.

[INFO] --------------< org.quickfixj:quickfixj-messages-fixt11 >---------------
[INFO] Building QuickFIX/J Message classes for FIXT 1.1 3.0.0-SNAPSHOT   [8/25]
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] Downloading from central: https://repo.maven.apache.org/maven2/org/apache/maven/plugins/maven-antrun-plugin/1.8/maven-antrun-plugin-1.8.pom
[INFO] Downloaded from central: https://repo.maven.apache.org/maven2/org/apache/maven/plugins/maven-antrun-plugin/1.8/maven-antrun-plugin-1.8.pom (3.3 kB at 53 kB/s)
[INFO] Downloading from central: https://repo.maven.apache.org/maven2/org/apache/maven/plugins/maven-plugins/27/maven-plugins-27.pom
[INFO] Downloaded from central: https://repo.maven.apache.org/maven2/org/apache/maven/plugins/maven-plugins/27/maven-plugins-27.pom (11 kB at 232 kB/s)
[INFO] Downloading from central: https://repo.maven.apache.org/maven2/org/apache/maven/plugins/maven-antrun-plugin/1.8/maven-antrun-plugin-1.8.jar
[INFO] Downloaded from central: https://repo.maven.apache.org/maven2/org/apache/maven/plugins/maven-antrun-plugin/1.8/maven-antrun-plugin-1.8.jar (36 kB at 692 kB/s)

david-gibbs-ig: mvn install should run integration tests, have I changed a default somewhere ?https://stackoverflow.com/questions/1399240/how-do-i-get-my-maven-integration-tests-to-run

david-gibbs-ig: base has : [INFO] Tests run: 153, Failures: 0, Errors: 0, Skipped: 0 
I think the 614 tests are the integration tests...

david-gibbs-ig: @chrjohn that's more like it, only problem is, there are test failures now.
2023-03-21T14:36:09.5779520Z [INFO] 
2023-03-21T14:36:09.5781950Z [ERROR] Tests run: 1422, Failures: 2, Errors: 0, Skipped: 0

david-gibbs-ig: Sure - is generated base class for message cracker wrong ?
Yes there is a lot of logging that we can do without once we are confident in the overall structure.

david-gibbs-ig: @chrjohn 
One thing I have noticed is a large delay in processing compared to some other tests, perhaps this is related to why there is a Heartbeat.

INFO: load test: F:\dev\git\quickfixj\quickfixj-core\target\test-classes\quickfix\test\acceptance\definitions\server\fix50\2r_UnregisteredMsgType.def
Mar 21, 2023 2:12:01 PM quickfix.test.acceptance.PrintComment run
INFO: # Message type is not supported by the application
Mar 21, 2023 **2:12:01** PM quickfix.mina.acceptor.AcceptorIoHandler sessionCreated
INFO: MINA session created: local=/127.0.0.1:9887, class org.apache.mina.transport.socket.nio.NioSocketSession, remote=/127.0.0.1:64223
Mar 21, 2023 **2:12:07** PM quickfix.test.acceptance.AcceptanceTestSuite$AcceptanceTest run
SEVERE: test F:\dev\git\quickfixj\quickfixj-core\target\test-classes\quickfix\test\acceptance\definitions\server\fix50\2r_UnregisteredMsgType.def failed with message: wrong msg type expected:<[j]> but was:<[0]>
Mar 21, 2023 2:12:07 PM quickfix.test.acceptance.AcceptanceTestSuite$AcceptanceTest load
INFO: load test: F:\dev\git\quickfixj\quickfixj-core\target\test-classes\quickfix\test\acceptance\definitions\server\fix50\2t_FirstThreeFieldsOutOfOrder.def
Mar 21, 2023 2:12:07 PM quickfix.test.acceptance.PrintComment run

chrjohn: The delay is because the expected message has not been received. Then the connection will be open for some seconds and heartbeats are sent/ received. Eventually there will be a timeout.

david-gibbs-ig: > The delay is because the expected message has not been received. Then the connection will be open for some seconds and heartbeats are sent/ received. Eventually there will be a timeout.

Is it received but an exception swallowed ? Other FIX50 cases seem to receive messages...

david-gibbs-ig: Sounds like the best approach , I already spent 40 minutes gazing and guessing about the code ;-)

chrjohn: Hi @david-gibbs-ig , I have added some temporary logging into `quickfix.MessageCracker` and it indeed looks like the thrown Exception is swallowed. But don't know where ATM.

https://github.com/david-gibbs-ig/quickfixj/blob/3a4d5c2bb982be46441fe540dfeb536543bfd2b3/quickfixj-core/src/main/java/quickfix/MessageCracker.java#L167

david-gibbs-ig:
@chrjohn I saw that you were working on this issue so I thought I would take a look. 2r_UnregisteredMsgType.def send this execution report, but I can't see "55=WLRI" in the logs output by the build. I wonder if the Exec Rpt message is not being **sent**. An exception is in InitiateMessageStep.java should be caught in AcceptanceTestSuite.java (from line 82) but I'm going to try adding some logging to the Acceptance test to see what I can discover about how the tests work.
`I8=FIXT.1.135=834=249=TW52=<TIME>56=ISLD37=id17=id150=039=055=WLRI54=1151=10014=06=0.0`

chrjohn:
Hi @david-gibbs-ig 

I think the problem with the logging is the following:

```
2023-04-19T16:46:14.6862310Z SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
2023-04-19T16:46:14.6878156Z SLF4J: Defaulting to no-operation (NOP) logger implementation
2023-04-19T16:46:14.6878965Z SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details. 
```

I removed the `slf4j-jdk14` binding some commits ago to fix a problem with the bundle plugin. See https://github.com/quickfix-j/quickfixj/pull/615
 I think we need to re-add another binding as `test` dependency. Maybe we should try the `simple` binding/provider. https://www.slf4j.org/manual.html#swapping

david-gibbs-ig:
@chrjohn 

```
NioProcessor-1345] INFO quickfixj.msg.incoming - FIXT.1.1:ISLD->TW: 8=FIXT.1.19=10435=834=249=TW52=20230507-19:00:28.48456=ISLD37=id17=id150=039=055=WLRI54=1151=10014=06=0.010=245
        XXXX 
java.lang.Exception
	at quickfix.MessageCracker.onMessage(MessageCracker.java:163)
	at quickfix.MessageCracker.crack(MessageCracker.java:140)
	at quickfix.test.acceptance.ATApplication.fromApp(ATApplication.java:89)
	at quickfix.Session.fromCallback(Session.java:1915)
	at quickfix.Session.verify(Session.java:1859)
	at quickfix.Session.verify(Session.java:1930)
	at quickfix.Session.next(Session.java:1115)
	at quickfix.Session.next(Session.java:1272)
	at quickfix.mina.SingleThreadedEventHandlingStrategy$SessionMessageEvent.processMessage(SingleThreadedEventHandlingStrategy.java:163)
	at quickfix.mina.SingleThreadedEventHandlingStrategy.block(SingleThreadedEventHandlingStrategy.java:113)
	at quickfix.mina.SingleThreadedEventHandlingStrategy.lambda$blockInThread$1(SingleThreadedEventHandlingStrategy.java:145)
	at quickfix.mina.SingleThreadedEventHandlingStrategy$ThreadAdapter$RunnableWrapper.run(SingleThreadedEventHandlingStrategy.java:270)
	at java.base/java.lang.Thread.run(Thread.java:829)
[QFJ Message Processor] ERROR quickfixj.errorEvent - FIXT.1.1:ISLD->TW: Reject sent for message 2: Unsupported Message Type
[QFJ Message Processor] ERROR quickfixj.errorEvent - FIXT.1.1:ISLD->TW: Error accessing message fields
quickfix.FieldNotFound: Field was not found in message, field=35
	at quickfix.FieldMap.getField(FieldMap.java:228)
	at quickfix.FieldMap.getString(FieldMap.java:242)
	at quickfix.Session.sendRaw(Session.java:2631)
	at quickfix.Session.generateBusinessReject(Session.java:1748)
	at quickfix.Session.next(Session.java:1178)
	at quickfix.Session.next(Session.java:1272)
	at quickfix.mina.SingleThreadedEventHandlingStrategy$SessionMessageEvent.processMessage(SingleThreadedEventHandlingStrategy.java:163)
	at quickfix.mina.SingleThreadedEventHandlingStrategy.block(SingleThreadedEventHandlingStrategy.java:113)
	at quickfix.mina.SingleThreadedEventHandlingStrategy.lambda$blockInThread$1(SingleThreadedEventHandlingStrategy.java:145)
	at quickfix.mina.SingleThreadedEventHandlingStrategy$ThreadAdapter$RunnableWrapper.run(SingleThreadedEventHandlingStrategy.java:270)
	at java.base/java.lang.Thread.run(Thread.java:829)
```

chrjohn:
@david-gibbs-ig 
Great, progress. :) What made that exception message appear? I cannot see it in the build logs.
I always thought that it must be something version-specific since only FIX5.0 did not work. Maybe the `MessageFactory`, although I thought I already diffed that one. Will try to take a look later today or tomorrow.
Thank you. :)

david-gibbs-ig:
I changed the logging for test as you had suggested, to slf4j-simple, removed slf4j-jdk which was pulling in java.util.logging. I comitted this change. I think think this might be due to an option on the code generation so I'll have a look at that.

david-gibbs-ig:
The constructed Business Message Reject looks pretty good except for lack of tag 35 ! I'll add some more debugging :-) At least I don't have to regenerate and recompile the messages every time ;-)

```
[QFJ Message Processor] ERROR quickfixj.errorEvent - FIXT.1.1:ISLD->TW: Reject sent for message number 2: Unsupported Message Type 
Message [8=FIXT.1.19=10435=834=249=TW52=20230508-10:23:59.41856=ISLD6=0.014=017=id37=id39=054=155=WLRI150=0151=10010=243]
Reject [8=FIXT.1.19=8934=249=ISLD52=20230508-10:23:59.44356=TW45=258=Unsupported Message Type372=8380=310=034]
```

david-gibbs-ig:
@chrjohn It was because FIXT1.1 MessageFactory is being used to try to create a BusinessMessageReject. In the new generated code BusinessMessageReject is present in the relevant Application Message Factory. quickfix.fix50.MessageFactory, for example.
I adapted the code in Session.java https://github.com/quickfix-j/quickfixj/pull/460/commits/006c79537cfb34c8aa98832668e80febb5cb504d
There now seems to be a problem with the logging test 

```
[ERROR]   SLF4JLogTest.testLog:63->assertMessageLogged:191 expected:<1> but was:<0>
[ERROR]   SLF4JLogTest.testLogHeartbeatFiltering:114->assertMessageLogged:191 expected:<1> but was:<0>
```

david-gibbs-ig:
@chrjohn , could you please trigger another build, or some of those individual builds ? I think the issue with the Business Message Reject acceptance test is resolved. I had to change the logging back to pass the SLF4JLogTest.

pom.xml
```diff
@@ -51,15 +51,14 @@
...
 
-	<prerequisites>
-		<maven>3.5.0</maven>
-	</prerequisites>
-
 	<modules>
 		<module>quickfixj-codegenerator</module>
 		<module>quickfixj-dictgenerator</module>
-		<module>quickfixj-core</module>
+		<module>class-pruner-maven-plugin</module>
```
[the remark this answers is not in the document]
david-gibbs-ig: renamed

david-gibbs-ig:
@chrjohn

The only thing that is failing in the recent builds is due to Windows page file size.  This is I feel a sort of github problem.  Other people have had this issue and I have found one work around as below. I think that one could test the os.context and run the page file step. What do you think ? It seems odd that github don't have a good, standard, solution for this. 

Error:  Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.5.0:jar (attach-javadocs) on project quickfixj-messages-fix50sp2: MavenReportException: Error while generating Javadoc: 
Error:  Exit code: 1
Error:  OpenJDK 64-Bit Server VM warning: INFO: os::commit_memory(0x00000006d6900000, 1053818880, 0) failed; error='The paging file is too small for this operation to complete' (DOS error/errno=1455)
Error:  Command line was: cmd.exe /X /C "C:\hostedtoolcache\windows\Java_Temurin-Hotspot_jdk\11.0.19-7\x64\bin\javadoc.exe -J-Xmx6144m @options @argfile"
Error:  
Error:  Refer to the generated Javadoc files in 'D:\a\quickfixj\quickfixj\quickfixj-messages\quickfixj-messages-fix50sp2\target\apidocs' dir.

https://stackoverflow.com/questions/57946173/github-actions-run-step-on-specific-os

https://github.com/al-cheb/configure-pagefile-action
https://github.com/LinuxForHealth/FHIR/blob/fe258203b6ed445d1c3ec87de9dfe529d52c0a8b/.github/workflows/windows.yml#L24

chrjohn:
@david-gibbs-ig

W.r.t. Windows page file size. Yes, please add the workaround. Thank you.

W.r.t. the failing test with the BusinessMessageReject. Thanks for the change. I cannot understand why it worked before to be honest. Are you suggesting that before your changes the BMR was in the wrong message factory?

david-gibbs-ig:
> W.r.t. the failing test with the BusinessMessageReject. Thanks for the change. I cannot understand why it worked before to be honest. Are you suggesting that before your changes the BMR was in the wrong message factory?

BMR is not in the old FIXT11.xml so I'm not certain, probably have to look at this again to make sure there is not a different cause. For example in the AT I found assumptions about the length of FIX version string, but this test was for FIX50 not for FIXLatest.

I am adding acceptance tests for FIX Latest but still getting some failures.

david-gibbs-ig:
@chrjohn  I think this code in quickfixj-core/FIXMessageDecoder.java looks suspect - PatternMatcher is case-sensitive by default isn't it ?

image

chrjohn:
> @chrjohn I think this code in quickfixj-core/FIXMessageDecoder.java looks suspect - PatternMatcher is case-sensitive by default isn't it ? I'm not sure why the 't' in "FIXt" is lower case.

Don't confuse it with the Java pattern matcher, it is https://github.com/quickfix-j/quickfixj/blob/dce91e62c7525b84448b974b796a1fac436015ae/quickfixj-core/src/main/java/quickfix/mina/message/PatternMatcher.java#L15

Stumbled about that myself some while ago.

david-gibbs-ig:
> > @chrjohn I think this code in quickfixj-core/FIXMessageDecoder.java looks suspect - PatternMatcher is case-sensitive by default isn't it ? I'm not sure why the 't' in "FIXt" is lower case.
> 
> Don't confuse it with the Java pattern matcher, it is
> 
> https://github.com/quickfix-j/quickfixj/blob/dce91e62c7525b84448b974b796a1fac436015ae/quickfixj-core/src/main/java/quickfix/mina/message/PatternMatcher.java#L15
> 
> Stumbled about that myself some while ago.

Ah thanks. I am grasping at straws. I am getting Out of Memory exceptions running the Acceptance Tests set up for FIX Latest. I know I ran the generated code before - but I think I generated the classes as FIX50SP2 rather than FIXLatest - so that we did not change our Rules of Engagement. Now I am wondering... based on what I think I recall,  I was trying to work through the code to find where "FIXLatest" makes a difference to QFJ.

david-gibbs-ig:
@chrjohn  I have been unable to solve the Out of Memory errors on FIXLatest acceptance tests, it looks like a leak to me - as the test that consistently were failing for OOM passed when I removed all the other tests (The test where OOM was showing up was fixLatest/LastProcessedMsgSeqNum.def) .  I tried to see if it was related to caching the DataDictionary but I was not seeing the print statements I added and expected to see. I'm a bit baffled so I have committed the changes and am asking for help. Note I did spot and "fix" a couple of issues in MessageUtils.java and ATServer.java to get the tests to run, please  see https://github.com/quickfix-j/quickfixj/pull/460/commits/51223508b207be9d61213070e28228a0cbb1219e

chrjohn:
> @chrjohn I have been unable to solve the Out of Memory errors on FIXLatest acceptance tests, it looks like a leak to me - as the test that consistently were failing for OOM passed when I removed all the other tests

Does not have to be a leak, I think it just takes up more memory due to more classes loaded. You mentioned that there was a lot of FIXLatest classes compared to the other FIX versions. Will try to increase heap memory.

chrjohn:
@david-gibbs-ig  memory seems ok now but failures in acceptance tests...

david-gibbs-ig:
> @david-gibbs-ig memory seems ok now but failures in acceptance tests...

Thank you, I think I have corrected acceptance tests, running builds now locally.

david-gibbs-ig:
@chrjohn  Looking much better, still seeing the usual problems with connecting to mvn central. FIXLatest Acceptance Tests  are passing.

david-gibbs-ig:
@crjohn Can you please re-run JDK8 on Ubuntu ? The failure seems to be the GitHub or maven repo reliability issue.

"Error:  Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:2.8:unpack (unpack-dependencies) on project quickfixj-orchestration: Execution unpack-dependencies of goal org.apache.maven.plugins:maven-dependency-plugin:2.8:unpack failed: Plugin org.apache.maven.plugins:maven-dependency-plugin:2.8 or one of its dependencies could not be resolved: Failed to collect dependencies at org.apache.maven.plugins:maven-dependency-plugin:jar:2.8 -> commons-lang:commons-lang:jar:2.6: Failed to read artifact descriptor for commons-lang:commons-lang:jar:2.6: Could not transfer artifact org.apache.commons:commons-parent:pom:17 from/to central (https://repo.maven.apache.org/maven2): transfer failed for https://repo.maven.apache.org/maven2/org/apache/commons/commons-parent/17/commons-parent-17.pom: Connect to repo.maven.apache.org:443 [repo.maven.apache.org/151.101.40.215, repo.maven.apache.org/2a04:4e42:a:0:0:0:0:215] failed: Network is unreachable (connect failed) -> [Help 1]"

david-gibbs-ig:
@chrjohn I did not understand how to get the Windows page file size work-around to be succeed so I reverted the commits. The head revision should be working now with acceptance tests for FIX Latest, just not for Windows builds.

Error:  Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.5.0:jar (attach-javadocs) on project quickfixj-messages-fix50sp2: MavenReportException: Error while generating Javadoc: 
Error:  Exit code: 1
Error:  OpenJDK 64-Bit Server VM warning: INFO: os::commit_memory(0x00000006a4000000, 1013972992, 0) failed; error='The paging file is too small for this operation to complete' (DOS error/errno=1455)
Error:  Command line was: cmd.exe /X /C "C:\hostedtoolcache\windows\Java_Temurin-Hotspot_jdk\11.0.19-7\x64\bin\javadoc.exe -J-Xmx6144m @options @argfile"

chrjohn:
@david-gibbs-ig 
I needed to add the pagefile action to a list of allowed actions. Could you please try again but with al-cheb/configure-pagefile-action@v1.3 which is the most recent version. Thank you.

david-gibbs-ig:
@chrjohn 
I  checked the reverted commit and I did not specify D. The output shows it has set D as the root for both versions. It seems the javadoc is taking a very long time. In these builds we don't skip the javadoc. I don't know why the performance of the Windows build is so execrable. It was part way through creating javadoc for FIX4.4. Perhaps it would help to set the javadoc build to "silent" if that's possible, I think I might have tried that before without success. 

image

image

david-gibbs-ig:
@chrjohn  I'll try it with the javadoc quiet option and if that does not help I will try to skip javadoc as an experiment.

david-gibbs-ig:
@chrjohn  Horrendous, I think it will be worth re-running those jobs to see if "quiet" has any effect.

david-gibbs-ig:
@chrjohn 
There seem to be endemic issues with github windows runners, there are more reports than the ones I find below. I've always really liked github but one has to wonder if it is fit for purpose now.

https://github.com/actions/runner-images/issues/5166
https://github.com/actions/runner-images/issues/7320
https://github.com/getsentry/sentry-dotnet/issues/1568

chrjohn:
@david-gibbs-ig I think we need more logging. It hangs in the acceptance tests for hours: 

```
2023-05-31T10:21:59.6316951Z [INFO] Running quickfix.test.acceptance.AcceptanceTestSuite
2023-05-31T16:06:19.8368167Z # build is aborted here
```

david-gibbs-ig:
> @david-gibbs-ig I think we need more logging. It hangs in the acceptance tests for hours:
> 
> ```
> 2023-05-31T10:21:59.6316951Z [INFO] Running quickfix.test.acceptance.AcceptanceTestSuite
> 2023-05-31T16:06:19.8368167Z # build is aborted here
> ```

I've been running it in WSL, I'll try running it in Windows. Have you tried running it locally?

chrjohn:
> I've been running it in WSL, I'll try running it in Windows. Have you tried running it locally?

I have my development environment completely in WSL, so cannot run it under native Windows. I'm wondering why the acceptance tests do not have any logging. Thought we turned it on for analysis some commits ago but cannot find where it was disabled again.

david-gibbs-ig:
> > I've been running it in WSL, I'll try running it in Windows. Have you tried running it locally?
> 
> I have my development environment completely in WSL, so cannot run it under native Windows. I'm wondering why the acceptance tests do not have any logging. Thought we turned it on for analysis some commits ago but cannot find where it was disabled again.

I put the logging back the way it had been, I will commit back the simple logging. I ran this in Windows and at first thought I could see the failure, but this was because there were some old zombie builds interfering with the AT. I killed the zombie processes and reran the build which worked except for the SLFJLogTest which relied on some non simple logging configuration.

chrjohn:
Maybe we could put back the logger to jdk14 instead of simple. Probably that makes the slf4j test pass. I think the problem why I changed the logger was https://github.com/quickfix-j/quickfixj/pull/615/files
Will check outcome of acceptance tests later.

chrjohn:
BTW I thought about using cache for maven https://github.com/actions/setup-java#caching-maven-dependencies

Given the many recent failures when downloading we might give it a try.

david-gibbs-ig:
@chrjohn I think those windows errors look like the ones I saw when I had zombie builds still running in background (suspected). I will have to check and see if I changed something about forking junit tests or similar.

david-gibbs-ig:
> @chrjohn I think those windows errors look like the ones I saw when I had zombie builds still running in background (suspected). I will have to check and see if I changed something about forking junit tests or similar.

I do not see a change like that, unless some default may have changed with a new version.

david-gibbs-ig:
> Maybe we could put back the logger to jdk14 instead of simple. Probably that makes the slf4j test pass. I think the problem why I changed the logger was https://github.com/quickfix-j/quickfixj/pull/615/files Will check outcome of acceptance tests later.

@chrjohn  Yes we can put that config and the test back easily but at least now you can see the AT logs. I have added the maven cache, hope it works. in the last run the windows build failed for mvn connectivity reasons before I could see if running one build at a time made a difference.

chrjohn:
https://github.com/david-gibbs-ig/quickfixj/blob/6a7915061ca501a1544871eb8fd726432003d778/.github/workflows/maven.yml#L44

What does this do?
But should runners really influence each other?

chrjohn:
Initial error in acceptance tests seems to be this:

```
2023-06-04T15:13:19.4217000Z [INFO] Running quickfix.test.acceptance.AcceptanceTestSuite
2023-06-04T15:13:19.4249959Z [main] INFO quickfix.test.acceptance.ATServer - creating sessions for []
2023-06-04T15:13:19.4251295Z [pool-16-thread-1] ERROR quickfix.test.acceptance.ATServer - error in AT server
2023-06-04T15:13:19.4254736Z quickfix.ConfigError: No acceptor sessions found in settings.
2023-06-04T15:13:19.4255347Z 	at quickfix.mina.acceptor.AbstractSocketAcceptor.createSessions(AbstractSocketAcceptor.java:259)
2023-06-04T15:13:19.4259979Z 	at quickfix.mina.acceptor.AbstractSocketAcceptor.startAcceptingConnections(AbstractSocketAcceptor.java:98)
2023-06-04T15:13:19.4260800Z 	at quickfix.SocketAcceptor.initialize(SocketAcceptor.java:110)
2023-06-04T15:13:19.4261269Z 	at quickfix.SocketAcceptor.start(SocketAcceptor.java:103)
2023-06-04T15:13:19.4261731Z 	at quickfix.test.acceptance.ATServer.run(ATServer.java:193)
2023-06-04T15:13:19.4262309Z 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
2023-06-04T15:13:19.4262879Z 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
2023-06-04T15:13:19.4263319Z 	at java.lang.Thread.run(Thread.java:750)
```

That's probably why all subsequent tests fail at session creation and run into a timeout.
But don't know why this does not work under Windows now. It worked before. ;)

david-gibbs-ig:
@chrjohn The parsing to define the test locations is certainly the problem. I'll work on it this morning

david-gibbs-ig:
@crjohn Just one build to rerun... :-)

david-gibbs-ig:
> > Maybe we could put back the logger to jdk14 instead of simple. Probably that makes the slf4j test pass. I think the problem why I changed the logger was https://github.com/quickfix-j/quickfixj/pull/615/files Will check outcome of acceptance tests later.
> 
> @chrjohn Yes we can put that config and the test back easily but at least now you can see the AT logs. I have added the maven cache, hope it works. in the last run the windows build failed for mvn connectivity reasons before I could see if running one build at a time made a difference.

@chrjohn Let me know if you want me to put the logging back the way it was ( jdk14 ) originally along with the slf4j test. We lose visibility of some logging in the AT but we have solved the immediate problem now. I think that fixing up the logging is a separate problem.

chrjohn:
In an effort to make ATServer fail faster I completely screwed something up. Will revert last commit.

david-gibbs-ig:
> We now also have `FileUtilTest.testURLLocation:66 Resource not found` all over the place. Seems to be some spurious errors since the test first checks if internet is accessible. Will leave that to later.

Thanks for investigating that, hopefully it is related to these transient but frequent GitHub connectivity issues. I was wondering if these connectivity issues may be related to ddos/bot protection but sometimes even the first job to run seems to fail.

david-gibbs-ig:
@chrjohn Can you please restart the failed checks so we can confirm the issue is transient ??

david-gibbs-ig:
@chrjohn - the failing steps are all Java 8 so that's something to consider.

chrjohn:
> @chrjohn - the failing steps are all Java 8 so that's something to consider.

That's strange. I don't recall that test failing on master.

chrjohn:
> > @chrjohn - the failing steps are all Java 8 so that's something to consider.
> 
> That's strange. I don't recall that test failing on master.

I was wrong: https://github.com/quickfix-j/quickfixj/actions/runs/5207778432

Need to check since when this failed. Probably caused by a Java update?

david-gibbs-ig:
@chrjohn It is : "Java version: 1.8.0_372, vendor: Temurin, runtime: /usr/lib/jvm/temurin-8-jdk-amd64/jre"

chrjohn:
But it is the same version as two days ago and that worked?!
https://github.com/quickfix-j/quickfixj/actions/runs/5184991632/jobs/9344421765

david-gibbs-ig:
> But it is the same version as two days ago and that worked?! https://github.com/quickfix-j/quickfixj/actions/runs/5184991632/jobs/9344421765
> 
> ```
> 14 Resolved Java 8.0.372+7 from tool-cache 
> 15 Setting Java 8.0.372+7 as the default
> ```

Maven and Ubuntu version also the same afaik.

david-gibbs-ig:
@chrjohn 
Exception opening stream from URL
java.io.IOException: Server returned HTTP response code: 403 for URL: http://www.quickfixj.org/

I don't know why yet though.

chrjohn:
@david-gibbs-ig great, looks like it's fixed. How did you find out what the problem was?

david-gibbs-ig:
> @david-gibbs-ig great, looks like it's fixed. How did you find out what the problem was?

I added in some exception logging to see that the problem was 403, then looked for a Java 8 solution. I think we should look at the OpenJDK implementation of URL in versions after Java8 and set the same agent header. I was in a hurry to push this and see if it worked.

david-gibbs-ig:
@chrjohn 
> > @david-gibbs-ig great, looks like it's fixed. How did you find out what the problem was?
> 
> I added in some exception logging to see that the problem was 403, then looked for a Java 8 solution. I think we should look at the OpenJDK implementation of URL in versions after Java8 and set the same agent header. I was in a hurry to push this and see if it worked.

I made an update to set User-Agent as follows
`httpURLConnection.setRequestProperty("User-Agent", "Java-QuickFIXJ-FileUtil");`

The default Java ones are specific to the Java version.

david-gibbs-ig:
@crjohn Good to see that the builds all passed.  Please et me know if you want me to put the logging back the way it was ( jdk14 ) originally along with the slf4j test. We lose visibility of some logging in the AT but we have solved the immediate problems now. I think that fixing up the AT logging can be treated as an independent issue.

chrjohn:
@david-gibbs-ig yes, please change the logging back to jdk14. Thank you.
Slowly further working through your changes... :)

chrjohn:
I have errors like this when building with `full-fix-latest`.

`ComplexEventStartDate.java:[22,16] error: incompatible types: String cannot be converted to LocalDate`

Any idea what might be wrong? Was this related to the changes you did here: https://github.com/quickfix-j/quickfixj-orchestra/pull/4 ?

david-gibbs-ig:
> I have errors like this when building with `full-fix-latest`.
> 
> `ComplexEventStartDate.java:[22,16] error: incompatible types: String cannot be converted to LocalDate`
> 
> Any idea what might be wrong? Was this related to the changes you did here: [quickfix-j/quickfixj-orchestra#4](https://github.com/quickfix-j/quickfixj-orchestra/pull/4) ?

Must be, or a closely related issue. Can you send me the log or is it on github ?

quickfixj-messages/quickfixj-messages-all/pom.xml
```diff
@@ -10,23 +10,137 @@
...
+							<outputDirectory>${project.build.directory}/generated-sources</outputDirectory>
+							<excludeSession>true</excludeSession><!-- fixt11 has already been generated -->
+							<generateFixt11Package>false</generateFixt11Package>
+						</configuration>
+					</plugin>
+				</plugins>
+			</build>
+		</profile>
+		<profile>
+			<id>full-fix-latest</id>
```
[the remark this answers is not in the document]
david-gibbs-ig: Do you mean "minimal-fix-latest" should be the default? Yes I think that should be the default, building "full" would be to look for new issues in new releases of FIX Orchestra, a specialized case.

chrjohn:
Moving `FileUtilTest` caused the build to fail. I'll look into it.

david-gibbs-ig:
> Moving `FileUtilTest` caused the build to fail. I'll look into it.

Thanks so much for working on this. I wish I could engage more right now.  I will try to catch up with your work. FileUtilTest rings a bell :-)

chrjohn:
> Thanks so much for working on this. I wish I could engage more right now.

Don't worry, you have done very much already. 👍 Just get back when you have the time.

chrjohn:
> I have errors like this when building with `full-fix-latest`.
> 
> `ComplexEventStartDate.java:[22,16] error: incompatible types: String cannot be converted to LocalDate`
> 
> Any idea what might be wrong? Was this related to the changes you did here: [quickfix-j/quickfixj-orchestra#4](https://github.com/quickfix-j/quickfixj-orchestra/pull/4) ?

@david-gibbs-ig , I think I know where this error comes from. All fields that have a type of `UTCDateOnly` are affected by this. The reason seems to be these lines of code that I think were added by you. Do you still recall why we need an extra constructor with `String`?

https://github.com/quickfix-j/quickfixj-orchestra/blob/1845c4306cb78c786ea3a1c38408eeae76764d6e/quickfixj-from-fix-orchestra-repository/quickfixj-from-fix-orchestra-generator/src/main/java/org/quickfixj/orchestra/CodeGeneratorJ.java#L858-L860

You did some (related ?) changes with this PR: https://github.com/quickfix-j/quickfixj-orchestra/pull/4
I agree that `LocalMktTime` and `LocalMktDate` shouldn't be handled by `LocalDate/Time` since the time zone might not be known. But IMHO `UTCDateOnly` could be handled by a `LocalDate` with time zone UTC.

What do you think?

david-gibbs-ig:
> > I have errors like this when building with `full-fix-latest`.
> > `ComplexEventStartDate.java:[22,16] error: incompatible types: String cannot be converted to LocalDate`
> > Any idea what might be wrong? Was this related to the changes you did here: [quickfix-j/quickfixj-orchestra#4](https://github.com/quickfix-j/quickfixj-orchestra/pull/4) ?
> 
> @david-gibbs-ig , I think I know where this error comes from. All fields that have a type of `UTCDateOnly` are affected by this. The reason seems to be these lines of code that I think were added by you. Do you still recall why we need an extra constructor with `String`?
> 
> https://github.com/quickfix-j/quickfixj-orchestra/blob/1845c4306cb78c786ea3a1c38408eeae76764d6e/quickfixj-from-fix-orchestra-repository/quickfixj-from-fix-orchestra-generator/src/main/java/org/quickfixj/orchestra/CodeGeneratorJ.java#L858-L860
> 
> You did some (related ?) changes with this PR: [quickfix-j/quickfixj-orchestra#4](https://github.com/quickfix-j/quickfixj-orchestra/pull/4) I agree that `LocalMktTime` and `LocalMktDate` shouldn't be handled by `LocalDate/Time` since the time zone might not be known. But IMHO `UTCDateOnly` could be handled by a `LocalDate` with time zone UTC.
> 
> What do you think?

I did not remove the extra ctor for `UTCDateOnly`  in https://github.com/quickfix-j/quickfixj-orchestra/pull/4. I'm testing a new version locally.

david-gibbs-ig:
@chrjohn I have made the necessary change to the code generation in PR https://github.com/quickfix-j/quickfixj-orchestra/pull/6 .
I have built and tested this change locally with ComplexEventStartDate and can commit the changes when 1.0.2 of quickfixj-orchestra is published.

david-gibbs-ig:
@chrjohn  I'm very pleased to see the build checks green https://github.com/quickfix-j/quickfixj/pull/460/checks

quickfixj-core/pom.xml
```diff
@@ -1,490 +1,368 @@
...
+		<testResources>
+			<testResource>
+				<directory>src/test/resources</directory>
+			</testResource>
+			<testResource>
+				<directory>src/main/resources</directory>
+			</testResource>
+		</testResources>
 
-    <build>
```
chrjohn, line 122: @david-gibbs-ig do you recall why you removed this? Formerly we had the XML dictionaries packaged into the core JAR, which does not work without these lines. Not sure right now if we really need the dictionaries in the core JAR, though.

david-gibbs-ig: @chrjohn  Yes, that's right. In this design the dictionaries should not be in core now. The will only come from (message) jars that have been added to the classpath. These can be the default ones provided by QFJ or custom jars.

quickfixj-codegenerator/src/main/java/org/quickfixj/codegenerator/MessageCodeGenerator.java
```diff
@@ -305,10 +305,6 @@ private void generateCodeFile(Task task, Document document, Map<String, String>
             if (!task.isOverwrite()) {
                 return;
             }
-            if (outputFile.lastModified() > task.getSpecificationLastModified()) {
```
chrjohn, line 308: @david-gibbs-ig any idea why this was removed? Now the code is generated on each compile run, regardless if it was already created in a former run.

chrjohn:
@david-gibbs-ig merging in 3, 2, 1... 🥳

PR #462, opened 2022-02-02
Fixed `doclint:none` setting in `pom.xml`

raipc:


quickfixj-distribution/pom.xml
```diff
@@ -137,7 +137,7 @@
 							<goal>javadoc</goal>
 						</goals>
 						<configuration>
-							<additionalparam>${javadoc.opts}</additionalparam>
+							<additionalOptions>${javadoc.opts}</additionalOptions>
```
chrjohn, line 140: Quick question: why is this looking differently compared to the other changes? Can't we use the same approach here?

raipc: Just because it was initially implemented this way by using maven profiles. Since profiles are not needed anymore (Java 7 not supported), doclint parameter can be used explicitly

chrjohn: Thanks. 👍

PR #480, opened 2022-03-29
Add limit on messages pending resend

### LINKED ISSUE #477: SessionState#messageQueue is unbound
**Describe the bug**
The message queue in SessionState is unbound. This can lead to a vast quantity of messages being stored in memory which can lead to an OOM error.

In the worst case of a misbehaving FIX engine that doesn't handle resends, this queue grows without limit.

**To Reproduce**
On the counter-party, induce the side to skip a large volume of sequence numbers. Then start sending messages at a high rate.

The QuickFIXJ FIX engine will cache all the received messages in memory while it waits for the counter-party to process the resend. This will use a potentially infinite amount of memory (but could easily store thousands of messages on a typical high volume system).

**Expected behavior**
QFJ should only keep a (albeit size-able) limit of messages for a session. Any later messages should be dropped and then handled with a future re-send if necessary.

**Implementation notes**

Perhaps something like `MaxMessagesQueuedWhilePendingResend=2000`
We could allow `-1` for people who are truly happy with unlimited.

philipwhiuk:
Fixed #477

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/DefaultSessionFactory.java
+++ b/quickfixj-core/src/main/java/quickfix/DefaultSessionFactory.java
@@ -221,6 +221,7 @@ public class DefaultSessionFactory implements SessionFactory {
             final SessionSchedule sessionSchedule = sessionScheduleFactory.create(sessionID, settings);
 
             final List<StringField> logonTags = getLogonTags(settings, sessionID);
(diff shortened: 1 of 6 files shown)


chrjohn:
Hi @philipwhiuk , I've just merged #380 .
So it should be a matter of creating your own queue implementation with an `enqueue` method similar to this:

```
public void enqueue(int sequence, Message message) {
    if (backingMap.size() < maxMessagesQueuedWhilePendingResend || maxMessagesQueuedWhilePendingResend == -1) {
        backingMap.put(sequence, message);
    }
}

```

Do you think it is OK to close this PR then? Would you like to submit a PR with a queue implementation that is limited in size?
Thanks,
Chris.

PR #493, opened 2022-07-21
Repository to quickfixj

david-gibbs-ig:
Initial support for FIX Latest https://github.com/quickfix-j/quickfixj/issues/317 

This is related to PR https://github.com/quickfix-j/quickfixj/pull/460, it is an extraction of just the code generation and QFJ Dictionary generation based on the FTC FIX Orchestra Project.

This is a pre-requisite for the work to implement FIXLatest in QFJ.


pom.xml
```diff
@@ -69,27 +70,159 @@
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 		<jdkLevel>1.8</jdkLevel>
-		<slf4j.version>1.7.32</slf4j.version>
+		<slf4j.version>1.7.35</slf4j.version>
 		<junit.version>4.13.1</junit.version>
+		<mockito-core.version>3.12.4</mockito-core.version>
+		<hamcrest-all.version>1.3</hamcrest-all.version>
+		<junit.jupiter.version>5.6.1</junit.jupiter.version>
```
chrjohn, line 77: If we have Jupiter, do we still need JUnit 4.13.1?

chrjohn:
Hi @david-gibbs-ig , thanks for extracting this smaller PR.
Just a question: the code generation runs with the build and generates the dictionaries from the Orchestra file instead of from the old way (i.e. the already existing dictionaries that were bundled with QFJ).

pom.xml
david-gibbs-ig: I'll check that.

david-gibbs-ig:
Hi @chrjohn, this PR just builds the code and dictionary generation and does not yet integrate it into the QFJ build to build FIXLatest. I was just trying to break down the PRs into more than one as you had requested. This PR basically just makes these tools part of the QFJ project and the QFJ project doesn't use it yet. The markdown documents in the other full PR explain the approach a little more than I can here. When you look at the other PR you will see that for the QFJ distribution I tried to take an approach of using the existing code generation for FIX4.2 to FIX50sp2 , this *reduces* compatibility issues for current users that may be using the pre-built QFJ artefacts rather than custom builds. They can continue to use those org.quickfixj.fix42->fix50sp2 artefacts with minimal changes. FIXLatest is built with this new code generation and represents the break with the older versions. Someone building custom versions of FIX4.2 to FIX50sp2 can choose whether to use the new code generation or the old. One of the issues is that the original code generation just can't build the whole FIXLatest and it creates really huge classes. One of the purposes of the restructured project (in other PR) is to make the message artefacts runtime dependencies rather than compile time dependencies. This will make it easier to maintain custom ROEs in all versions of FIX as you will be able to just use the latest QFJ3.n with your compatible generated messages.  So the QFJ project will build FIXLatest but really for reference and to assure build/test compatibility when FIXLatest gets updated. Users will be encouraged to build their own message artefacts using the tools provided here. I have proven the whole stack in some projects that depend on the restructured QFJ and if we end up taking this approach then we can add some examples. P.S. My other PR *just* broke due to new constraints on xpath (I pushed some changes), I thought I had worked around that already. I'll check some of my other projects as my first attempts to work around the issues aren't correct. I have Covid at the moment so apologies if this is a little incoherent.

pom.xml
david-gibbs-ig: Jupiter is used in the code and dictionary generator tests but I recall JUnit compatibility problems with the maven plugin testing harness so the repository-to-qfj maven plugin module uses Junit 4.  Also there are many other places in the overall project that use junit4 which have not been changed - in this PR.
For ease of migration to junit5 there is a Junit Vintage runtime that requires Junit 4.12 or later.
"JUnit 5 = JUnit Platform + JUnit Jupiter + JUnit Vintage"
https://junit.org/junit5/docs/current/user-guide/
So I think it's good to have JUnit in the dependency management for now.

image

chrjohn:
I could use a few more eyes on this. Any comments from other regular committers? @philipwhiuk @the-thing maybe?
Personally I think it is a step in the right direction especially because it aims to break up the long-standing issue with the generated fields being compile time dependencies.

david-gibbs-ig:
I should mention that users that want to continue using FIX50sp2 as the FIX version, incorporating extensions from FIXLatest will still be able to do that, either by using the a custom FIX Dictionary generated from an Orchestration with the existing code generation (as long as they cut it down) OR by using the new code generation but  keeping FIX50sp2 as the FIX version. I think the way to make sense of this PR is to check it out and build it, this PR really just sits on the side of the existing QFJ without affecting the existing project.  Then do the same for the structural changes PR, https://github.com/quickfix-j/quickfixj/pull/460 and use the readmes to understand what it's trying to accomplish. I will make time to introduce and discuss it. This work is intended to simplify maintenance of custom ROEs and unlock FIXLatest for QFJ users. For my company this will be beneficial so I'm very keen to progress it.

donmendelson:
Perusing the code, I do not see any red flags, but I haven't tested its operation. I just want to lend my support for the concept. In practice, no one uses all of FIX Latest or even earlier FIX versions. The big idea of Orchestra is to provide machine readable rules of engagement. A FIX engine should use it to support the exact message layouts as specified rather than a theoretical universe of possibilities. Agree that the build of message classes should be separated from the build of the core engine.

david-gibbs-ig:
@chrjohn Can you rerun these checks please ? The problem with the LGTM check was Bad Gateway trying to download from maven. One needs write permissions to rerun checks, thanks Dave

jvirtanen:
> @chrjohn Can you rerun these checks please ? The problem with the LGTM check was Bad Gateway trying to download from maven. One needs write permissions to rerun checks, thanks Dave

[LGTM will be shut down in the coming months](https://github.blog/2022-08-15-the-next-step-for-lgtm-com-github-code-scanning/) with GitHub code scanning available as a replacement (from the same development team). As the latter is already in use, perhaps it would be time to disable LGTM on this repository.

chrjohn:
> As the latter is already in use, perhaps it would be time to disable LGTM on this repository.

Thanks for the reminder. I've disabled it now. Hopefully the check will be gone after closing and reopening.

pom.xml
```diff
@@ -89,7 +93,135 @@
...
+				<artifactId>mina-core</artifactId>
+				<version>${apache.mina.version}</version>
+			</dependency>
+			<dependency>
+			    <groupId>commons-io</groupId>
+			    <artifactId>commons-io</artifactId>
+			    <version>${commons.io.version}</version>
+			</dependency>
+			<dependency>
+			  <groupId>com.google.guava</groupId>
```
chrjohn, line 179: What is guava used for?

pom.xml
```diff
@@ -223,6 +361,53 @@
 						</execution>
 					</executions>
 				</plugin>
+				<plugin>
+					<groupId>org.codehaus.mojo</groupId>
+					<artifactId>xml-maven-plugin</artifactId>
+					<version>${xml-maven-plugin-version}</version>
+				</plugin>
+				<plugin>
+					<groupId>org.jboss.tattletale</groupId>
```
chrjohn, line 370: What is this used for?

chrjohn (commented):
Is it possible that the name of the new module is changed to something like `quickfixj-...` to have all module names follow the same schema?

chrjohn (commented):
There seems to be some duplication inside the new module as far as the Orchestra logo and the readme is concerned. The parent contains the same readme and logo as the submodule `repository-to-quickfixj`.

pom.xml
david-gibbs-ig: Tattletale generates some very useful build-time reports on dependencies.  https://tattletale.jboss.org/ 
image

chrjohn: OK, can this be integrated on github somehow? I suppose I need to run it manually? What did you use it for?

david-gibbs-ig:
> Is it possible that the name of the new module is changed to something like `quickfixj-...` to have all module names follow the same schema?

Yes that makes sense, I think we can change artefact name as well. Just suggest we agree with FTC.

pom.xml
chrjohn: It would be good if we only had the bare minimum of dependencies that are strictly necessary for the code generation and FIXLatest respectively.

david-gibbs-ig:
> There seems to be some duplication inside the new module as far as the Orchestra logo and the readme is concerned. The parent contains the same readme and logo as the submodule `repository-to-quickfixj`.

in progress.

pom.xml
david-gibbs-ig: Used in Maven plugins.
image

chrjohn: OK, but isn't this a transitive dependency then? Does it need to be declared in the POM?

david-gibbs-ig: It does need to be declared when you coerce (manage)  a transitive dependency to use a later version, in this case 31. 
I looked through the code and in the _full restructured_ code a guava feature is used in the project's mvn plugin source code. 
In this PR I will test to see if mvn plugins work with version 18.

david-gibbs-ig: @chrjohn - PR updated.

pom.xml
david-gibbs-ig: @chrjohn - PR updated.

david-gibbs-ig:
@chrjohn  I think I have updated the PR to conform to your comments. Please re-run the build as 2 macOS steps failed. Thanks and regards. Dave

david-gibbs-ig:
@chrjohn Could you let me know please if you are happy with the changes as per your comments ? If so I will also update [PR 460 ](https://github.com/quickfix-j/quickfixj/pull/460)

chrjohn:
Hi Dave. Thanks for all your work. The only thing I am thinking about is about the code generator being a copy. Ideally we should only have one source. @donmendelson any thoughts?

david-gibbs-ig:
@chrjohn @donmendelson It's a good point. There a few dimensions to the question. To help me make a case, I'll restate how I think using QFJ may work in the future. See https://github.com/quickfix-j/quickfixj/pull/460 for additional context.
To make QFJ implementations easier to maintain for users I think it can help to make the base/core runtime distinct from the generated Field and Message Classes. The base/core runtime only depends on a small number of Fields that can be included in the distribution. In this way users that want a type safe customised Rules of Engagement can maintain the Field and Messages classes and do not need to maintain base/core. They can upgrade Major/Minor versions of QFJ base/core by simply updating their dependencies (and build and test). This is very useful for users that maintain more than one integration that have different Rules of Engagement.
FIX Orchestra is another part of the picture, it is intended to provide a machine readable and standard way to define a Rules of Engagement. It can be helpful if QuickFIX/J users can readily adapt to ROEs defined using a FIX Orchestra repository. Similarly FIX API provider can build and distribute resources (Field/Message distributions, Dictionaries, documents) generated from their  FIX Orchestra repository. 
N.B. The direction I have proposed is to allow QuickFIX/J to support code generation from FIX Orchestra and/or using the XSLT approach. The intention is to provide the greatest compatibility and even full compatibility with users' existing implementations while also allowing a path ahead for code generation directly from a FIX Orchestra repository. The FIX Dictionary is, of course, also a runtime resource for QFJ so this comes into the picture even where a FIX Orchestra repository is the canonical definition of the Rules of Engagement. 
So :

Firstly the QFJ Code Generation, I think this belongs in the QFJ project because they are coupled, the reason for the Code Generation to change is when QuickFIX/J changes. This code utility can be productively maintained in the QuickFIX/J project. It's all about QuickFIX/J. The generated code can be optimised within the scope of the QFJ project.

The Dictionary generation is a perhaps debatable, it's a QuickFIX dictionary and can be used for any of the QuickFIX family of implementations and even Artio. We could probably make a case that this could be maintained by FTC, but :
* "primary use cases" that I would propose for using FIX Orchestra with QFJ do end up depending on the Dictionary generation, having the tools be part of QuickFIX/J gives good control of the users' experience and processes
* this Dictionary format does belong to the QuickFIX family so I think it is more ideal for the implementation to be hosted in the QuickFIX/J family. The tools are all set up for Java and Maven so QFJ is a good home.
* I don't think FTC is focussed on these works as a key deliverable so I'll say "control" again. The code here is derived from FTC code but the original repo is not really maintained, all the "proof of the pudding" is now in this project. @donmendelson can confirm if he is content for the code to be maintained here.
* The QuickFIX/J project has appropriate tooling to build the artefacts and publish them to Maven Central.
* Having these tools within the QuickFIX/J project should make it easier to verify new releases of FIX Orchestra specifications and any impact of changes to QuickFIX/J on the FIX Orchestra QuickFIX/J tools.

The if it is OK with @donmendelson and @chrjohn,  I  would propose that the original [repo](https://github.com/FIXTradingCommunity/fix-orchestra-quickfix) README be updated with comments pointing here (and optionally be made read-only).

Some of the ways QFJ Messages can be built, with recommended versions:

* QFJ Dictionary -> QFJ XSLT -> generated classes (FIX 4.0->FIX50SP2)

* FIX Orchestra Repository -> Dictionary Generation -> QFJ Dictionary -> QFJ XSLT -> generated classes (FIX 4.4->FIX50SP2)

* FIX Orchestra Repository -> Dictionary Generation -> QFJ Dictionary
   FIX Orchestra Repository -> Code Generation -> generated classes (FIX 4.4, FIX50SP2, FIX Latest)

donmendelson:
@david-gibbs-ig  and @chrjohn, I agree that Orchestra-to-QuickFIX data dictionary and message code generation should be owned and maintained by the quickfixj project. The module under FIXTradingCommunity should be considered a prototype. QuickFIX was chosen for the demonstration simply because of its prevalence as an open-source FIX engine; no vendor favoritism was intended. Nevertheless, the quickfixj project is the authority on its own API.

I also agree that the lifecycle for maintenance of the core engine will likely be different than the message definitions as represented by Orchestra. If for example, an exchange or counterparty issues an updated Orchestra file, then the message classes need to be regenerated even if there is no change to the core technology or API. Also consider the case where a party communicates with two or more counterparties that each issue an Orchestra file. Messages must be separately generated for each.

You may wish to break out the Orchestra to data dictionary and code generation into a separate GitHub repository under quickfixj organization.

chrjohn:
Thanks @donmendelson and @david-gibbs-ig 
I think to create a repo under the quickfixj org, as Don suggested, sounds sensible. 
IANAL so how is this from the legal standpoint? If I copied the code over from the FTC repo, I assume I have to keep the current Apache license with it?

david-gibbs-ig:
Yes we need to keep the Apache License but that is the extent of the complexity. It’s a good angle for having the separate repo.   They should have only the Apache license.
The artefacts can be published under the QFJ organization in Maven Central.
I looked yesterday and I think that I factored out any direct dependencies on QFJ, the dependencies should all be one way – from QFJ to these tools.
I will double check the dependencies.
Dave

chrjohn:
For now I've created https://github.com/quickfix-j/quickfixj-orchestra but still need to remove everything that we don't need and adapt package names and so on...

david-gibbs-ig:
> For now I've created https://github.com/quickfix-j/quickfixj-orchestra but still need to remove everything that we don't need and adapt package names and so on...

Do you want me to do that as I made a number of changes that are only in the quickfix/j copy of the source ? Do you want the new quickfix/j module names or the original ?

chrjohn:
> Do you want me to do that as I made a number of changes that are only in the quickfix/j copy of the source ? Do you want the new quickfix/j module names or the original ?

Would be great if you could do that. I'd favour new module names unless there is a reason to keep the old ones?
 Thank you in advance and happy new year.

david-gibbs-ig:
@donmendelson @chrjohn I created a first cut for updates to https://github.com/quickfix-j/quickfixj-orchestra and put a short explanation about why I pruned the repository.

chrjohn:
@david-gibbs-ig I did an initial release of https://github.com/quickfix-j/quickfixj-orchestra which hopefully should be available through Maven Central soon.

```
[INFO] Remote staging repositories released.
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for org.quickfixj.orchestra:parent 1.0.0:
[INFO] 
[INFO] org.quickfixj.orchestra:parent ..................... SUCCESS [  2.873 s]
[INFO] quickfixj-from-fix-orchestra ....................... SUCCESS [  1.314 s]
[INFO] quickfixj-from-fix-orchestra-generator ............. SUCCESS [  4.177 s]
[INFO] quickfixj-from-fix-orchestra-code-generator-maven-plugin SUCCESS [  6.263 s]
[INFO] quickfixj-from-fix-orchestra-dictionary-generator-maven-plugin SUCCESS [01:27 min]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  01:42 min
[INFO] Finished at: 2023-01-02T17:38:57+01:00
```

chrjohn:
@david-gibbs-ig did you close this by mistake?

david-gibbs-ig:
> @david-gibbs-ig did you close this by mistake?

No, I think we can close it, This is the part that is now in https://github.com/quickfix-j/quickfixj-orchestra

PR #494, opened 2022-07-26
Move `unbind` into `closeManagedSessionsAndDispose`

chrjohn:
Sometimes the unit test `SocketAcceptorTest` is failing because it cannot bind on an address that was used before. More often than not this seems to happen on MacOS.
In this PR we try a little harder to get rid of all sessions but it also may be a race condition somewhere in MINA...


chrjohn:
Closing since failures were still happening from time to time.

PR #529, opened 2022-08-04
Optimize `convertToLocalDateTime` method

leonchen83:
Signed-off-by: Baoyi Chen <chen.bao.yi@qq.com>
Original PR: #456

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/field/converter/UtcTimestampConverter.java
+++ b/quickfixj-core/src/main/java/quickfix/field/converter/UtcTimestampConverter.java
@@ -24,10 +24,10 @@ import quickfix.FieldConvertError;
 import quickfix.SystemTime;
 
 import java.text.DateFormat;
+import java.time.DateTimeException;
 import java.time.LocalDateTime;
 import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
-import java.time.format.DateTimeParseException;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.GregorianCalendar;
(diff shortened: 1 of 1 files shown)


chrjohn:
@leonchen83 
Could you remove the leap second fix from this PR and open a separate with the changes from this (i.e. #529) PR plus the leap second fix against master?
Or alternatively open another PR *only* with the optimization but *without* the leap second fix against master.
Thank you.

leonchen83:
sorry reply late. I rollbacked leap second code. please merge this PR first

quickfixj-core/src/main/java/quickfix/field/converter/UtcTimestampConverter.java
```diff
@@ -135,26 +135,37 @@ public static Date convert(String value) throws FieldConvertError {
...
             }
-        } catch (DateTimeParseException e) {
+    
+            int yy = parseInt(value, 0, 4);
+            int mm = parseInt(value, 4, 2);
+            int dd = parseInt(value, 6, 2);
+            int h = parseInt(value, 9, 2);
+            int m = parseInt(value, 12, 2);
+            int s = parseInt(value, 15, 2);
+            return LocalDateTime.of(yy, mm, dd, h, m, s, ns);
```
chrjohn, line 155: Probably a micro optimisation, but could we have only the `LocalDateTime.of` in the try-block? It shouldn't be necessary for the `parseInt()` method.

leonchen83: that make sense

chrjohn:
Hi @leonchen83 ,
thanks for the PR. I have just merged it. Before you create a PR for the leap second stuff, please wait until further notice. I need to clean something up first.
Cheers,
Chris.

PR #531, opened 2022-08-04
Test that replicates a scenario where an initiator gets logged out by…

cristipop93:
… the acceptor and it can not logon again until the acceptor restarts.

The steps to replicate are:
1. The initiator connects to the acceptor
2. The acceptor accepts the connection, but after a while (it could be any reason. Here after the 4th message received fromAdmin, the acceptor logs out the initiator.) the acceptor decides to logout the initiator.
3. The initiator is logged out and can not reconnect to the acceptor because: "Logon attempt when session is disabled".
The initiator can reconnect only after the acceptor is restarted, and the enabled session variable is back to true.

### DIFF
--- a/quickfixj-core/src/test/java/quickfix/test/acceptance/ATApplication.java
+++ b/quickfixj-core/src/test/java/quickfix/test/acceptance/ATApplication.java
@@ -38,6 +38,8 @@ public class ATApplication implements Application {
     private final MessageCracker outboundCracker = new MessageCracker(new Object());
     private boolean isLoggedOn;
 
+    private static int count = 0;
+
     public void onCreate(SessionID sessionID) {
         assertNoSessionLock(sessionID);
         Session.lookupSession(sessionID).reset();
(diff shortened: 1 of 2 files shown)


chrjohn:
The changes to `ATApplication` will probably make other tests fail too. But I get the idea. :)

quickfixj-core/src/test/java/quickfix/test/acceptance/ATApplication.java
```diff
@@ -81,6 +83,11 @@ public void toApp(Message message, SessionID sessionID) throws DoNotSend {
     public void fromAdmin(Message message, SessionID sessionID) throws FieldNotFound,
             IncorrectDataFormat, IncorrectTagValue, RejectLogon {
         assertNoSessionLock(sessionID);
+
+        if (count > 2) { // leave the logon and some heartbeats go through, then logout the session.
+            Session.lookupSession(sessionID).logout();
```
chrjohn, line 88: As communicated using `logout()` will disable the session until `logon()` is called again. Please use `generateLogout()` or throw `RejectLogon`.

Issue #474, opened 2022-03-24
ResetMsgSeqNum flag in Logon message is processed prior to application level logon message validation


dilanperera87:
**Describe the bug**
We’ve developed a fix trading gateway using quickfixj library and user validation is performed in **fromAdmin** callback function when receiving a logon message. 
If an user sends logon request with invalid credentials and ResetSeqNumFlag set to `Y`, then sequence numbers are getting reset prior to the application level logon message validation since **fromAdmin** callback is fired after resetting the session message store.
Due to this issue, an unauthorized user will be able to reset session sequence numbers and messages. 

**To Reproduce**
1. Establish TCP/IP connection and send Logon message (MsgSeqNum=1)
2. Acceptor accepts logon and send Logon response (MsgSeqNum=1)
3. Exchange few messages to increase sequence numbers.
4. Logout/disconnect session.
5. Assume next target sequence number is 5 and next sender sequence number is 10.
6. Re-establish TCP/IP connection and send Logon message with ResetSeqNumFlag=Y and MsgSeqNum=1. (Consider this as an invalid logon attempt)
7. Acceptor rejects logon request (fromAdmin function implementation throws RejectLogon exception) and connection gets terminated.
8. Re-establish TCP/IP connection and send Logon message (MsgSeqNum=5) to continue previous valid session disconnected at step 4.
9. Acceptor accepts logon and send Logon response with (MsgSeqNum=1)
10. Acceptor sends ResendRequest requesting messages from sequence number 2

**Expected behavior**
MsgSeqNum should be 10 in the Logon response sent in step 9 above and acceptor should not request missed messages via ResendRequest as in step 10.

**system information:**
 - OS: Linux
 - Java version: JDK11
 - QFJ Version: 2.3.0

chrjohn:
Hmm, probably the `verify()` method call needs to be moved further up in `nextLogon()` but this needs to be tested carefully.
Are you able to provide a unit test?

dilanperera87:
Sure, I'll do the changes and test it locally first and the raise a PR.

harinda05:
@chrjohn raised a PR for this issue. Please let me know if everything's in line. This is my first PR for QuickFixJ

Issue #482, opened 2022-04-22
 System doesn't break TCP/IP connection when sending Logon message with ResetSeqNumFlag (141)=Y and Invalid MsgSeqNum (!=1)


harinda05:
**Describe the bug**
We’ve developed a fix trading gateway using quickfixj library.

When a user sends a Logon message to the system for a user, with ResetSeqNumFlag = Y and MsgSeqNum !=1 we expect the system to break the TCP/IP connection after sending a Logout message due to the MsgSeqNum being invalid.

**To Reproduce**
Step to reproduce:

1. Establish TCP/IP connection 
2.  Send Logon message to the system with SenderCompID of active user
3.  Receive Logon message from the system 
4. Receive Heartbeat message
5.  Send Heartbeat message 
6.  Send Logout message 
7.  Receive Logout message 
8.  Send Logon message to the system with SenderCompID of active user and ResetSeqNumFlag = Y and MsgSeqNum !=1 
9. Receive Logout message

**Expected behavior**

- Expected behavior:
logout message from the system since MsgSeqNum != 1

Actual behavior:
System isn’t  sending logout. Instead sends a logon response.

**system information:**
 - OS: [e.g. Linux]
 - Java version [e.g. JDK11]
 - QFJ Version [e.g. 2.0.3]

chrjohn:
Wow, I had to do some FIX spec digging to find out if this really is a bug. At the moment I am not sure. Can you present some excerpt from the FIX spec that says that this really is a bug?
Thanks,
Chris.

harinda05:
According to [FIX_Session_Layer_with_Errata_November_2020.pdf](https://www.fixtrading.org/packages/fix-session-layer-technical-specification/) Section 5.4.7

```
5.4.7 Validation of message sequence numbers
If a received message has a MsgSeqNum(34) less than NextNumIn, a serious error exists. The FIX session and transport
layer connection should be terminated unless the message received is:

• a SequenceReset(35=4) message with PossDupFlag(43) present with a value of “Y”.
• any message of a message type that supports retransmission with a PossDupFlag(43) present with a value
of “Y”.
• a Logon(35=A) message with ResetSeqNumFlag(141) present with a value of “Y”.

_**If a received message has a MsgSeqNum(34) greater than NextNumIn, a serious error exists. The FIX session and
transport layer connection should be terminated.**_
```
[...]

PR #549, opened 2022-09-19
Validate `Logon` message before acting upon `ResetMsgSeqNum` flag

harinda05:
Fixes https://github.com/quickfix-j/quickfixj/issues/474 - 

verify() method call moved further up in nextLogon() to prevent seqNumber resets over invalid logins.

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/Session.java
+++ b/quickfixj-core/src/main/java/quickfix/Session.java
@@ -1771,12 +1771,14 @@ public class Session implements Closeable {
                 return false;
             }
 
-            if (checkTooHigh && isTargetTooHigh(msgSeqNum)) {
-                doTargetTooHigh(msg);
-                return false;
-            } else if (checkTooLow && isTargetTooLow(msgSeqNum)) {
-                doTargetTooLow(msg);
-                return false;
+            if(!state.isResetReceived()){
+                if (checkTooHigh && isTargetTooHigh(msgSeqNum)) {
+                    doTargetTooHigh(msg);
+                    return false;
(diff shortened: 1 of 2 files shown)


chrjohn:
@harinda05 
Thanks for the PR. Could you please base this against master?
Thanks in advance

harinda05:
> [chrjohn](/chrjohn)

Changed the base branch to master

quickfixj-core/src/main/java/quickfix/Session.java
```diff
@@ -1771,12 +1771,14 @@ private boolean verify(Message msg, boolean checkTooHigh, boolean checkTooLow)
                 return false;
             }
 
-            if (checkTooHigh && isTargetTooHigh(msgSeqNum)) {
-                doTargetTooHigh(msg);
-                return false;
-            } else if (checkTooLow && isTargetTooLow(msgSeqNum)) {
-                doTargetTooLow(msg);
-                return false;
+            if(!state.isResetReceived()){
```
chrjohn, line 1774: So if you receive a ResetSeqNum flag you reset in either case, without checking sequence numbers of the Logon message. This somewhat contradicts what you write in #482 

> When a user sends a Logon message to the system for a user, with ResetSeqNumFlag = Y and MsgSeqNum !=1 we expect the system to break the TCP/IP connection after sending a Logout message due to the MsgSeqNum being invalid.

So is #482 no longer a problem and can be closed?

harinda05: I think for the time being #482 can be ignored. Better to revisit that scenario in my opinion after more investigation

PR #559, opened 2022-11-03
Improve `IntConverter` performance

chrjohn:
Replaced `Long.toString()` by call to `NumbersCache.get()` to prevent unnecessary allocation for integers ranging from 0 to 99999.

Improved performance of integer parsing for integers shorter than `Integer.MAX_VALUE`.
Heuristic: since we have no range check in our `parseInt()` we only parse values which have at least one digit less than `Integer.MAX_VALUE` and leave longer Strings to `Integer.parseInt()`.
NB: we must not simply reject strings longer than `MAX_VALUE` since they could possibly include an arbitrary number of leading zeros.

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/field/converter/IntConverter.java
+++ b/quickfixj-core/src/main/java/quickfix/field/converter/IntConverter.java
@@ -27,6 +27,8 @@ import quickfix.NumbersCache;
  */
 public final class IntConverter {
 
+    private static final String INT_MAX_STRING = String.valueOf(Integer.MAX_VALUE);
+    
     /**
      * Convert an integer to a String
      *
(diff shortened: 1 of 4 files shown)


chrjohn:
Used benchmark:

```
package org.sample;

import org.openjdk.jmh.annotations.Benchmark;
import org.openjdk.jmh.annotations.Fork;
import org.openjdk.jmh.annotations.Measurement;
import org.openjdk.jmh.annotations.Scope;
import org.openjdk.jmh.annotations.State;
import org.openjdk.jmh.annotations.Warmup;
import org.openjdk.jmh.infra.Blackhole;

@Fork( 1 )
@State( Scope.Benchmark )
@Warmup( iterations = 3, time = 3, timeUnit = TimeUnit.SECONDS )
@Measurement( iterations = 5, time = 3, timeUnit = TimeUnit.SECONDS )
public class IntConverterBenchmark3 {

    public static void main( String[] args ) throws Exception {
        Options opt = new OptionsBuilder().shouldDoGC( true )
                                          .include( IntConverterBenchmark3.class.getSimpleName() )
                                          .build();
        new Runner( opt ).run();
    }



    @Benchmark
    public void benchmarkOld( Blackhole bh ) throws Exception {
        bh.consume( oldConvert( "220912" ) );
    }



    @Benchmark
    public void benchmarkNew( Blackhole bh ) throws Exception {
        bh.consume( convert( "220912" ) );
    }

    private static final String INT_MAX_STRING = String.valueOf( Integer.MAX_VALUE );

    public static int convert( String value ) throws Exception {

        if ( !value.isEmpty() ) {
            final char firstChar = value.charAt( 0 );
            boolean isNegative = ( firstChar == '-' );
            if ( !Character.isDigit( firstChar ) && !isNegative ) {
                throw new Exception( "invalid integral value: " + value );
            }
            int minLength = ( isNegative ? 2 : 1 );
            if ( value.length() < minLength ) {
                throw new Exception( "invalid integral value: " + value );
            }

            // Heuristic: since we have no range check in parseInt() we only parse
            // values which have at least one digit less than Integer.MAX_VALUE and
            // leave the rest to Integer.parseInt().
            // NB: we must not simply reject strings longer than MAX_VALUE since
            // they could possibly include an arbitrary number of leading zeros.
            int maxLength = ( isNegative ? INT_MAX_STRING.length() : INT_MAX_STRING.length() - 1 );
            if ( value.length() <= maxLength ) {
                return parseInt( value, isNegative );
            } else {
                try {
                    return Integer.parseInt( value );
                } catch ( NumberFormatException e ) {
                    throw new Exception( "invalid integral value: " + value + ": " + e );
                }
            }
        } else {
            throw new Exception( "invalid integral value: empty string" );
        }
    }



    static int parseInt( String value, boolean isNegative ) throws Exception {
        int num = 0;
        int firstIndex = ( isNegative ? 1 : 0 );
        for ( int i = firstIndex; i < value.length(); i++ ) {
            if ( Character.isDigit( value.charAt( i ) ) ) {
                num = ( num * 10 ) + ( value.charAt( i ) - '0' );
            } else {
                throw new Exception( "invalid integral value: " + value );
            }
        }
        return isNegative ? -num : num;
    }



    static int oldConvert( String value ) throws Exception {
        try {
            for ( int i = 0; i < value.length(); i++ ) {
                if ( !Character.isDigit( value.charAt( i ) ) && ! ( i == 0 && value.charAt( i ) == '-' ) ) {
                    throw new Exception( "invalid integral value: " + value );
                }
            }
            return Integer.parseInt( value );
        } catch ( NumberFormatException e ) {
            throw new Exception( "invalid integral value: " + value + ": " + e );
        }
    }

}

```

Result:

```
Benchmark                                                 Mode  Cnt           Score           Error   Units
IntConverterBenchmark3.benchmarkNew                      thrpt    5  1608495228.416 ± 181808582.138   ops/s
IntConverterBenchmark3.benchmarkNew:·gc.alloc.rate       thrpt    5          ≈ 10⁻⁴                  MB/sec
IntConverterBenchmark3.benchmarkNew:·gc.alloc.rate.norm  thrpt    5          ≈ 10⁻⁷                    B/op
IntConverterBenchmark3.benchmarkNew:·gc.count            thrpt    5             ≈ 0                  counts
IntConverterBenchmark3.benchmarkOld                      thrpt    5   200878711.062 ±  24333721.274   ops/s
IntConverterBenchmark3.benchmarkOld:·gc.alloc.rate       thrpt    5          ≈ 10⁻⁴                  MB/sec
IntConverterBenchmark3.benchmarkOld:·gc.alloc.rate.norm  thrpt    5          ≈ 10⁻⁶                    B/op
IntConverterBenchmark3.benchmarkOld:·gc.count            thrpt    5             ≈ 0                  counts

```

quickfixj-core/src/main/java/quickfix/field/converter/IntConverter.java
```diff
@@ -43,20 +46,66 @@ public static String convert(int i) {
...
+     * @param value the String to convert
+     * @param isNegative if passed String is negative, first character will
+     * be skipped since it is assumed that it contains the negative sign
+     * @return the converted int
+     */
+    private static int parseInt(String value, boolean isNegative) throws FieldConvertError {
+        int num = 0;
+        int firstIndex = (isNegative ? 1 : 0);
+        for (int i = firstIndex; i < value.length(); i++) {
+            if (Character.isDigit(value.charAt(i))) {
```
jvirtanen, line 102: Note that [`Character.isDigit(char)`](https://docs.oracle.com/javase/8/docs/api/java/lang/Character.html#isDigit-char-) returns true also for other than the Arabic numerals in the ASCII character range. Would it be better to explicitly check for the range from `'0'` to `'9'`?

chrjohn: I used that 0-9 check at first but then realised that formerly already `Character.isDigit(char)` was used in the old code. I checked the implementation and was under the impression that the 0-9 check was used in `Character.isDigit` as well for the Latin charset.
https://github.com/openjdk/jdk/blob/f43bb9feaa03008bad9708a4d7ed850d2532e102/src/java.base/share/classes/java/lang/CharacterDataLatin1.java.template#L84

jvirtanen: If we already know that the `String` contains only Latin-1 characters, then it will indeed be like that. This logic has been in place since Java 9’s compact strings. In Java 8, [it will still check the Unicode character range](https://github.com/openjdk/jdk/blob/9a9add8825a040565051a09010b29b099c2e7d49/jdk/src/share/classes/java/lang/Character.java#L5668).

chrjohn: I didn't check the JDK8 code. So I'll better make it explicit, thank you for your review. :+1:

PR #567, opened 2022-11-15
`FieldMap` - enhancing thrown exception for null values

### LINKED ISSUE #458: Enhance exception type and message in `FieldMap` class `setField()` method
This is an enhancement.
In class FieldMap, method setField(StringField field) throws a NullPointerException, but it does not mention which field:
`throw new NullPointerException("Null field values are not allowed.");`

I would suggest to throw an exception like following to mention the field name and also a better exception type like FieldException:
`throw new FieldException(SessionRejectReason.NULL_DATA_FOR_VALUE,
                    field.getField());`

vinipx:
fixes #458  -  Raising this PR to trigger the discussion.
If we decide to create new `SessionRejectReason.NULL_DATA_FOR_VALUE`, then we might need to adjust **quickfixj-messages** XML for all versions - _e.g. 40, 41, 42, 43, 44, 50, 50sp1, 50sp2, fixt11_ - since `SessionRejectReason` seems to be generated based on them.

For now, decided to keep **OTHER**.

**Alternative**: create specific e.g. `NullFieldValueException` similar to `FieldNotFound`

@chrjohn please review and let me know your thoughts on how to implement it properly.

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/FieldMap.java
+++ b/quickfixj-core/src/main/java/quickfix/FieldMap.java
@@ -337,7 +337,7 @@ public abstract class FieldMap implements Serializable, Iterable<Field<?>> {
 
     public void setField(StringField field) {
         if (field.getValue() == null) {
-            throw new NullPointerException("Null field values are not allowed.");
+            throw new FieldException(SessionRejectReason.OTHER, field.getField());
         }
         fields.put(field.getField(), field);
     }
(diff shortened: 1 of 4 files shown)


chrjohn:
I think I'd go for number 4="Tag specified without a value" as reject reason.

vinipx:
> I think I'd go for number 4="Tag specified without a value" as reject reason.

Good catch! I will use it.

vinipx:
@chrjohn PR is ready. Please, feel free to review - comments are welcome :)

chrjohn:
@vinipx 
There is a unit test failure. Could you please take a look?

vinipx:
> @vinipx There is a unit test failure. Could you please take a look?

fixed! @chrjohn could you please review? tks

chrjohn (approved):

Issue #486, opened 2022-05-09
Configuring `StartTime` and `EndTime` on different timezones results in wrong schedule calculation


sregueiroqb:
**Describe the bug**
We are connecting to an FX market that has an opening time on the Madrid timezone (at 00:00) and close time on the New York timezone (at 17:00), Monday to Friday, due to fixing reasons.
The local time is UTC, the application runs in kubernetes.
To avoid having to change the config when daylight savings changes, we setup the session the following way:

```
StartTime=23:55:00 Europe/Madrid
EndTime=17:05:00 America/New_York
Weekdays=Sun,Mon,Tue,Wed,Thu
```

We want to connect 5 minutes before the session opens and disconnect 5 minutes after.

Using this configuration, results on the session connecting at 23:55 Madrid (21:55 UTC) and disconnecting at 00:00 Madrid (22:00 UTC), reconnecting again at 06:00 Madrid (04:00 UTC)

The issue is found in `DefaultSessionSchedule`. `theMostRecentIntervalBefore` calculates the wrong interval when changing days and the close time is before open time.

The log shows the correct schedule `[FIX.5.0:SENDER->TARGET] sunday, monday, tuesday, wednesday, thursday, 21:55:00-UTC - 21:05:00-UTC (sunday, monday, tuesday, wednesday, thursday, 23:55:00-MESZ - 17:05:00-EDT)`

**To Reproduce**
```
package quickfix;

import static org.junit.Assert.assertEquals;

import java.io.ByteArrayInputStream;
import java.util.Calendar;
import java.util.GregorianCalendar;
import java.util.Locale;
import java.util.TimeZone;

import org.junit.After;
import org.junit.Before;
import org.junit.Test;

import quickfix.ConfigError;
import quickfix.DefaultSessionSchedule;
import quickfix.SessionID;
import quickfix.SessionSchedule;
import quickfix.SessionSettings;
import quickfix.SystemTime;

import java.text.DateFormat;
import java.text.DateFormatSymbols;
import java.text.ParseException;
import java.text.SimpleDateFormat;
import java.util.Date;

import quickfix.field.converter.UtcTimeOnlyConverter;

public class WrongFixSessionScheduleTest {

	private MockSystemTimeSource mockSystemTimeSource;
	private Locale defaultLocale;

	@Before
	public void setUp() throws Exception {
		mockSystemTimeSource = new MockSystemTimeSource();
		SystemTime.setTimeSource(mockSystemTimeSource);
[...]

PR #568, opened 2022-11-15
Improve handling of Weekday schedules when EndTime < StartTime

AndreyNudko:
In FX session typically starts on Sun evening and is reset around 5PM NewYork time. So daily schedule may look like this
 Sun 17:05 - Mon 16:55
 Mon 17:05 - Tue 16:55
 Tue 17:05 - Wed 16:55
 Wed 17:05 - Thu 16:55
 Thu 17:05 - Fri 16:55

I was expecting that below settings would achieve this:
```
  Weekdays = SUNDAY,MONDAY,TUESDAY,WEDNESDAY,THURSDAY,FRIDAY
  StartTime = 17:05:00
  EndTime = 16:55:00
```
But they don't, because DefaultSessionSchedule considers the interval [Fri 17:05 - Sat 16:55] to be in session - even if Sat is not a weekday.

This change improves weekday behaviour by going backwards until both start and end days are weekdays. It does not have any impact in cases when EndTime > StartTime.

As a precaution, if suitable interval could not be found we revert to original behaviour. This should be configuration error, but even then returning something which is not quite right is better than getting stuck in infinite loop.

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/DefaultSessionSchedule.java
+++ b/quickfixj-core/src/main/java/quickfix/DefaultSessionSchedule.java
@@ -210,6 +210,24 @@ public class DefaultSessionSchedule implements SessionSchedule {
 
             if (intervalEnd.getTimeInMillis() <= intervalStart.getTimeInMillis()) {
                 intervalEnd.add(Calendar.DAY_OF_WEEK, 1);
+
+                // go back even further so both start and end time are weekdays
+                // this allows handling schedules like "Sun 17:05 - Mon 16:55,  ..., Thu 17:05 - Fri 16:55"
+                // in case of a really screwed up configuration we may not find a suitable interval,
+                // so limit this loop to a week and undo the change afterwards if we failed
+                int undoOffset = 0;
+                for (int i = 0; i < 7; i++) {
+                    if (validDayOfWeek(intervalStart) && validDayOfWeek(intervalEnd)) {
+                        undoOffset = 0;
+                        break;
+                    }
+
+                    undoOffset++;
+                    intervalStart.add(Calendar.DAY_OF_WEEK, -1);
+                    intervalEnd.add(Calendar.DAY_OF_WEEK, -1);
+                }
+                intervalStart.add(Calendar.DAY_OF_WEEK, undoOffset);
+                intervalEnd.add(Calendar.DAY_OF_WEEK, undoOffset);
             }
 
         } else {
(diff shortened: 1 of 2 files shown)


AndreyNudko:
As a bit of background
It's not really a problem that session is still trying to login after the end of last weekday - the other end (acceptor) either doesn't care or doesn't allow it.
But there are monitoring tools using DefaultSessionSchedule to decide whether or not session should be active, and current behaviour confuses them
I could implement a custom schedule, but this seems like a generally useful change
Let me know your thoughts - I don't fully understand possible impact, but implementation is quite defensive

chrjohn:
Hi @AndreyNudko , thanks for the PR. :+1: 

Sounds like this bug, do you agree? https://github.com/quickfix-j/quickfixj/issues/486

AndreyNudko:
It looks similar, but two TimeZone-s makes it even more complicated
I don't think my change addresses #486 as it doesn't affect attached reproducer test
BTW, that test misleadingly passes as it has assertions on the current behaviour rather than expected one - at least they are marked as `// Wrong, should be in session`

AndreyNudko:
I think the other issue is that in certain configurations adjusting time by 1 day is not enough; here's how interval calculations go on Sun in the "23:55:00 Europe/Madrid - 17:05:00 America/New_York" scenario:

```
Mon 21:55:00 UTC --> Sun 21:05:00 UTC
Sun 21:55:00 UTC --> Sat 21:05:00 UTC
Sun 21:55:00 UTC --> Sun 21:05:00 UTC // stop here ; same day, but times are still the other way around
```

I can make the other test pass by swapping condition for a loop at DefaultSessionSchedule:212
```
            while (intervalEnd.getTimeInMillis() <= intervalStart.getTimeInMillis()) {
                intervalEnd.add(Calendar.DAY_OF_WEEK, 1);
            }
```

I would be inclined to submit that a separate patch, because it seems less intrusive than my change.

One thing which I realized is that my understanding of Weekdays is slightly different from 486.
I expect in FX-like schedule both Sun _and_ Fri to be configured as weekdays, i.e. the setting applies to both start and end of the session.
But they set weekdays as Sun-Thu - and maybe it actually makes more sense to only apply Weekdays to the start. If I misunderstood semantics of this option, my change could have side effects for a number of people - and I probably can just tweak my schedule.

AndreyNudko:
Actually, looking at https://github.com/quickfix-j/quickfixj/pull/133 which introduced Weekdays: "comma-delimited list of the days to **start** the session on"
Which I think is also enforced by existing tests

I'm going to close this PR and maybe submit another one to just improve documentation, as it's currently a bit vague

chrjohn:
@AndreyNudko but #574 is still valid?

AndreyNudko:
@chrjohn Yes, #574 is actual bugfix

PR #578, opened 2022-11-20
`MessageCracker.xsl` optimization using `HashMap`

### LINKED ISSUE #577: Remove "if else" statements which is slower by nearly 7 times compared to `HashMap` in generated `MessageCracker` class
**Describe the bug**
MessageCracker generated class contains a lengthy if else statement to resolve the method call for the respective  message type.
Current logic is about 7 times slower than using HashMap key'd by message type and value as  the method reference.

I wrote a new MessageCracker.xsl which use the HashMap to resolve the method respective to message type. 

Following is the JMH benchmarking results.

### Before:

benchmarkCrackingGenerated    - use the current MessageCracker.xls generated class
benchmarkCrackingWithHashMap  use custom MessageCracker class which use HashMaps to resolve methods

Benchmark                                              Mode  Cnt         Score         Error  Units
MessageCrackerBenchmark.benchmarkCrackingGenerated    thrpt    5  10**923194**.106 ± 2861197.934  ops/s
MessageCrackerBenchmark.benchmarkCrackingWithHashMap  thrpt    5  69**723897**.515 ± 3242125.694  ops/s

### After:

benchmarkCrackingGenerated    - use the new MessageCracker.xls generated class
benchmarkCrackingWithHashMap  use custom MessageCracker class which use HashMaps to resolve methods

Benchmark                                              Mode  Cnt         Score          Error  Units
MessageCrackerBenchmark.benchmarkCrackingGenerated    thrpt    5  71**897498**.754 ± 19147681.365  ops/s
MessageCrackerBenchmark.benchmarkCrackingWithHashMap  thrpt    5  72**165054**.010 ±  4905711.116  ops/s

**To Reproduce**
Run my JMH benchmark code in my forked repo for QFJ

### Before
[Before Code Benchmark](https://github.com/isurudevj/quickfixj/blob/feature/new-message-crakcer-stage1/quickfixj-performance-regression/src/main/java/org/quickfixj/MessageCrackerBenchmark.java)

### After
[After Code Benchmark](https://github.com/isurudevj/quickfixj/blob/feature/new-message-cracker/quickfixj-performance-regression/src/main/java/org/quickfixj/MessageCrackerBenchmark.java)

**Expected behavior**
N/A

**system information:**
 - OS: [Windows11]
 - Java version [JDK8]
 - QFJ Version [3.0.0-SNAPSHOT]

**Additional context**
N/A

busy-spin:
Fixes #577

### DIFF
--- a/quickfixj-codegenerator/src/main/resources/org/quickfixj/codegenerator/MessageCracker.xsl
+++ b/quickfixj-codegenerator/src/main/resources/org/quickfixj/codegenerator/MessageCracker.xsl
@@ -49,8 +49,19 @@ package <xsl:value-of select="$messagePackage"/>;
 
 import quickfix.*;
 import <xsl:value-of select="$fieldPackage"/>.*;
+import java.util.HashMap;
 
 public class MessageCracker {
+
+	private final HashMap&lt;String, MessageConsumer&gt; methodRegistry = new HashMap&lt;&gt;();
+	private final MessageConsumer defaultFunction = this::onMessage;
+
+	public MessageCracker() {
+	<xsl:for-each select="//fix/messages/message">
+		methodRegistry.put(<xsl:value-of select="@name"/>.MSGTYPE,
+			(message, sessionID) -> onMessage((<xsl:value-of select="@name"/>) message, sessionID));
+	</xsl:for-each>
+	}
 
 	/**
 	 * Callback for quickfix.Message message.
@@ -116,13 +127,13 @@ public class MessageCracker {
 			throws UnsupportedMessageType, FieldNotFound, IncorrectTagValue {
 
 		String type = message.getHeader().getString(MsgType.FIELD);
+		methodRegistry.getOrDefault(type, defaultFunction).accept(message, sessionID);
+	}
 
-		<xsl:for-each select="//fix/messages/message">
-		<xsl:if test="position()!=1">
-		else </xsl:if>if (type.equals(<xsl:value-of select="@name"/>.MSGTYPE))
-			onMessage((<xsl:value-of select="@name"/>)message, sessionID);</xsl:for-each>
-		else
-			onMessage(message, sessionID);
+	@FunctionalInterface
+	private interface MessageConsumer {
+		void accept(Message message, SessionID sessionID)
+			throws UnsupportedMessageType, IncorrectTagValue, FieldNotFound;
 	}
 </xsl:template>
 


isurudevj:
Following is the JMH benchmarking results.


### Before:

benchmarkCrackingGenerated    - use the current MessageCracker.xls generated class
benchmarkCrackingWithHashMap  use custom MessageCracker class which use HashMaps to resolve methods

Benchmark                                              Mode  Cnt         Score         Error  Units
MessageCrackerBenchmark.benchmarkCrackingGenerated    thrpt    5  10**923194**.106 ± 2861197.934  ops/s
MessageCrackerBenchmark.benchmarkCrackingWithHashMap  thrpt    5  69**723897**.515 ± 3242125.694  ops/s

### After:

benchmarkCrackingGenerated    - use the current MessageCracker.xls generated class
benchmarkCrackingWithHashMap  use custom MessageCracker class which use HashMaps to resolve methods

Benchmark                                              Mode  Cnt         Score          Error  Units
MessageCrackerBenchmark.benchmarkCrackingGenerated    thrpt    5  71**897498**.754 ± 19147681.365  ops/s
MessageCrackerBenchmark.benchmarkCrackingWithHashMap  thrpt    5  72**165054**.010 ±  4905711.116  ops/s



**To Reproduce**
Run my JMH benchmark code in my forked repo for QFJ

### Before
[Before Code Benchmark](https://github.com/isurudevj/quickfixj/blob/feature/new-message-crakcer-stage1/quickfixj-performance-regression/src/main/java/org/quickfixj/MessageCrackerBenchmark.java)


### After
[After Code Benchmark](https://github.com/isurudevj/quickfixj/blob/feature/new-message-cracker/quickfixj-performance-regression/src/main/java/org/quickfixj/MessageCrackerBenchmark.java)

chrjohn:
@chrjohn need to check if new code generator for FIX Latest can benefit from this.

mrFloony:
hi, I implemented my version of MessageCracker based on HashMap, and he showed a deterioration in processing speed (~30 ms difference compared to If-Else on one message). The replacement with a Switch turned out to be a little faster (a gain of ~15 ms with the same parameters). Tested on a pre-battle stand (with slightly reduced parameters of the virtual machine and allocated memory), as a result, the implementation on the Switch went to the product stand (three weeks - normal flight)
Therefore, I would rather replace the basic MessageCracker with an implementation from Switch

chrjohn:
@mrFloony thanks for your comment. Are you maybe able to create a benchmark as linked in https://github.com/quickfix-j/quickfixj/pull/578#issuecomment-1321206293 or as used in the quickfixj-perf-test module?

busy-spin:
hi @mrFloony

Can let me know more details about your test. 30ms or 15ms for cracking message seems  quiet high. Even the if-else approach is much faster than that. Do you have similar JMH test for this.

busy-spin:
> @chrjohn need to check if new code generator for FIX Latest can benefit from this.

Can let me know more details about this, I should be able to update my test. And also add comparison with switch statement vs hash map also.

chrjohn:
> > @chrjohn need to check if new code generator for FIX Latest can benefit from this.
> 
> Can let me know more details about this, I should be able to update my test. And also add comparison with switch statement vs hash map also.

Actually this is in a different project which generates the classes from orchestrations: https://github.com/quickfix-j/quickfixj-orchestra/blob/fac2493c93f79e21ffcb3cb67cc25459dc2f467d/quickfixj-from-fix-orchestra-repository/quickfixj-from-fix-orchestra-generator/src/main/java/org/quickfixj/orchestra/CodeGeneratorJ.java#L577

Issue #258, opened 2020-02-04
`DataDictionary.getVersion()` ignores service pack


drekbour:
**Describe the bug**
`dataDictionary.getVersion()` ignores the XML's `@servicepack` attribute. A FIX.5.0.SP2 dictionary XML will return "FIX.5.0". This feels wrong ...

**To Reproduce**
This from your own `quickfix.MessageTest`: ...

        final DataDictionary dataDictionary = new DataDictionary("FIX44.xml");
        final Message message = new DefaultMessageFactory()
                .create(dataDictionary.getVersion(), "D");
        message.fromString(expectedMessageString, dataDictionary, false);

... is a useful shorthand (creating/using the correct MessageFactory requiring only a vendor-provided dictionary). It is likely to cause bugs if it was FIX.5.0SP2. If `message` is passed to downstream code, it is highly likely somewhere it will be tested/casted with

    if (message instanceof quickfix.fix50sp2.ExecutionReport) ...

and fail since `quickfix.fix50.ExecutionReport` is binary incompatible.

**Expected behavior**
OPTION1 `getVersion()` to return "FIX.5.0SP2" where appropriate.
OPTION2 add `String getFullVersion()`
OPTION3 Explicitly expose `int getServicePack()`

**system information:**
 - QFJ Version 2.1.1

I'm happy to PR this little change if stakeholders approve a design option.

drekbour:
To preempt a likely question: The issue encountered here is a generic test-harness loads pre-canned QFJ Messages then passes them to version-specific production code.

philipwhiuk:
I like doing option 3 regardless. We'd need to agree what it should return for FIX.4.4 et al.

However whatever we do should ideally also handle EP builds and FIX Latest. 

I'm not too worried about changing `getVersion()` personally.

chrjohn:
Hi @drekbour , where is that `@servicepack` coming from? We don't have it in our data dictionary XMLs. Is it in the original FIX repository that can be downloaded from the FIX Protocol website? We could of course add that information manually to our dictionaries.

BTW, what is the official notation for the versions + servicepack? E.g. you write `FIX5.0.SP2`. In `FixVersions` we have:

https://github.com/quickfix-j/quickfixj/blob/3a9e1046c3a6e98da766c74ddb82cec651bb4221/quickfixj-core/src/main/java/quickfix/FixVersions.java#L36-L37

philipwhiuk:
[...]

PR #579, opened 2022-11-23
`DataDictionary.getVersion()` - return version including service pack 

### LINKED ISSUE #381: Make `DataDictionary.getVersion()` return version including service pack
Follow-up to #258 where `getVersion()` was not changed for a minor QFJ release. But with 3.0.0 we should change this.

vinipx:
fixes #381 

@chrjohn please, review if it makes sense.

Also, should we keep the `setFullVersion`, `getFullVersion`, `setExtensionPackget`, `getExtensionPack` etc?

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/DataDictionary.java
+++ b/quickfixj-core/src/main/java/quickfix/DataDictionary.java
@@ -176,7 +176,7 @@ public class DataDictionary {
     }
 
     /**
-     * Get the FIX major/minor version associated with this dictionary.
+     * Get the FIX major/minor version - including service pack if existing - associated with this dictionary.
      * E.g. FIX.5.0
      *
      * @return the FIX version
--- a/quickfixj-core/src/main/java/quickfix/DataDictionary.java
+++ b/quickfixj-core/src/main/java/quickfix/DataDictionary.java
@@ -94,6 +94,7 @@ public class DataDictionary {
     private boolean checkUnorderedGroupFields = true;
     private boolean allowUnknownMessageFields = false;
     private String beginString;
+    private String servicePackVersion;
     private String fullVersion;
     private String majorVersion;
     private int minorVersion;
@@ -181,7 +182,7 @@ public class DataDictionary {
      * @return the FIX version
      */
     public String getVersion() {
-        return isServicePackVersion() ? beginString + servicePack : beginString;
+        return isServicePackVersion() ? servicePackVersion : beginString;
     }
 
     private boolean isServicePackVersion() {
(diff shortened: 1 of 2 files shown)


quickfixj-core/src/main/java/quickfix/DataDictionary.java
```diff
@@ -882,15 +895,19 @@ private void checkValue(StringField field) throws IncorrectTagValue {
...
     private void checkHasValue(StringField field) {
         if (checkFieldsHaveValues && field.getValue().length() == 0) {
             throw new FieldException(SessionRejectReason.TAG_SPECIFIED_WITHOUT_A_VALUE,
                     field.getField());
         }
     }
 
-    /** Check if group count matches number of groups in **/
+    /**
+     * Check if group count matches number of groups in
```
chrjohn, line 909:
```suggestion
     * Check if group count matches number of groups in message.
```

vinipx:
@chrjohn Please, review. I've introduced `servicePackVersion` to support implementation with minimal code changes. 
Let me know if its appropriate or we should think about better solution.

quickfixj-core/src/main/java/quickfix/DataDictionary.java
```diff
@@ -177,10 +178,15 @@ private void setVersion(String beginString) {
     /**
      * Get the FIX major/minor version associated with this dictionary.
```
vinipx, line 179: potentially here we might add description about servicePack version too

vinipx:
@chrjohn pls, review. I've added the servicepack reference to the method doc.

chrjohn:
Hi @vinipx currently out of the office. I still need to check in more detail what `getVersion()` should return.

vinipx:
> Hi @vinipx currently out of the office. I still need to check in more detail what `getVersion()` should return.

Sure, no rush. Thanks! We can elaborate it better, count on me.

vinipx:
> Hi @vinipx ,
> 
> I thought about this and now am of the opinion that `getVersion()` should not return the full version.
> 
> Reasons for this:
> 
> 1. The `version` in the dictionary represents the `BeginString` from the FIX message. The `BeginString` does not contain service or extension pack.
> 2. The other QuickFIX variants (C++, .NET) also do not return service or extension pack there. Although we cannot be 100% compatible between the variants, I believe we should not change this.
> 
> Sorry, that you did some of the work already and sorry that I did not think this completely through beforehand. :-/
> 
> The original issue #258 also lists option 2 and 3 which are valid so I think we should leave `getVersion()` as it is now.

No problem at all - to be honest, I also found it a bit strange, but decided to proceed as a proposal only. 
Agreed - we can close this PR since there is already servicePack and fullversion available.

Thanks!

chrjohn:
@vinipx 
Thank you for your understanding and contributions! :+1:

PR #582, opened 2022-12-04
Bugfix/issue 576 perf regression

isurudevj:


quickfixj-perf-test/pom.xml
```diff
@@ -0,0 +1,71 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>org.quickfixj</groupId>
+            <artifactId>quickfixj-core</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.quickfixj</groupId>
+            <artifactId>quickfixj-messages-all</artifactId>
+            <version>3.0.0-SNAPSHOT</version>
```
chrjohn, line 30: This should probably be the `project.version` ?!

chrjohn:
Thanks for the PR @isurudevj , would it be possible to add a form of `README.md` or similar to show how the tests can be run?

quickfixj-perf-test/pom.xml
isurudevj: Done, thank you for pointing.

isurudevj:
> Thanks for the PR @isurudevj , would it be possible to add a form of `README.md` or similar to show how the tests can be run?

hi @chrjohn added README.md file.

quickfixj-perf-test/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+```
+$ java -jar target/quickfixj-perf-test.jar -h
+```
+
+#### Guideline for future performance enhancements
+
+1. Check if there is already benchmark available for the code you are planning to optimize.
+2. If there is no benchmark code, first create a benchmark and make pull request.
+3. Your performance improvement can use the benchmark code as the baseline for comparison.
```
chrjohn, line 53: Just a quick question:
So a user wants to change some code and supplies a PR with a benchmark. Then that user should add the current code to the `baseline()` method and the changed code in an additional method (as you have done in your benchmarks below)?
What happens with the benchmark after the PR has been merged?

isurudevj: hi @chrjohn 

I think I use the baseline() work incorrectly here. baseline is always empty method. 
When someone make code changes, its difficult expect that previous code will still be there to create a baseline.
Rather what I suggest is comparison between two branches. Say master and `a-perf-improvement` branch.

I have updated the README.md to make this more clear. Kindly have a look.

busy-spin:
hi @chrjohn 

It was quiet a break between our last conversation. I have fixed the merge conflicts. 
Kindly see if you can review and provide me feedback on new perf testing module.

chrjohn:
Hi @busy-spin , thank you, looks good to me. 👍   Will merge in due course.

Issue #237, opened 2019-10-25
Move webpage to github pages


chrjohn:
I'm always frustrated when quickfixj.org is offline.

We should put the content on github pages.

jgoldrich:
Can the user manual and jira also be moved? https://www.quickfixj.org/usermanual/1.5.3/ and https://www.quickfixj.org/jira/browse/ are both inaccessible now

chrjohn:
Currently no. We'll have to check how they could be imported here.

felipewind:
While the user manual is not migrated to github pages, can I try to improve the documentation at `quickfixj/quickfixj-core/src/main/doc/usermanual` ?

As I did this [small fix](https://github.com/quickfix-j/quickfixj/issues/350) in the README.md, I thought in doing it to on the `quickfixj/quickfixj-core/src/main/doc/usermanual/usage/application.html` too.

### From
```
This allows you to add fields before an adminstrative message before it is sent out.
```

### To
```
This allows you to add fields to an administrative message before it is sent out.
```

chrjohn:
@felipewind of course, please do. Thanks!

PR #615, opened 2023-02-27
Remove `slf4j-jdk14` from `examples` and `distribution` module

chrjohn:
See https://github.com/quickfix-j/quickfixj/discussions/566#discussioncomment-5113958 and following comments.

This should correct the `NullPointerException` during builds with the `-PskipBundlePlugin` profile.


the-thing:
There is a reference to slf4j-jdk14 in quickfixj-core/src/main/doc/usermanual/installation.html which is probably not required anymore.

chrjohn:
@the-thing , thanks for the pointer. I realized that whole document is outdated. I've added a note to https://github.com/quickfix-j/quickfixj/issues/237.

Issue #250, opened 2019-12-29
Fail over retry attempt


MuhammadhAadhil:
Hi All,

I'm really new to this community and this is my 1st comment. We are using QuickFix library for our application. There is a requirement received to handle the fail-over scenario for initiator session as below.

1. If a disconnection experienced for the primary host(SocketAcceptHost), We should try the same connection for a configurable times.

2. If all the retry attempts are failed only we need to start trying to the other hosts(SocketAcceptHost1, SocketAcceptHost2...SocketAcceptHost[N]).
As of my understanding, the above requirement is not available with existing code and it seems doable while introducing new parameter to the class "IoSessionInitiator". Please be kind enough to advise me further on this.

As i mentioned, I'm really new to this community and i'm not sure this is the right place to discuss this. If it is not please someone point me to the right direction. Thanks in advance.

chrjohn:
Maybe you can outline the changes that you want to introduce into the `IoSessionInitiator`? IMHO it would be good to introduce some kind of strategy that can easily be changed without changing QFJ itself.

MuhammadhAadhil:
Hi @chrjohn,
Thanks a lot for the Response. I've tried the change and tested. It is working as expected. As you mentioned Main change is in the class IoSessionInitiator. But couple of classes also need to be modified in order to introduce the New Configuration.
I've Add those classes here. I would be grateful if you could able overview my changes. I've done this change on top of Version 2.0.0.
If this change is acceptable please be kind enough to guide me to add this change to the Latest Code.
[ChangedClasses.zip](https://github.com/quickfix-j/quickfixj/files/4026085/ChangedClasses.zip)

philipwhiuk:
To submit a change so we can modify/merge it:

* Go to: https://github.com/quickfix-j/quickfixj
* Click "Fork"
* Go to your new repository - i.e. https://github.com/MuhammadhAadhil/quickfixj
* Clone the repository
* Checkout a new branch
* Make the changes to the classes (tests and source code)
* Commit the change
* Push the branch
* Go to https://github.com/quickfix-j/quickfixj
* Click New pull request
* Click 'compare across forks'
* Select your repository and branch
* Make sure the changes look like what you did
* Click 'Create pull request'

PetteriPertola:
Hi, was this ever added as a PR / merged into a release version by any chance? I tried to look but I couldn't find anything that suggested it had. 
Thanks!

chrjohn:
[...]

Issue #402, opened 2021-06-24
Failover mechanism is not working when connection is reset by peer on Initiator


esanchezros:
**Describe the bug**
We have an initiator configured with 2 acceptors and it connects to them via an sTunnel service running locally:

    SocketConnectHost=localhost
    SocketConnectPort=44445
    SocketConnectHost1=localhost
    SocketConnectPort1=44446

If the first acceptor is offline, the initiator keeps on trying the first acceptor and never moves on to the failover ones.

**To Reproduce**
- Configure the initiator with 2 acceptors using any IP address and port that accepts tcp traffic. 
- Start the initiator
- The connection is attempted with the first acceptor, a logon attempt is made, the connection times out, the connection is disconnected (see the logs).

    .l.a.c.f.a.Application                   : Adding logon details. Username TargetCompID
    quickfixj.msg.outgoing                   : FIXT.1.1:SenderCompID/TargetCompID->MAIN: 8=FIXT.1.1|9=112|35=A|34=459|49=SenderCompID|50=TargetCompID|52=20210624-09:29:29.788|56=MAIN|98=0|108=30|553=TargetCompID|554=xxxxxxxx|1137=9|10=158|
    quickfixj.event                          : FIXT.1.1:SenderCompID/TargetCompID->MAIN: Initiated logon request
    quickfixj.errorEvent                     : FIXT.1.1:SenderCompID/TargetCompID->MAIN: Disconnecting: Socket exception (localhost/127.0.0.1:44445): java.io.IOException: Connection reset by peer
    l.a.c.f.m.MonitoringSessionStateListener : onDisconnect: session FIXT.1.1:SenderCompID/TargetCompID->MAIN[in:461,out:460] disconnected
    l.a.c.f.m.MonitoringSessionStateListener : onLogout: session FIXT.1.1:SenderCompID/TargetCompID->MAIN[in:461,out:460] logged out
    l.a.c.f.m.MonitoringSessionStateListener : onConnect: session FIXT.1.1:SenderCompID/TargetCompID->MAIN[in:461,out:460] connected
    quickfixj.event                          : FIXT.1.1:SenderCompID/TargetCompID->MAIN: MINA session created: local=/127.0.0.1:49158, class org.apache.mina.transport.socket.nio.NioSocketSession, remote=localhost/127.0.0.1:44445
    .l.a.c.f.a.Application                   : Adding logon details. Username TargetCompID
[...]

PR #624, opened 2023-03-24
https://github.com/quickfix-j/quickfixj/issues/250

MuhammadhAadhil:
Enhancement.

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/Initiator.java
+++ b/quickfixj-core/src/main/java/quickfix/Initiator.java
@@ -122,5 +122,11 @@ public interface Initiator extends Connector {
      * AbstractSocketInitiator.createDynamicSession is called
      */
     String SETTING_DYNAMIC_SESSION = "DynamicSession";
-
+    /**
+     * Initiator setting for reconnect attempts. Only valid when
+     * session connection type is "initiator".
+     *
+     * @see quickfix.SessionFactory#SETTING_CONNECTION_TYPE
+     */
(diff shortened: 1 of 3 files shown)


chrjohn:
Hi @MuhammadhAadhil 

thanks for the PR. In the meantime there was a request which is quite similar #402. See my comment there: https://github.com/quickfix-j/quickfixj/issues/402#issuecomment-877085372

The question is: do you want to give it a try to introduce some kind of interface which enables users to customize the failover behaviour? The default behaviour could still have your changes with the number of retries (so we would have an improvement to the status quo).

Cheers
Chris

MuhammadhAadhil:
Hi @chrjohn ,

It is an interesting change. Let me see on this and keep you posted.

MuhammadhAadhil:
Hi @chrjohn , May i know when this change will be added to the release. If there any more actions needed from my side, please be kind enough to guide me since this is my first change.

chrjohn:
Hi @MuhammadhAadhil , do you wanted to consider the points from https://github.com/quickfix-j/quickfixj/pull/624#issuecomment-1498860723 ?

MuhammadhAadhil:
Hi @chrjohn , Actually i'm not in a position to work on that enhancement at the moment. This year being a little hard on my career. Lets give the feature with next release. We can enhance in future.[#624](https://github.com/quickfix-j/quickfixj/pull/624#issuecomment-1498860723)

PR #631, opened 2023-04-20
Proxy-Authorization header solves Proxy 407

MarcosVillacanas:
HTTP Proxy authentication was failing with a 407 error code. After an in-depth review, we discovered that the header sending was incorrect since the Proxy-Authorization key was missing. This pull request aims to solve this issue by configuring the appropriate header for HttpProxyRequest whenever credentials are provided. Note that the code supports both NTLM and Basic authentication methods. This code fixes our problem due to the correct authentication process from the headers set.

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/mina/ProtocolFactory.java
+++ b/quickfixj-core/src/main/java/quickfix/mina/ProtocolFactory.java
@@ -21,7 +21,11 @@ package quickfix.mina;
 
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
+import java.util.Base64;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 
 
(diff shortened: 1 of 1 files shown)


chrjohn:
Hi @MarcosVillacanas , thanks for the PR 👍 
However, I personally cannot validate if this change is OK (also see #596) so it will probably take some time unless someone with more proxy experience can validate it for me. ;)
BTW, are you able to give me a hand with #596 ? Basically it is about testing QFJ against the current MINA 2.2 release.
Thanks and cheers
Chris

the-thing:
Hi @MarcosVillacanas 

Can you share the proxy software and version you are using? Thanks.

slachiewicz (approved):

chrjohn:
@slachiewicz thanks for the review. Can you confirm that this is working?

chrjohn:
Superseded by #1136

PR #441, opened 2021-11-14
Use MINA 2.2.1-SNAPSHOT

chrjohn:
To test new implementation of `SSLFilter`

See https://issues.apache.org/jira/browse/DIRMINA-1144

PR #635, opened 2023-05-13
Apache MINA 2.2.X update

### LINKED ISSUE #634: Upgrade to latest mina release
quickfixj still uses a mina release that's a couple years old, and is not compatible with the more modern mina releases (e.g. constructor changes in SSLFilter). We should upgrade the mina dependency and fix the code to work with it.

the-thing:
Fixes #634

Attempt to update MINA

I had a look at the branch - https://github.com/quickfix-j/quickfixj/pull/441, but it was slightly behind the master so I decided to do it from scratch on master. Feel free not to merge this pull requires if you want to apply similar changes to the other branch.

I was able to improve and remove some of the SSL code including cipher suites config and the temporary useSNI flag etc. Test cases for SNI were also added.

I'm not sure about the proxy code. The problem with proxy is that there are no tests for it so we don't know if we are breaking things or fixing them. I think it is crucial to add some embedded proxy server at some point and make sure that it works.


quickfixj-core/src/main/java/quickfix/mina/initiator/InitiatorProxyIoHandler.java
```diff
@@ -61,8 +61,8 @@ public void exceptionCaught(IoSession ioSession, Throwable cause) throws Excepti
 
     @Override
     public void proxySessionOpened(IoSession ioSession) throws Exception {
-        if (this.sslFilter != null) {
-            this.sslFilter.initiateHandshake(ioSession);
-        }
+//        if (this.sslFilter != null) {
```
the-thing, line 64: Not sure if this is even required, mina proxy connector should be handling handshake...

quickfixj-core/src/main/java/quickfix/mina/ssl/SSLFilter.java
```diff
@@ -19,70 +19,17 @@
...
+
+import javax.net.ssl.SSLContext;
 
 /**
  * An extended SSL filter based on MINA {@link SslFilter} that applies
  * some adaptations.
  */
 public class SSLFilter extends SslFilter {
 
-    private final Logger log = LoggerFactory.getLogger(getClass());
```
the-thing, line 39: In theory this class is no longer required as there is no custom functionality here.

quickfixj-core/src/main/java/quickfix/mina/ssl/SSLFilter.java
```diff
@@ -19,70 +19,17 @@
...
         super(sslContext);
     }
-
-    /**
-     * Called from {@link SslFilter#onPreAdd} every time a new
-     * session is created which makes it impossible to override enabled cipher
-     * suites configuration.
-     */
-    @Override
-    public void setEnabledCipherSuites(String[] cipherSuites) {
```
the-thing, line 56: This was fixed in MINA so this hack is not required anymore.

quickfixj-core/pom.xml
```diff
@@ -59,7 +59,7 @@
         <dependency>
             <groupId>org.apache.mina</groupId>
             <artifactId>mina-core</artifactId>
-            <version>2.1.6</version>
+            <version>2.2.2-SNAPSHOT</version>
```
the-thing, line 62: I had this artifact in my local repository so the build here will fail.

chrjohn: Should we wait for 2.2.2 release or will 2.2.1 do?

chrjohn:
Thanks for the PR. I'm more comfortable merging your PR than mine since your have more experience in SSL and MINA internals than me.

quickfixj-core/pom.xml
the-thing: 2.2.1 will not work, it doesn't have endpoint identification algorithm for SNI.

quickfixj-core/src/main/java/quickfix/mina/ssl/SSLFilter.java
chrjohn: Feel free to remove it.

chrjohn (commented):
Could you please also update https://github.com/quickfix-j/quickfixj/blob/master/quickfixj-core/src/main/doc/usermanual/usage/configuration.html with the new configuration and remove the old `UseSNI` config? Thank you.

quickfixj-core/src/main/java/quickfix/mina/initiator/InitiatorProxyIoHandler.java
chrjohn: I cannot tell for sure TBH.

quickfixj-core/src/main/java/quickfix/mina/ssl/SSLFilter.java
the-thing: I just did.

quickfixj-core/src/main/java/quickfix/mina/initiator/InitiatorProxyIoHandler.java
the-thing: For the time being I've removed the whole SslFilter from this class. Maybe once there is some working proxy integration we can check if it actually works.

quickfixj-core/src/main/doc/usermanual/usage/configuration.html
```diff
@@ -692,10 +692,10 @@ <H3>QuickFIX Settings</H3>
     <TD><a href="https://docs.oracle.com/javase/8/docs/technotes/guides/security/SunProviders.html">Java default cipher suites</a></TD>
   </TR>
   <TR ALIGN="left" VALIGN="middle">
-    <TD valign="top"> <I>UseSNI</I></TD>
-    <TD>Configures the SSL engine to use Server Name Indication. This option is only useful to initiators.</TD>
-    <TD>Y<BR>N</TD>
-    <TD>N</TD>
+    <TD valign="top"> <I>EndpointIdentificationAlgorithm</I></TD>
```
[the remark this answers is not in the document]
the-thing: https://docs.oracle.com/javase/8/docs/technotes/guides/security/StandardNames.html

I will put it in the docs as well.

masonedmison:
Hi there, is this work slated to be released soon? I am facing a similar issue as described in [this](https://stackoverflow.com/questions/71953301/quickfixj-ssl-socket-failing-on-java-17?noredirect=1#comment127157301_71953301) SO post - I'm currently on JDK11 and would rather have to go back to JDK 8 if at all possible.

chrjohn:
@masonedmison did you test that it works with this change.?

PR #638, opened 2023-05-28
Introduce `SocketConnectTimeout` configuration option

AndreyNudko:
Stopping MINA connector can take up to a minute if there is in-flight TCP handshake which can't complete due to firewall cutting the traffic. What happens in this case is NioSocketConnector background task keeps running until connection timeout expires even after dispose() was called, and default timeout is a minute (see o.a.m.c.s.AbstractIoConnector). This behaviour is surprising: intuitively you expect closing the session to block no more than logon/logout timeout; but currently even with stricter than default settings it's still not safe to do this from a time-sensitive thread.

This patch makes connection timeout configurable on QuickFIX side - which may be a bit of a workaround, but seems like a generally useful socket option. Default value for now matches MINA for backward compatibility. However, it can potentially be lowered to something like 5 seconds (half of the default LogonTimeout), and then reconnection logic would still ensure the session is eventually established.


chrjohn:
Hi @AndreyNudko , thanks for the PR. This is definitely useful.

quickfixj-core/src/main/java/quickfix/mina/initiator/AbstractSocketInitiator.java
```diff
@@ -123,6 +123,11 @@ private void createInitiator(final Session session, final boolean continueInitOn
             throw new ConfigError("Must specify at least one socket address");
         }
 
+        int connectTimeout = 60; // 1 minute by default, matches MINA
+        if (getSettings().isSetting(sessionID, Initiator.SETTING_SOCKET_CONNECT_TIMEOUT)) {
+            connectTimeout = getSettings().getInt(sessionID, Initiator.SETTING_SOCKET_CONNECT_TIMEOUT);
+        }
```
chrjohn, line 129: This could be simplified using SessionSettings.getIntOrDefault(60)


```suggestion
        // 1 minute by default, matches MINA
        int connectTimeout = getSettings().getIntOrDefault(sessionID, Initiator.SETTING_SOCKET_CONNECT_TIMEOUT, 60);
        }
```

quickfixj-core/src/main/java/quickfix/mina/initiator/AbstractSocketInitiator.java
```diff
@@ -174,7 +177,7 @@ && getSettings().isSetting(sessionID, Initiator.SETTING_PROXY_DOMAIN)) {
         ScheduledExecutorService scheduledExecutorService = (scheduledReconnectExecutor != null ? scheduledReconnectExecutor : getScheduledExecutorService());
         try {
             final IoSessionInitiator ioSessionInitiator = new IoSessionInitiator(session,
-                    socketAddresses, localAddress, reconnectingIntervals,
+                    socketAddresses, localAddress, connectTimeout, reconnectingIntervals,
```
chrjohn, line 180: After more thoughts on this, I think it would be a cleaner approach to move the SocketConnectTimeout to the `NetworkingOptions` class. That way we do not need to hand through yet another parameter. What do you think?

chrjohn (changes requested):

quickfixj-core/src/main/java/quickfix/mina/initiator/AbstractSocketInitiator.java
AndreyNudko: yeah, sure
do you want me to make the change? I think you can edit PR as maintainer, this may be just quicker

chrjohn: @AndreyNudko , quickly checked if this is possible. Turned out that `NetworkingOptions` only sets options in MINA's `SocketSessionConfig`, so moving the parameter to `NetworkingOptions` does not work.

chrjohn (approved):

PR #643, opened 2023-06-16
Only retrieve one batch of messages at a time when responding to a large resend request

### LINKED ISSUE #639: The request for messages from storage allows for unbounded memory usage
When a resend request to Infinity (or even just a large capped number is made, this can cause a huge number of messages to be fetched from storage into memory which can cause the application crash.

In addition even if they fit in memory without a fix to #271 this can lead to some fairly ugly behaviour where we continue to attempt to send messages to a session that's already disconnected us.

**To Reproduce**
* Send several thousand messages (volume necessary dependent on application memory profile)
* Instruct the counter-party to perform a resend-request of all these messages

**Expected behavior**
We should fetch the messages in batches and then send them a batch at a time

**System information:**
N/A

**Additional context**
I've got a hot fix for this - will try to tidy it up and submit for review.

This is probably additionally key if we implement #621

philipwhiuk:
Fixes #639

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/Session.java
+++ b/quickfixj-core/src/main/java/quickfix/Session.java
@@ -470,6 +470,8 @@ public class Session implements Closeable {
 
     protected static final Logger LOG = LoggerFactory.getLogger(Session.class);
 
+    private final int MAX_RESEND_BATCH_RETRIEVAL_SIZE = 1000;
+
     Session(Application application, MessageStoreFactory messageStoreFactory, SessionID sessionID,
             DataDictionaryProvider dataDictionaryProvider, SessionSchedule sessionSchedule, LogFactory logFactory,
             MessageFactory messageFactory, int heartbeatInterval) {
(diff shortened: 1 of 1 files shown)


chrjohn:
@philipwhiuk do you happen to have the time to create a unit test for this? So we can ensure that normal resend processing works as expected? Probably a test where we resend more than 1000 messages and check that every message is resent. Something along that lines?
Thank you :)

quickfixj-core/src/main/java/quickfix/Session.java
```diff
@@ -2321,71 +2323,85 @@ private void nextLogon(Message logon) throws FieldNotFound, RejectLogon, Incorre
...
+                    throw e;
+                }
             }
+            for (final String message : messages) {
+                appMessageJustSent = false;
+                final Message msg;
+                try {
+                    // QFJ-626
+                    msg = parseMessage(message);
+                    if (msg.getException() != null) {
```
chrjohn, line 2360: @philipwhiuk I am not sure about this. This will not resend messages which were formerly (without this PR) sent. I mean we could discuss if it is sensible but I feel that this should be a separate change. WDYT?

chrjohn:
@jetbrains-junie could you please create a unit test that checks if the new functionality works as expected? Especially that there are no messages missing between batches.

PR #658, opened 2023-08-08
Consider `NonStopSession` parameter when printing the session schedule

### LINKED ISSUE #648: `NonStopSession` is reported as a daily session
**Describe the bug**
The configuration contains NonStopSession=Y but the log still says:
schedule is daily, 00:00:00-UTC - 00:00:00-UTC
The problem seems to be in the method DefaultSessionSchedule.formatTimeInterval() which does not consider the NonStopSession setting.

msmat0:
Fixes #648


chrjohn:
Thanks for the PR @msmat0

PR #662, opened 2023-08-11
Correct `Resource not found` in `FileUtilTest` (#654) on QFJ 2.3.x

chrjohn:
Same PR as #654 on master

* update FileUtil to resolve new 403 error for requests with User-Agent header
* Updated User-Agent value for FileUtil
* Added check for HttpUrlConnection to changes done in #460


chrjohn:
Same as #654 on master

Issue #271, opened 2020-03-17
Rate-limiting for resend request responses


philipwhiuk:
When a session reconnects mid-session it's possible for it to miss a high volume of messages.

At the application level it's possible for the application to self-throttle the release of messages.

It would be useful if resend-request responses could be optionally-throttled to only release messages at a given rate.

This would ideally take the form of a per-session configuration parameter e.g. `ResendRequestMaxMessagesPerSecond`.

An alternative is to implement it in the form of MINA filters, however this solution can lead to high memory usage because MINA will be forced to store in memory all the resend response messages while they are dequeued.

chrjohn:
I think there can be problems with this approach because of the following reasons:
 - When there is a high volume of messages and you throttle resends I assume it can take a long time until all resent messages have been received.
 - Moreover, the receiving side needs to cache all messages until the ResendRequest has been satisfied.
 - Also you are probably going to generate even more messages while the resend is still running.

So I am asking myself if this won't cause more problems than it will solve. But probably depends on the number of messages you are wanting to resend. What could possibly cause such a high volume of messages?

My approach would be to increase the socket receive buffer on the receiving side. ;)

philipwhiuk:
So we’re already throttling the standard flow in our app code and caching messages. That’s not an issue.

The problem is that while this works fine in the connected case, when they lose connection we send something like ‘HeartbeatInterval*2*ThrottleRate’ messages before we disconnect them.

Then when we reconnect QFJ unleashes the whole flow in a fraction of a second which often causes another disconnect.

As for messages sent while pending resend - I think we handle that in app code (or could do). Anyway twice throttle isn’t much of a problem vs  100*Throttle.

And yeah if I could fix the other side I would!

This might well be a custom one off use case. I just try to upstream as much as I can.

On a very related note - is there any good documentation of the threading model in QFJ?

AbhinavAtul:
Would it be a good idea to customize the apache MINA ```IOSession#writeRequestQueue``` with a ```Deque``` and place response payloads for `resend requests` at the start of the ```Deque``` while regular response payloads at the end? The queue processor will consume from the start prioritizing resend data so that counter party apps catch up with a higher probability rather than being overwhelmed with messages that will can only be processed after resend requests are satisfied.

AbhinavAtul:
[...]

PR #688, opened 2023-11-27
QFJ-271: Prioritize Resend, if configured.

AbhinavAtul:
Prioritize Resend, if configured, allows lagging counter-party app to catch up by scheduling resend responses first over new/pending messages.

### DIFF
--- a/quickfixj-core/src/main/doc/usermanual/usage/configuration.html
+++ b/quickfixj-core/src/main/doc/usermanual/usage/configuration.html
@@ -258,6 +258,14 @@
     <TD> Y<br/>N</TD>
     <TD>N</TD>
   </TR>
+  <TR ALIGN="left" VALIGN="middle">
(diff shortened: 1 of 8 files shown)


chrjohn:
This probably relates to #271 

@philipwhiuk you created #271 but this seems to be the other way round. You wanted to throttle resend request responses, this PR right here wants to prioritize them. So it seems we have two competing concepts here. Any thoughts?

philipwhiuk:
In terms of priority there wasn't really a concern it was more about rate-limiting - I imagine the counter-party would have had a similar issue if the actual flow was high.

My intention is to spend some of next week on re-implementing some throttling code we had in our app. I wanted to rebase our changes first. At that point it seemed sensible to upstream as much as possible to minimise the work we have to do to maintain our internal fork (which is necessary in order to fix production issues ASAP).

PR #721, opened 2023-12-05
added 3rd parameter for configure-pagefile-action

chrjohn:
... to prevent warnings


chrjohn:
Somehow the warning does not go away. I thought it was a warning of the script but seems to be an underlying warning of the Windows function.

https://github.com/al-cheb/configure-pagefile-action/issues/18

PR #767, opened 2024-02-16
increase banzai's OCP a little

spoerri:
allow people to change the most "business-logicy" parts of banzai by extending the Banzai class and one of OrderTableModel, ExecutionTableModel, and BanzaiApplication


chrjohn:
Thanks for the PR 👍

PR #819, opened 2024-06-06
Add Character.class java type for FieldType.CHAR

ishabandi67:
While parsing a rawstring back to a quickfix.Message object, CHAR fields are created as StringField because it String.java is the default javaType.

All other primitive types have their wrapper classes defined except CHAR.


noblemaster:
There should probably also be a unit test with a concrete example to validate the functionality?

PR #828, opened 2024-06-18
Update Maven libs and plugin

chrjohn:
https://github.com/quickfix-j/quickfixj/pull/826#issuecomment-2175711410

Thanks @cstamas


cstamas:
Just FTR: this is not maven but one foundation libs of it: Sisu Inject (JSR330) and Sisu Plexus must be updated in "lockstep", so user must ensure it is _same version_ of those two artifacts present on classpath at any time. 
At runtime of the plugin it is Maven that make sure about this, but during build it is user duty to enforce this.

PR #833, opened 2024-06-26
Mina proxy handshake fix and proxy integration tests

the-thing:
**Changes**

- added custom SSL wrapper to fix proxy connectivity before official MINA patch
- removed unused PROXY protocol type
- added SOCKS proxy server (via Netty)
- added Maven build extension to help build tests using Netty
- added SSL and non-SSL integration tests

### DIFF
--- a/quickfixj-core/pom.xml
+++ b/quickfixj-core/pom.xml
@@ -226,6 +226,7 @@
 		</plugins>
 
 	  <extensions>
+		  <!-- required by Netty cross-platform build -->
 		  <extension>
 			  <groupId>kr.motd.maven</groupId>
 			  <artifactId>os-maven-plugin</artifactId>
--- a/quickfixj-core/pom.xml
+++ b/quickfixj-core/pom.xml
@@ -118,6 +118,12 @@
 			<scope>test</scope>
 		</dependency>
 		<dependency>
+			<groupId>io.netty</groupId>
+			<artifactId>netty-example</artifactId>
+			<version>4.1.111.Final</version>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
 			<groupId>org.apache.mina</groupId>
 			<artifactId>mina-core</artifactId>
 		</dependency>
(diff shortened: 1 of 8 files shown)


chrjohn:
@the-thing , thanks for the PR 👍 
On another note: do you know if #631 could be related?

the-thing:
That's a different problem and not related to the current MINA issue.

https://github.com/apache/mina/blob/2.2.X/mina-core/src/main/java/org/apache/mina/proxy/handlers/http/HttpSmartProxyHandler.java

When MINA's HTTP proxy handler does the handshake, the proxy server can respond with "407 - Proxy Authentication Required" status code which will make MINA's handler to automatically switch and choose HTTP auth method based on the response (no-auth, ntlm, digest and basic) and populate headers. However, I assume this might not be the case for all HTTP proxy servers and some some of them might require you to know the auth method in advance and must ensure that headers are already populated.

I think https://github.com/quickfix-j/quickfixj/pull/631 looks reasonable, but we don't have HTTP tests yet so we don't know if this works. Also one thing about PR - the auth method is chosen based on properties provided, but I have a feeling that HTTP auth method should be configured explicitly.

chrjohn:
It was used for unit tests some time ago and works AFAICT. I only had the suspicion that it sometimes hung the connector during the acceptance test run where the connectorwas started and stopped in quick succession. But I never followed it up.

the-thing:
I was just wondering if this should be removed at some point. I have a feeling that nobody uses it.

chrjohn:
Is it complicating the code much? I don't know if it has performance benefits when connecting intra VM. It was said it has but that was some years ago.

the-thing:
I don't think so, but could be slightly cleaner. Functionally it might have some benefits, but I don't really know how this works. I was more thinking about removing so "SocketConnectProtocol" parameter goes away and less config. Still probably not a target for this JIRA.

chrjohn:
Maybe we could (if it is not already the case in the code) assume TCP as default so one does not explicitly need to define the parameter.

PR #836, opened 2024-06-29
extra logging first draft

### LINKED ISSUE #824: Extra logging for fields when validation is disabled

Currently there are such options as AllowUnknownMsgFields and ValidateUserDefinedFields. We are working in an environment where dictionary changes faster than we would like it to happen. At the same time we do not want messages to be rejected when some new field appears. So we set AllowUnknownMsgFields=Y and ValidateUserDefinedFields=N to allow messages new fields to come through. However, quickfix does not really say when field has been allowed through because of disabled validation. 

I would like quickfix to log fields which would fail validation if these two options would be AllowUnknownMsgFields=N and ValidateUserDefinedFields=Y. Something like
User Defined Field X is detected
Unknown message field X is detected

borkabu:
Fixes #824 
Added new configuration setting FieldValidationLogging. 
When it is Y, then prints log messages instead of failing validation for unknown user defined fields, unknown fields and unknown enum values.

### DIFF
--- a/quickfixj-base/src/main/java/quickfix/DataDictionary.java
+++ b/quickfixj-base/src/main/java/quickfix/DataDictionary.java
@@ -19,6 +19,8 @@
 
 package quickfix;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.w3c.dom.Document;
 import org.w3c.dom.Element;
 import org.w3c.dom.NamedNodeMap;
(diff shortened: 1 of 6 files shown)


chrjohn:
I'm still thinking about a better way to do the logging. As it is now there is no correlation to the actual message and it could be hard to see which message is actually affected. Maybe we could do this via the sessions's log? But still thinking about how this could be accomplished.

borkabu:
Updated documentation

borkabu:
> I'm still thinking about a better way to do the logging. As it is now there is no correlation to the actual message and it could be hard to see which message is actually affected. Maybe we could do this via the sessions's log? But still thinking about how this could be accomplished.

If I understand correctly, DataDictionary is independent of Session. Validation of message can be done, without having existing Session. I see no obstacles of printing message related information. In my case, this logging is to warn about schema being out of sync, so message itself is not really important.

borkabu:
> > If I understand correctly, DataDictionary is independent of Session. Validation of message can be done, without having existing Session. I see no obstacles of printing message related information. In my case, this logging is to warn about schema being out of sync, so message itself is not really important.
> 
> But if you have lots of sessions it's necessary to identify which counter-party is sending out-of-spec messages. Doing it this way you'll have to go back and find the message in the log in a separate log line. That's terrible.
> 
> A nicer way might be a callback:
> 
> Something a little like:
> 
> `onOutOfSpecMessage(Message m, int tag, String value);`
> 
> then at least the app can decide to look it up from the Message.
> 
> You can implement a pure logging handler if that's all you need.

I have kind of warmed up to an idea of callback. I see there would be a default callback with a default logger, when validation is used outside of a session and "integrated" session callback with session logger, when validation is inside session.

PR #851, opened 2024-08-02
Populate Text(58) field when rejectMessageOnUnhandledException=true

mgatny:
refs #639 
closes #667 

If rejectMessageOnUnhandledException=true and an unhandled exception is thrown from the fromCallback(), include the exception message in the Text(58) field of the resulting Reject or BusinessMessageReject.

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/Session.java
+++ b/quickfixj-core/src/main/java/quickfix/Session.java
@@ -1214,13 +1214,13 @@ public class Session implements Closeable {
                 }
                 if (!(MessageUtils.isAdminMessage(msgType))
                         && (sessionBeginString.compareTo(FixVersions.BEGINSTRING_FIX42) >= 0)) {
-                    generateBusinessReject(message, BusinessRejectReason.APPLICATION_NOT_AVAILABLE,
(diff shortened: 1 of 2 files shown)


chrjohn:
@mgatny any comments on my proposal? 
https://github.com/quickfix-j/quickfixj/pull/851#issuecomment-2268801218

Thank you

mgatny:
@chrjohn here's what I'm thinking:

1. If we add Text(58) to the generated Reject at all, we need an additional boolean SessionSetting that enables this behavior, and defaults to "N".  Otherwise, anyone already using `RejectMessageOnUnhandledException` will be surprised that Text(58) automcatically gets set on Rejects.
2. Instead of QF/j adding a prefix to the Text(58), what if we recommend this approach:
    a. Set `RejectMessageOnUnhandledExceptionSetsTextField=Y` (or whatever we call the new boolean setting).
    b. Use a well-known string (i.e. a constant) as the exception message when intentionally throwing an exception to cause a Reject.
    c. In `toAdmin`/`toApp`, compare the Text(58) to your well-known string(s).
    d. If it's a match, leave the Text(58) on the message.
    e. Otherwise remove Text(58) from the message.

This would allow users to get exactly what they want in Text(58), but still give them the ability to filter.  And it will not change any behavior for anyone already using `RejectMessageOnUnhandledException`.

PR #892, opened 2024-10-16
TLSv1.3 test

chrjohn:
Simple test against TLSv1.3...


chrjohn:
When looking around, I realise that in the last weeks sometimes I get this error on all kind of different PR runs, e.g. for #919 or #920 

`SecureSocketTest.testLogonWithBadCertificate:78 no exception thrown`

As far as I am aware it is always JDK17 that fails with that error.

the-thing:
@chrjohn I will have a look.

BTW. I haven't built QuickFIX for a while locally and now I struggle to run unit tests from Intellij IDEA directly. It seems I can build with Maven, but I run out of memory when running single test case such SSLCertificateTest via IDEA. I forgot if there was something to change to run them locally.

the-thing:
`SecureSocketTest.testLogonWithBadCertificate:78` no exception thrown is most likely timeout but `Certificate was authenticated` doesn't look right.

chrjohn:
@the-thing I don't use IntelliJ, for Netbeans the following works for me: https://github.com/quickfix-j/quickfixj#importing-the-project-into-the-ide
I can only imagine that it is trying to compile the whole message and field classes and that gives OOME. Here are the settings that are used when compiling at command line: https://github.com/quickfix-j/quickfixj/blob/master/.mvn/jvm.config

the-thing:
Did you have any luck reproducing this issue locally? I am unable to do so.

I'm running Windows 11 Pro and I specifically downloaded Eclipse Temurin 17.0.13 and all the builds seem to pass. I'm running with `mvn clean test -pl quickfixj-core -PskipAT,skipBundlePlugin` or `clean test -pl quickfixj-core -PskipBundlePlugin`

chrjohn:
> `SecureSocketTest.testLogonWithBadCertificate:78` no exception thrown is most likely timeout but `Certificate was authenticated` doesn't look right.

@the-thing after your changes on #931 I only get `Certificate was authenticated` intermittently and `no exception thrown` seems to be fixed.  Thank you.

the-thing:
@chrjohn This is probably still an issue, because I got it on my branch.

the-thing:
@chrjohn 

What is the preferred (or reliable) way to enable logging from unit tests for the GitHub CI action?

System.out seems to work. Is it possible to make slf4j work? Only QuickFIXj internals seem to be logged - messages, events etc..

chrjohn:
@the-thing hmm, I'm puzzled that slf4j does not work although there is `slf4j-jdk14` in the pom.xml.
https://github.com/quickfix-j/quickfixj/blob/chrjohn-tls13-test/quickfixj-core/src/test/resources/logging.properties seems to look fine... Sorry, no idea really at the moment.

chrjohn:
Superseded by #937

Issue #459, opened 2022-01-06
Move sleepycat code to quickfix-sleepycat


philipwhiuk:
Sleepycat seems unmaintained. Depending on sleepycat involves including an extra database implementation that (most?) people probably aren't using.

To my knowledge this only affects `quickfix.SleepycatStore` and `quickfix.SleepycatStoreFactory` and 

Options:

1) Extract Sleepycat code to `quickfixj-sleepycat` which depends on `quickfixj-core`. Breaking change for Sleepycat Store users, no impact on everyone else.

2) Just remove Sleepycat code. Would mean any Sleepycat Store users have to copy in the code to their own projects and maintain it separately. Completely removes QuickfixJ's dependency on it.

chrjohn:
I think option 1 would be the way to go. But I also don't have any insight into how much Sleepycat is used.
Also see #449 which might be related w.r.t. creating a new submodule.

PR #920, opened 2024-12-31
Remove sleepycat je

philipwhiuk:
No release of Sleepycat JE since 2018 so probably unmaintained.

From a vulnerability management perspective I'd rather storage engines were separate plugins than part of core.

### DIFF
--- a/quickfixj-core/pom.xml
+++ b/quickfixj-core/pom.xml
@@ -133,12 +133,6 @@
 			<version>${slf4j.version}</version>
 		</dependency>
 		<dependency>
-			<groupId>com.sleepycat</groupId>
-			<artifactId>je</artifactId>
-			<version>18.3.12</version>
-			<optional>true</optional>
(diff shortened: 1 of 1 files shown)


philipwhiuk:
I guess #459 wants this to be a separate module. Anyway, I'm just upstreaming what we've done locally at the moment. I can move it to a separate module if that's a blocker.

slachiewicz (approved):

the-thing (changes requested):
There are still references or details related to sleepycat.

quickfixj-core/src/main/doc/usermanual/usage/acceptor_failover.html
quickfixj-core/src/main/doc/usermanual/installation.html
quickfixj-all/pom.xml
quickfixj-distribution/pom.xml

Possibly more... I think it would be good to remove them.

chrjohn:
I'm still not sure whether we should remove this or move it to a separate module.

PR #926, opened 2025-01-09
Backport build changes from master to the 2.3.x branch

gtoison:
Hello @chrjohn, thank you for starting the release of 2.3.2!
I've noticed that the build failed because branch 2.3.x is not up to date with the recent changes, in particular upgrading from Java 15 to Java 17 and from Windows-2016 to Windows latest.

I've backported the latest build from master to the 2.3.x to restore the build, so it also needs the Maven wrapper.
This also entails upgrading Mockito to a version compatible with Java 17.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -1,28 +1,58 @@
 name: Java CI
 
 on:
+  workflow_dispatch:
   push:
   pull_request:
     types: [reopened, opened, synchronize]
+  schedule:
+    - cron:  '30 5 * * *'
 
 jobs:
   test:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-18.04, macOS-latest, windows-2016]
-        java: [8, 11, 15]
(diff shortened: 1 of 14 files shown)


chrjohn:
Hi @gtoison  thanks, I wanted to skip this part since 2.3.2 will be the last 2.3.x release, so I only built and tested locally.
Could you do me a favour and create a separate PR with the workflow changes only and let it use `mvn` instead of `mvnw`? There are changes in the workflow which were introduced by changes in 3.0.0 that are not part of the 2.3.x release.

gtoison:
Sure, I was also wondering if you want to run the build on JDK 17 (instead of JDK 15)
That change requires the mockito upgrade so I'm not sure if it was a good idea for me to include that.

gtoison:
I have added PR #927, it only makes small changes to branch 2.3.x

chrjohn:
Thanks @gtoison , I think this PR is no longer needed with #927 merged. Do you agree?

gtoison:
Superseded by #927

PR #931, opened 2025-01-17
Fixing test assertion for different JVM's

the-thing:
Small SSL test improvements.

**Changes**

- SNI algorithm should be set for two cases
- Oracle JVM gives slightly different error message for `quickfix.mina.ssl.SSLCertificateTest#shouldFailWhenHostnameDoesNotMatchServerName` test case (error message is prepended with exception class name)

expected: "No name matching localhost found"
actual: "java.security.cert.CertificateException: No name matching localhost found"

I think there are different messages in JVM's higher than > 17 which might require changes in the future


chrjohn:
Thank you

PR #934, opened 2025-01-24
Fixing CI logging configuration

the-thing:
Fixing CI test logging.

Seems to work. Have a look at my CI branch build:
https://github.com/the-thing/quickfixj/actions/runs/12946375432/job/36110769380


the-thing:
@chrjohn We could argue if we should change the log level to info on top of that.

chrjohn:
Wow, thanks for that. 
Change to INFO is ok with me.

chrjohn (approved):

PR #937, opened 2025-01-28
Fixing test assertions for TLSv1.3

the-thing:
This will fix the SSL test suite for TLS v1.3.

**Changes**
 - switching cipher suite and TLS version for `SSLCertificateTest` (see https://github.com/quickfix-j/quickfixj/pull/892)
 - refactored some SSL util methods and extracted them to a separate class
 - modified `assertNotAuthenticated` method
 
The problem with  authentication and TLS v1.3 seems to be with MacOS Java version 8 and 11 as it fails continuously with this combination (works fine with MacOS Java 17). I've seen it failing occasionally on Ubuntu.

The authentication actually works fine, but the assertions were not correct. Seeing `Certificate was authenticated` was misleading since the previous checks - `assertSslExceptionThrown`  and `assertNotLoggedOn` were successful, which means that SSL session was actually NOT  established with invalid certificates as expected.

Even when authentication is enabled and SSL handshake fails, the peer certificates (and peer principal) are set for `javax.net.ssl.SSLSession#getPeerCertificates` (there is a note in later versions of Java > 8 in JavaDoc for `getPeerCertificates` saying _Note: The returned value may not be a valid certificate chain and should not be relied on for trust decisions._). This means that we can't rely on the value there. 

It seems that peer certificates and peer princpial are set for all SSL sessions that require authentication regardless if the check was successful or not. Only when auth is disabled e.g by setting `javax.net.ssl.SSLParameters#needClientAuth` to false these certificate chains and peers can be expected to be empty.

For authentication we have to rely on MINA removing the `org.apache.mina.filter.ssl.SslFilter` from `org.apache.mina.core.session.IoSession` filter chain.

Putting a few second delay after initiator start solves the problem as it is enough for MINA to tear down the SSLFilter - that's what I did, but with busy spin loop. I also kept the old validation code for peers that don't require auth.

### DIFF
--- a/quickfixj-core/src/test/java/quickfix/mina/ssl/SSLCertificateTest.java
+++ b/quickfixj-core/src/test/java/quickfix/mina/ssl/SSLCertificateTest.java
@@ -43,6 +43,8 @@ import quickfix.ThreadedSocketInitiator;
 import quickfix.mina.ProtocolFactory;
 import quickfix.mina.SessionConnector;
 
+import javax.net.ssl.SSLEngine;
+import javax.net.ssl.SSLEngineResult;
 import javax.net.ssl.SSLHandshakeException;
 import javax.net.ssl.SSLSession;
 import java.math.BigInteger;
@@ -757,8 +759,9 @@ public class SSLCertificateTest {
             Session session = findSession(sessionID);
             SSLSession sslSession = SSLUtil.findSSLSession(session);
 
-            if (sslSession == null)
+            if (sslSession == null) {
                 return;
+            }
 
             if (authOn) {
                 // when authentication is on, the SSL session maybe still be alive (invalid) for some time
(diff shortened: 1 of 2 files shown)


the-thing:
We could also add `org.junit.runners.Parameterized` runner and make this test suite run for both TLSv1.2 and TLSv1.3.

chrjohn:
> We could also add `org.junit.runners.Parameterized` runner and make this test suite run for both TLSv1.2 and TLSv1.3.
> 

I'll leave that up to you. Thank you

the-thing:
It is ready, but some other test have failed.

`2025-01-28T13:33:55.1936450Z [ERROR]   Run 933: AcceptanceTestSuite$AcceptanceTest.run:77 connection to server failed:`

chrjohn:
> I see the acceptance test suite failing occasionally here and there.

I'm pretty sure it does not have to do with your changes. One of the acceptance tests fails from time to time, did not have time to follow up.

PR #950, opened 2025-03-09
Add nightly SNAPSHOT release to GitHub Packages

### LINKED ISSUE #769: Publish nightly snapshots
QuickFixJ isn't released that often but there are often new commits that may have already fixed any issue I might currently have.

I'm experiencing the same as [this stackoverflow issue](https://stackoverflow.com/questions/71953301/quickfixj-ssl-socket-failing-on-java-17) and I'd like to test unreleased versions of QuickFixJ to see if they fix my problem.

I'd like to see snapshots posted to a public repo whenever a new PR has been merged.  That means I can easily test to see if my issue has been fixed.

I could build from source but that does involve a number of steps.  It also doesn't allow me to easily switch between a number of snapshots to check which fixes my issue.

Another option would be more frequent proper releases of QuickFixJ but I imagine that's a lot of overhead for you to support.

krisbanas:
Fixes #769 

I came across the discussions under https://github.com/quickfix-j/quickfixj/discussions/941 and the issue https://github.com/quickfix-j/quickfixj/issues/769 and had a look at the snapshot releases.

### Results

- A draft of a GitHub Actions pipeline that **builds with JDK 17** and **publishes to GitHub Packages**
  - The pipeline is set to run on cron (1AM) and on new commits to master
- I tested the publishing to Packages [on my fork](https://github.com/krisbanas?tab=packages&repo_name=quickfixj). .
- The GitHub Packages repo indexes the artifacts as `3.0.0-SNAPSHOT` and keeps them timestamped under the hood.
- I set up a test Gradle project to test resolving the snapshot, it downloads the most recent artifact (upon forcing it to `--refresh-dependencies`).

### Questions/Issues

- The pipeline uses `altDeploymentRepository` flag ([see documentation here](https://maven.apache.org/plugins/maven-deploy-plugin/deploy-mojo.html#altDeploymentRepository)) to override the repositories defined in `pom.xml` instead of editing the file itself. The snapshot repo can instead be adjusted in the POM file.
- GitHub Packages seem not to implement any retention policy for snapshots. It means producing new artifacts every day with no clean-up of older ones. GitHub hints at using a [maintenance pipeline](https://github.com/marketplace/actions/delete-package-versions) to handle it.
- Compared to what I've found on the maven central repo, the signatures (`.asc`) files are missing in the Packages.
- Another issue is **lack of support for unauthenticated dependency fetching from the Packages**. You need a PAT to authenticate and GitHub is not likely to change it ([documentation](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-apache-maven-registry#authenticating-to-github-packages)) ([community discussion](https://github.com/orgs/community/discussions/26634)).
- Looking at the CI pipeline, the `./mvnw deploy...` command in the snapshot release pipeline likely misses some parameters.
- I will document the process in README if you decide to go with this solution.

Let me know what you think.

### DIFF
--- a/.github/workflows/nightly-build.yml
+++ b/.github/workflows/nightly-build.yml
@@ -17,7 +17,7 @@ jobs:
         uses: actions/setup-java@v4
         with:
           distribution: 'temurin'
-          java-version: 17
+          java-version: 21
           cache: 'maven'
 
       - name: Build and Publish SNAPSHOT
--- a/.github/workflows/nightly-build.yml
+++ b/.github/workflows/nightly-build.yml
@@ -21,6 +21,7 @@ jobs:
           cache: 'maven'
 
       - name: Build and Publish SNAPSHOT
-        run: ./mvnw deploy -B -V -DaltDeploymentRepository=github::default::https://maven.pkg.github.com/${{ github.repository }}
+        run: ./mvnw deploy -B -V -DaltDeploymentRepository=github::default::https://maven.pkg.github.com/${{ github.repository }} -D"maven.javadoc.skip"="true" -D"java.util.logging.config.file"="${{github.workspace}}/quickfixj-core/src/test/resources/logging.properties" -D"http.keepAlive"="false" -D"maven.wagon.http.pool"="false" -D"maven.wagon.httpconnectionManager.ttlSeconds"="120"
+
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/nightly-build.yml
+++ b/.github/workflows/nightly-build.yml
@@ -1,5 +1,8 @@
 name: Java SNAPSHOT Release
 
+permissions:
+  packages: write
+
 on:
   workflow_dispatch:
   push:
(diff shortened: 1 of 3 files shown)


krisbanas:
Hi @chrjohn, did you find a moment to have a look?

chrjohn:
Hi @krisbanas , no unfortunately not. Totally swamped in regular work work. :-/

chrjohn:
Hi @krisbanas , thanks for your work. 👍  Please go ahead with the mentioned documentation.
It improves the current situation a lot so I'm happy to go with this. We can adapt and care for the questions/issues along the way...

krisbanas:
Thanks for the feedback. I added the documentation - please have a look!

chrjohn:
Just a quick question: we need to be compatible to JDK8. Why does the build use JDK17? Was this intentional?
I guess the `maven.compiler.target` should do the trick to produce JDK8 compatible classes. But why then not choose JDK21 for compilation?

krisbanas:
I checked and the created files are compatible with JDK8. I chose JDK17 as this is the most recent version that's used [in the CI pipeline](https://github.com/quickfix-j/quickfixj/blob/master/.github/workflows/maven.yml), I didn't give it much though :)

Should I change it to Java 21?

chrjohn:
> Should I change it to Java 21?

Yes, please do.

chrjohn:
And can/should we also put this info here:

https://github.com/quickfix-j/quickfixj/blob/bf6a39c11a2748310f1ea8c8810334ca4b76ead3/pom.xml#L606-L615 ???

The sonatype snapshot repo is not useful for our purpose since we can only upload one snapshot with the same version.

.github/workflows/nightly-build.yml
```diff
@@ -0,0 +1,27 @@
...
+jobs:
+  release-snapshot:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up JDK
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'temurin'
+          java-version: 17
```
krisbanas, line 20:
```suggestion
          java-version: 21
```

krisbanas:
Alright, Java version set to 21 and the `<distributionManagement>` now declares GitHub Packages as the repository.

I tested the change on my profile's GitHub Packages and worked OK.

Let me know what you think!

chrjohn:
> the `<distributionManagement>` now declares GitHub Packages as the repository.

Thank you.
Sorry, I didn't make myself clear. Only the snapshot repo should be changed, for releases we will still use Sonatype.

chrjohn:
Looking good @krisbanas , thank you very much.

PR #957, opened 2025-04-07
Fixed Replace Order UI functionality in Banzai Example

frankk-d:
When I was testing the example, Banzai, I found an issue where it wouldn't update the new values in the UI when it receives a REPLACE ExecutionReport. It would retain its previous value. The new version handles and updates the order with the updated value.

### DIFF
--- a/quickfixj-examples/banzai/src/main/java/quickfix/examples/banzai/BanzaiApplication.java
+++ b/quickfixj-examples/banzai/src/main/java/quickfix/examples/banzai/BanzaiApplication.java
@@ -223,17 +223,34 @@ public class BanzaiApplication implements Application {
         }
 
         OrdStatus ordStatus = (OrdStatus) message.getField(new OrdStatus());
-
+            //handles reject
         if (ordStatus.valueEquals(OrdStatus.REJECTED)) {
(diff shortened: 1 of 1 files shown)


quickfixj-examples/banzai/src/main/java/quickfix/examples/banzai/BanzaiApplication.java
```diff
@@ -235,6 +234,29 @@ private void executionReport(Message message, SessionID sessionID) throws FieldN
             if (order.isNew()) {
                 order.setNew(false);
             }
+        } else if (ordStatus.valueEquals(OrdStatus.REPLACED)) {
+            OrderQty orderQty = new OrderQty();
+            message.getField(orderQty);
+            order.setQuantity((int)orderQty.getValue());
```
chrjohn, line 240: Quantities can also take decimal values.

chrjohn: You are right, I totally forgot about that simplification in the example app.

chrjohn:
Thank you for the PR! 👍

PR #988, opened 2025-06-16
Improve disconnect reason when `ResetOnLogon` is not honoured by counter party

### LINKED ISSUE #960: "Received logon response before sending request" is unclear with ResetOnLogon=Y
**Describe the bug**
With the session set to ResetOnLogon = Y, if the counter-party does not honour this, the logging is unclear what's going on:

**To Reproduce**

FIX Logs (assume valid checksum and lengths)

```
8=FIX.4.4|9=100|35=A|49=OUT|56=TEST_CLIENT_UAT|34=1|52=20250415-11:34:17.247|50=A1|98=0|108=30|141=Y|10=156|
8=FIX.4.4|9=90|35=A|49=TEST_CLIENT_UAT|56=OUT|34=599|52=20250415-11:34:17.371|108=30|98=0|10=166|
```

**Expected behavior**

* The logs should say something like `Disconnecting: Expected Logon response to have reset sequence numbers in response to ResetSeqNumFlag`

**Actual behavior**

* `Disconnecting: Received logon response before sending request`

**Versions**

Present in https://github.com/quickfix-j/quickfixj/blob/cc7e3f4f965511a2adf01d9fc2d6d5cb8bb201c4/quickfixj-core/src/main/java/quickfix/Session.java#L2280

Ramonster1:
This fixes #960 by clarifying the reason for sequence number in response not being 1 when resetOnLogon is configured to true.

As far as I could tell, it isn't necessary to add/amend a unit test for this fix as the disconnect is already tested by SessionTest.testInferResetSeqNumAcceptedWithNonInitialSequenceNumber()

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -5,8 +5,8 @@
     "ghcr.io/devcontainers/features/java:1": {
       "version": "none",
       "installMaven": "true",
-      "mavenVersion": "3.8.6",
-      "installGradle": "true"
+      "mavenVersion": "3.9.9",
+      "installGradle": "false"
     }
   }
 }
--- a/.mvn/wrapper/maven-wrapper.properties
+++ b/.mvn/wrapper/maven-wrapper.properties
@@ -16,5 +16,5 @@
 # under the License.
 #
 # NOTE: maven version should also be changed in pom.xml
-distributionUrl=https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.9.9/apache-maven-3.9.9-bin.zip
+distributionUrl=https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.9.10/apache-maven-3.9.10-bin.zip
 wrapperUrl=https://repo.maven.apache.org/maven2/org/apache/maven/wrapper/maven-wrapper/3.3.2/maven-wrapper-3.3.2.jar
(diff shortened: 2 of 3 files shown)


quickfixj-core/src/main/java/quickfix/Session.java
```diff
@@ -2277,7 +2277,11 @@ private void nextLogon(Message logon) throws FieldNotFound, RejectLogon, Incorre
 
         // Check for proper sequence reset response
         if (state.isResetSent() && !state.isResetReceived()) {
-            disconnect("Received logon response before sending request", true);
+            if (resetOnLogon) {
+                disconnect("Expected Logon response to have reset sequence numbers in response to ResetSeqNumFlag", true);
+            } else {
+                disconnect("Received Logon response before sending request", true);
+            }
```
chrjohn, line 2284: Hi @Ramonster1 , IMHO we can remove this because that case is handled in https://github.com/quickfix-j/quickfixj/blob/cdd34d73e16eebc495c7fa82cc0ebfdb0a1fb481/quickfixj-core/src/main/java/quickfix/Session.java#L2216-L2219

chrjohn (changes requested):

quickfixj-core/src/main/java/quickfix/Session.java
chrjohn: On another note, I am asking myself why we don't `return` after calling `disconnect()`... in some places in the code we do it, but not in every case.

Ramonster1: Thanks @chrjohn, that makes sense.

chrjohn (approved):

Issue #801, opened 2024-04-29
Generated messages should return header from their fix version


Svanar:
**Is your feature request related to a problem? Please describe.**
Generated messages should have overridden `getHeader()` method to return header that corresponds to used protocol version.

**Describe the solution you'd like**
In order to set header fields I would like to have returned header that if specific to fix version that message comes from.
I presented that with code example in additional context.

What I think would be nice is to make method `getHeader` return `quickfix.fix44.Message.Header` so building messages can be done without any casting, making it more fluent.
What is more changing behavior to one I described would make it easier to work with for new devs, because IDE would hint available methods.

**Describe alternatives you've considered**
Casting `quickfix.Message.Header` to actual header, for example `quickfix.fix44.Message.Header`

**Additional context**
Example code pattern that I currently use, but it is a bit tedious to write it each time I add fields in header.
```java
import quickfix.Message;
import quickfix.field.TargetSubID;
import quickfix.fix44.QuoteRequest;

class Main {

    public static void main(String[] args) {
        QuoteRequest quoteRequest = new QuoteRequest();

        Message.Header header = quoteRequest.getHeader();
        // header.set(new TargetSubID("subId")); // Does not work

        // casting is needed
        quickfix.fix44.Message.Header fix44Header = (quickfix.fix44.Message.Header) header;
        fix44Header.set(new TargetSubID("subId"));

    }

}
```

chrjohn:
Thanks for opening this issue. 

If you're not tied to using the type safe methods you could also use e.g.:

https://github.com/quickfix-j/quickfixj/blob/3366b36fdd65ec5af83d33e4a8335db4ffdcff2b/quickfixj-base/src/main/java/quickfix/FieldMap.java#L157

Svanar:
That was one of my attempts and while it is working I think that having it type safe makes it great and generated classes are really good with that hence my idea.

PR #1048, opened 2025-09-24
`quickfix.Message#getHeader` subclasses to return Header with respective FIX version header

the-thing:
Fixes https://github.com/quickfix-j/quickfixj/issues/801

The change is backwards compatible since extended message header e.g. `quickfix.fix44.Message.Header` extends `quickfix.Message.Header`.

### DIFF
--- a/quickfixj-base/src/main/java/quickfix/Message.java
+++ b/quickfixj-base/src/main/java/quickfix/Message.java
@@ -432,7 +432,7 @@ public class Message extends FieldMap {
         }
     }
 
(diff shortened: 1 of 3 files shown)


the-thing:
@chrjohn 

Good point. I don't think it will work not. Let me see if I can fix it.

chrjohn:
@the-thing , I just published the orchestra tools 1.0.3.
Thanks in advance!

chrjohn:
@the-thing thank you 👍

Issue #928, opened 2025-01-14
Too many WriteToClosedSessionException when upgrade to quickfixj-2.3.2


leonchen83:
**Describe the bug**

```
org.apache.mina.core.write.WriteToClosedSessionException: null
	at org.apache.mina.core.polling.AbstractPollingIoProcessor$Processor.clearWriteRequestQueue(AbstractPollingIoProcessor.java:1192)
	at org.apache.mina.core.polling.AbstractPollingIoProcessor$Processor.removeNow(AbstractPollingIoProcessor.java:1153)
	at org.apache.mina.core.polling.AbstractPollingIoProcessor$Processor.removeSessions(AbstractPollingIoProcessor.java:864)
	at org.apache.mina.core.polling.AbstractPollingIoProcessor$Processor.run(AbstractPollingIoProcessor.java:694)
	at org.apache.mina.util.NamePreservingRunnable.run(NamePreservingRunnable.java:64)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1583)
```

**To Reproduce**
Steps to reproduce the behavior.
Or even better, a unit test or reproducer.

**Expected behavior**

No Exception

**system information:**
 - OS: Linux 4.18.0-193.el8.x86_64 #1 SMP Fri May 8 10:59:10 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
 - Java version Java21
 - QFJ Version 2.3.2

**Additional context**

Could this issue be related to the changes introduced in [543](https://github.com/quickfix-j/quickfixj/pull/543)?

chrjohn:
When exactly does this occur?

leonchen83:
In our testing environment, we need to connect to a test LP. If the connection attempt receives no response, this exception occurs every time we reconnect. This issue did not exist in the previous quickfixj version.

chrjohn:
Could you show some log output where the connection attempts are shown in conjunction with the exception?

leonchen83:
```
250114.121008.706: Initiated logon request
250114.121019.705: Disconnecting: Timed out waiting for logon response
[...]

PR #1056, opened 2025-10-03
Remove unnecessary `IoSession.close()` for session that is already closed

the-thing:
Close is definitely not required here. At this stage the session is already closed. Close is idempotent so it has no effect.

Call stack.

- `quickfix.mina.AbstractIoHandler#sessionClosed`
- `quickfix.mina.initiator.InitiatorProxyIoHandler#sessionClosed`
- `org.apache.mina.core.filterchain.DefaultIoFilterChain.TailFilter#sessionClosed`

I created additional logging on a different branch to make sure that `org.apache.mina.core.session.IoSession#isClosing` is true every time we get here.

```
@Override
public void sessionClosed(IoSession ioSession) {
    try {
        Session quickFixSession = findQFSession(ioSession);
        if (quickFixSession != null) {
            eventHandlingStrategy.onMessage(quickFixSession, EventHandlingStrategy.END_OF_STREAM);
        }
    } catch (Exception e) {
        throw e;
    } finally {
        Object oldAttribute = ioSession.removeAttribute(SessionConnector.QF_SESSION);
        log.info("test_close [closing={},oldAttribute={}]", ioSession.isClosing(), oldAttribute);
        ioSession.closeOnFlush();
    }
}
```

See build logs in my fork - https://github.com/the-thing/quickfixj/actions/runs/18196717686

Also code https://github.com/apache/mina/blob/04a9b94a3705728f0c805444bf0d54d770fa4b47/mina-core/src/main/java/org/apache/mina/core/session/AbstractIoSession.java#L332

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/mina/AbstractIoHandler.java
+++ b/quickfixj-core/src/main/java/quickfix/mina/AbstractIoHandler.java
@@ -127,7 +127,6 @@ public abstract class AbstractIoHandler extends IoHandlerAdapter {
             throw e;
         } finally {
             ioSession.removeAttribute(SessionConnector.QF_SESSION);
-            ioSession.closeOnFlush();
         }
     }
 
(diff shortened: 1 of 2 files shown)


chrjohn:
Is this related to #928 ? But probably just coincidence.

the-thing:
Related in the sense that I found while looking at it. It will not fix anything except one less unnecessary line of code.

PR #1057, opened 2025-10-03
Improvements

abhishekraha:
Changes : 

1. Removed redundant directory existence check as mkdir command does it when executed.
2. Provided class type to parameterized class to remove unnecessary type casting 
3. Removed instances of unnecessary object creation
4. IntelliJ remove/regroup-imports and indentation
5. grammatical fixes

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/FileLog.java
+++ b/quickfixj-core/src/main/java/quickfix/FileLog.java
@@ -19,16 +19,10 @@
 
 package quickfix;
 
-import java.io.File;
-import java.io.FileNotFoundException;
-import java.io.FileOutputStream;
-import java.io.IOException;
-import java.io.OutputStream;
-import java.io.UnsupportedEncodingException;
-
 import org.quickfixj.CharsetSupport;
-
 import quickfix.field.converter.UtcTimestampConverter;
+
+import java.io.*;
 
 /**
  * File log implementation. THIS CLASS IS PUBLIC ONLY TO MAINTAIN COMPATIBILITY
@@ -69,9 +63,8 @@ public class FileLog extends AbstractLog {
         eventFileName = prefix + "event.log";
 
         File directory = new File(messagesFileName).getParentFile();
-        if (!directory.exists()) {
-            directory.mkdirs();
-        }
+        directory.mkdirs();
+
 
         this.includeMillis = includeMillis;
         this.includeTimestampForMessages = includeTimestampForMessages;
@@ -94,7 +87,7 @@ public class FileLog extends AbstractLog {
 
     private void writeMessage(FileOutputStream stream, Object lock, String message, boolean forceTimestamp) {
         try {
-            synchronized(lock) {
+            synchronized (lock) {
                 if (forceTimestamp || includeTimestampForMessages) {
                     writeTimeStamp(stream);
                 }
--- a/quickfixj-core/src/main/java/quickfix/MessageCracker.java
+++ b/quickfixj-core/src/main/java/quickfix/MessageCracker.java
@@ -41,7 +41,7 @@ public class MessageCracker {
     public @interface Handler {
     }
 
-    public class RedundantHandlerException extends RuntimeException {
+    public static class RedundantHandlerException extends RuntimeException {
         private final Class<?> messageClass;
         private final Method originalMethod;
         private final Method redundantMethod;
@@ -55,7 +55,7 @@ public class MessageCracker {
 
         @Override
         public String toString() {
-            return "Duplicate handler method for " + messageClass + ", orginal method is "
+            return "Duplicate handler method for " + messageClass + ", original method is "
                     + originalMethod + ", redundant method is " + redundantMethod;
         }
     }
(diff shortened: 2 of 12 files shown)


quickfixj-core/src/main/java/quickfix/FileLog.java
```diff
@@ -69,9 +63,8 @@ public class FileLog extends AbstractLog {
         eventFileName = prefix + "event.log";
 
         File directory = new File(messagesFileName).getParentFile();
-        if (!directory.exists()) {
```
abhishekraha, line 72: redundant check as `directory.mkdir()` has exists check inbuilt

quickfixj-core/src/main/java/quickfix/MessageCracker.java
```diff
@@ -41,7 +41,7 @@ public class MessageCracker {
     public @interface Handler {
     }
 
-    public class RedundantHandlerException extends RuntimeException {
+    public static class RedundantHandlerException extends RuntimeException {
```
abhishekraha, line 44: non-static inner class will unnecessarily carry a reference to the outer class

quickfixj-core/src/main/java/quickfix/MessageCracker.java
```diff
@@ -97,7 +97,7 @@ private boolean matchesConventionOrAnnotation(Method method) {
         return method.getName().equals("onMessage") || method.isAnnotationPresent(Handler.class);
     }
 
-    private class Invoker {
+    private static class Invoker {
```
abhishekraha, line 100: non-static inner class will unnecessarily carry a reference to the outer class

quickfixj-core/src/main/java/quickfix/SleepycatStore.java
```diff
@@ -211,7 +202,7 @@ public synchronized void get(int startSequence, int endSequence, Collection<Stri
         Cursor cursor = null;
         try {
             DatabaseEntry sequenceKey = new DatabaseEntry();
-            EntryBinding sequenceBinding = TupleBinding.getPrimitiveBinding(Integer.class);
+            EntryBinding<Integer> sequenceBinding = TupleBinding.getPrimitiveBinding(Integer.class);
```
abhishekraha, line 205: Making variable type safe at source instead of doing type-casting everytime it is used

quickfixj-core/src/main/java/quickfix/mina/initiator/AbstractSocketInitiator.java
```diff
@@ -263,20 +242,18 @@ private int[] getReconnectIntervalInSeconds(SessionID sessionID) throws ConfigEr
...
     }
 
     private SocketAddress[] getSocketAddresses(SessionID sessionID) throws ConfigError {
         final SessionSettings settings = getSettings();
         final ArrayList<SocketAddress> addresses = new ArrayList<>();
-        for (int index = 0;; index++) {
+        for (int index = 0; ; index++) {
             try {
-                final String protocolKey = Initiator.SETTING_SOCKET_CONNECT_PROTOCOL
-                        + (index == 0 ? "" : Integer.toString(index));
```
abhishekraha, line 275: every `Integer.toString(index)` will create a new object.
reducing innecessary garbage creation

quickfixj-core/src/main/java/quickfix/mina/ssl/InitiatorSslFilter.java
```diff
@@ -53,7 +53,7 @@ protected SSLEngine createEngine(IoSession session, InetSocketAddress addr) {
 
         if (sniHostName != null) {
             SSLParameters sslParameters = sslEngine.getSSLParameters();
-            sslParameters.setServerNames(Arrays.asList(new SNIHostName(sniHostName)));
+            sslParameters.setServerNames(Collections.singletonList(new SNIHostName(sniHostName)));
```
abhishekraha, line 56: `Arrays.asList(new SNIHostName(sniHostName))` - creates 2 objects
1. new SNIHostName(sniHostName)
2. Array object when we do Arrays.asList

`Collections.singletonList(new SNIHostName(sniHostName))` - creates 1 object of SNIHostName and wraps it in the Collection obeject

chrjohn: Will check the other points, but this code is called typically only once. Good point, though.

chrjohn:
> out of curiosity - is the intention that we only import the classes that are required and not the whole package because importing is a compile time task and I don't think it has any latency implications.

The intention is simply that one uses IntelliJ, another uses Eclipse, a third Netbeans and either has its own taste of how imports should be organized (unless you configure it the same for all IDEs).

chrjohn:
@abhishekraha thanks for your PR! 👍

quickfixj-core/src/main/java/quickfix/mina/acceptor/AbstractSocketAcceptor.java
```diff
@@ -123,15 +123,15 @@ protected synchronized void startAcceptingConnections() throws ConfigError {
                 if (continueInitOnError) {
                     log.warn("error during session initialization for session(s) {}, continuing...", socketDescriptor.getAcceptedSessions().keySet(), e);
                 } else {
-                    log.error("Cannot start acceptor session for {}, error: {}", address, e);
+                    log.error("Cannot start acceptor session for {}, error: ", address, e);
```
chrjohn:
```suggestion
                    log.error("Cannot start acceptor session for {}", address, e);
```

PR #1100, opened 2025-12-16
`JdbcStore` concurrency bug fix

the-thing:
Fixes https://github.com/quickfix-j/quickfixj/issues/357

Changes

- sequence numbers' updates happen in their own individual update queries which will prevent overriding each other sequence numbers with stale values

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/JdbcStore.java
+++ b/quickfixj-core/src/main/java/quickfix/JdbcStore.java
@@ -299,12 +299,12 @@ class JdbcStore implements MessageStore {
 
     public void setNextSenderMsgSeqNum(int next) throws IOException {
         cache.setNextSenderMsgSeqNum(next);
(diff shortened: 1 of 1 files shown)


quickfixj-core/src/main/java/quickfix/JdbcStore.java
```diff
@@ -293,23 +299,39 @@ public boolean set(int sequence, String message) throws IOException {
...
     }
 
     public void setNextTargetMsgSeqNum(int next) throws IOException {
         cache.setNextTargetMsgSeqNum(next);
-        storeSequenceNumbers();
+        storeIncomingSequenceNumbers();
     }
 
-    private void storeSequenceNumbers() throws IOException {
+    private void storeIncomingSequenceNumbers() throws IOException {
```
chrjohn: Just a minor thing but would IMHO be cleaner (with less duplication): keep one method storeSequenceNumbers() but pass in the parameters for the seqnum and the SQL?

PR #1142, opened 2026-02-16
Backport #384: Always re-resolve `SocketAddress` on reconnect

mjkam:
Backport of #384 to the `QFJ_2_3_x` release branch.

### Bug

In all released versions (up to v2.3.2), DNS is **not** re-resolved after Logout, Connection Reset, or any session-level disconnect. `unresolveCurrentSocketAddress()` is only called from `handleConnectException()`, which only runs on TCP connect failures (e.g. Connection Refused). When an established session disconnects, the resolved IP in `socketAddresses[]` is reused as-is.

This means if the counterparty changes their DNS entry, the client keeps connecting to the old IP indefinitely — especially problematic when the old IP still accepts TCP connections but no longer runs the FIX service (Logon timeout → reconnect → same old IP → infinite loop).

### Why there is no performance concern

A concern was [raised](https://github.com/quickfix-j/quickfixj/pull/384#issuecomment-841723882) that re-resolving DNS on every connect could be excessive. However, `connect()` is only called when `shouldReconnect()` returns `true`:

```java
private boolean shouldReconnect() {
    return (ioSession == null || !ioSession.isConnected()) && ...;
}
```

`getNextSocketAddress()` is **never called while a session is connected**. "Every connect" and "only after disconnect" are the same thing. There is no scenario where this change causes unnecessary DNS lookups.

### Why this should be backported

- The fix has been on `master` since May 2021 (4 years) with no reported issues.
- A user [reported the same problem in production with AWS](https://github.com/quickfix-j/quickfixj/pull/384#issuecomment-1160660583) in June 2022, and was told no release was planned.
- Cloud environments (AWS ELB/NLB, DNS failover, DR) rely on DNS changes being picked up on reconnect. This is increasingly the norm, not the exception.

### DIFF
--- a/quickfixj-core/src/test/java/quickfix/mina/ssl/SSLCertificateTest.java
+++ b/quickfixj-core/src/test/java/quickfix/mina/ssl/SSLCertificateTest.java
@@ -80,9 +80,24 @@ public class SSLCertificateTest {
 
     private static final String LOCALHOST_ALIAS = "localhost-quickfixj";
 
+    // Cipher suites that require certificates (excludes anonymous suites)
+    // This list provides compatibility across different JDK versions and platforms
+    private static final String CERTIFICATE_REQUIRED_CIPHER_SUITES = String.join(",",
+            "TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256",
+            "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384",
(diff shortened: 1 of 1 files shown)


chrjohn:
Does it help to only backport it without a 2.3.x release? 3.0.0 will be released in the near future but no plans on 2.3.x currently...

PR #1181, opened 2026-04-08
reset disabled session per `SessionSchedule` to avoid message loss on Logon

### LINKED ISSUE #965: Enabling a previously disabled session might lead to message loss
**Describe the bug**
When a session is disabled (i.e. after calling `logout()`) it will no longer get reset as defined by its `SessionSchedule`. This is because `next()` returns early when a session is disabled:

https://github.com/quickfix-j/quickfixj/blob/f8ba074829c5e362e5caf1820804be878d02b63d/quickfixj-core/src/main/java/quickfix/Session.java#L1952-L1963

**To Reproduce**

Will try to create a unit test in due course.

Manual steps:
* have a session that is disabled and has not been used at the current day (i.e. there was no established session for that day)
* send messages to that Session via `sendToTarget()` - the messages will end up in the Session's message store. 
* establish a connection to the session
* based on QFJ's internal timer, the `next()` method will be called within a second and a reset will be done, effectively removing the Sessions's message store

**Expected behavior**

Session should get reset based on its schedule even when it is disabled.

Ra9huvansh:
Fixes #965

When a session is disabled (e.g. after calling logout()) and has no active connection, Session.next() was returning early before it could reach the SessionSchedule check. This meant that even if the session schedule said it was time to reset, the reset never happened.

The problem shows up when messages are sent via sendToTarget() to a disabled session. They get queued in the message store and advance the sequence numbers. When next() is called by the internal timer, it hits the early return and exits without resetting, so those queued messages are never cleared. The next time the session connects, method next() is called by the timer thread within a second, resetting the message store and losing the stored messages.

The fix is straightforward. Instead of returning immediately when the session is disabled and not logged on, we now let execution fall through to the session schedule block first, and return after it. The existing state.isResetNeeded() check already handles @philipwhiuk's suggestion of only resetting when sequence numbers have actually advanced beyond 1, so no extra logic was needed there.

Changes:
- Session.java: removed the early return for the disabled+not-logged-on case in next(), added the guard after the schedule block instead
- SessionTest.java: added testDisabledSessionIsResetBySchedule which reproduces the exact scenario from the bug report

Tests run:
- SessionTest — 72/72 passing (includes the new test)
- Full quickfixj-core test suite — 0 failures, 0 errors

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/Session.java
+++ b/quickfixj-core/src/main/java/quickfix/Session.java
@@ -1980,7 +1980,8 @@ public class Session implements Closeable {
             }
         }
 
-        // QFJ-965: allow the session schedule block above to run even when disabled,
+        // https://github.com/quickfix-j/quickfixj/issues/965
+        // allow the session schedule block above to run even when disabled,
         // so that sequence numbers are reset as scheduled and message loss is avoided.
         if (!isEnabled() && !isLoggedOn()) {
             return;
--- a/quickfixj-core/src/main/java/quickfix/Session.java
+++ b/quickfixj-core/src/main/java/quickfix/Session.java
@@ -1979,8 +1979,8 @@ public class Session implements Closeable {
         }
 
         // https://github.com/quickfix-j/quickfixj/issues/965
-        // allow the session schedule block above to run even when disabled,
-        // so that sequence numbers are reset as scheduled and message loss is avoided.
+        // allow the session schedule block above to run even when session is disabled,
+        // so that sequence numbers are reset as scheduled.
         if (!isEnabled() && !isLoggedOn()) {
             return;
         }
--- a/quickfixj-core/src/main/java/quickfix/Session.java
+++ b/quickfixj-core/src/main/java/quickfix/Session.java
@@ -1951,12 +1951,10 @@ public class Session implements Closeable {
      */
     public void next() throws IOException {
 
-        if (!isEnabled()) {
-            if (isLoggedOn()) {
-                if (!state.isLogoutSent()) {
-                    getLog().onEvent("Initiated logout request");
-                    generateLogout(state.getLogoutReason());
-                }
+        if (!isEnabled() && isLoggedOn()) {
+            if (!state.isLogoutSent()) {
+                getLog().onEvent("Initiated logout request");
+                generateLogout(state.getLogoutReason());
             }
         }
 
--- a/quickfixj-core/src/main/java/quickfix/Session.java
+++ b/quickfixj-core/src/main/java/quickfix/Session.java
@@ -1957,8 +1957,6 @@ public class Session implements Closeable {
                     getLog().onEvent("Initiated logout request");
                     generateLogout(state.getLogoutReason());
                 }
-            } else {
-                return;
             }
         }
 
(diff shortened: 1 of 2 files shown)


Ra9huvansh:
Any adjustments to make, happy to help.

chrjohn:
Not at the moment, slowly working through my backlog. ;) thank you

quickfixj-core/src/main/java/quickfix/Session.java
```diff
@@ -1982,6 +1980,12 @@ public void next() throws IOException {
             }
         }
 
+        // QFJ-965: allow the session schedule block above to run even when disabled,
```
chrjohn:
```suggestion
        // https://github.com/quickfix-j/quickfixj/issues/965
        // allow the session schedule block above to run even when disabled,
```

chrjohn:
@Ra9huvansh 
I'll let copilot take a look but don't act on its feedback, I'll let you know.

quickfixj-core/src/main/java/quickfix/Session.java
chrjohn:
```suggestion
if (!isEnabled() && isLoggedOn()) {
```

After removing the `else` branch IMHO looks nicer to merge these. Also has the same pattern than your check below the schedule code block...

chrjohn: My first thought was actually to extract the schedule block into a method and call it in the else branch. But that does not work since the schedule block might also `return` from the method.

chrjohn: @Ra9huvansh could you please fix the indentation and excess curly brace for me? Thank you :)

quickfixj-core/src/test/java/quickfix/SessionTest.java
```diff
@@ -3192,6 +3192,63 @@ public void testSend_ShouldKeepPossDupFlagAndOrigSendingTime_GivenAllowPosDupCon
...
+            assertFalse(session.isLoggedOn());
+
+            // next() should trigger a reset per the session schedule even though
+            // the session is disabled, to avoid message loss (QFJ-965)
+            session.next();
+
+            assertEquals(1, state.getNextSenderMsgSeqNum());
+            assertEquals(1, state.getNextTargetMsgSeqNum());
+            assertEquals(1, application.sessionResets);
+        }
```
chrjohn, line 3251:
```suggestion
            assertFalse(session.isEnabled());
            assertFalse(session.isLoggedOn());
        }
```
Paranoia check ;)

quickfixj-core/src/main/java/quickfix/Session.java
Ra9huvansh: sure fixing it ASAP!

quickfixj-core/src/main/java/quickfix/Session.java
```diff
@@ -1982,6 +1978,13 @@ public void next() throws IOException {
             }
         }
 
+        // https://github.com/quickfix-j/quickfixj/issues/965
+        // allow the session schedule block above to run even when disabled,
+        // so that sequence numbers are reset as scheduled and message loss is avoided.
```
chrjohn:
```suggestion
        // allow the session schedule block above to run even when session is disabled,
        // so that sequence numbers are reset as scheduled.
```

chrjohn:
@Ra9huvansh thank you 👍

PR #1236, opened 2026-05-21
Validate sequence number in `nextReject()`

### LINKED ISSUE #1235: Too high sequence numbers for `Reject` messages are not verified
Reject messages are generally expected to be part of normal message flow (unlike say SequenceReset which necessarily will take the place of a different message). Receiving a 35=3 with a given sequence number generally means not receiving an application message with the same sequence number.

However the lack of verification on "tooHigh" means that QFJ allows this:

https://github.com/quickfix-j/quickfixj/blob/b43457b6facff8a9cf1590e09903d9b9eb442086/quickfixj-core/src/main/java/quickfix/Session.java#L1316

We've seen in practice, where session times are wrong, the following from third party FIX engines.

```
(normal session flow)
...
outgoing 34=100|35=0
incoming 34=200|35=0
outgoing 34=101|35=0
incoming 34=201|35=0
outgoing 34=101|35=5|58=Out of session time
incoming 34=202|35=5|58=Acknowledged
....
(initiator resets)
....
outgoing 34=1|35=A|
**incoming 34=203|35=3|58=Invalid sequence number**
incoming 34=204|35=5|58=Invalid sequence number
...
(after acceptor resets)
...
outgoing 34=10|35=A|
incoming 34=1|35=A|
....
outgoing 34=100|35=D
incoming 34=200|35=8
incoming 34=201|35=8
incoming 34=202|35=8
**incoming 34=203|35=8**
```

Because we don't validate for too high, the engine is able to receive an application message for same sequence as the reject at a later point, which is unexpected.

What problem is QFJ avoiding by not validating the reject being too high?
Maybe we need to at least flag the Message being 'wrong' so that applications can handle it correctly?

chrjohn:
Fixes #1235

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/Session.java
+++ b/quickfixj-core/src/main/java/quickfix/Session.java
@@ -1313,7 +1313,7 @@ public class Session implements Closeable {
 
     private void nextReject(Message reject) throws FieldNotFound, RejectLogon, IncorrectDataFormat,
             IncorrectTagValue, UnsupportedMessageType, IOException, InvalidMessage {
-        if (!verify(reject, false, validateSequenceNumbers)) {
+        if (!verify(reject, validateSequenceNumbers, validateSequenceNumbers)) {
             return;
         }
         if (getExpectedTargetNum() == reject.getHeader().getInt(MsgSeqNum.FIELD)) {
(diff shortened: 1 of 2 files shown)


chrjohn:
@copilot please create a unit test that ensures that sequence numbers are always validated when receiving a Reject message. Test different scenarios (too high, too low) and check that the expected sequence number is incremented when it is the expected sequence number.

PR #1255, opened 2026-06-11
Update mina-core from 2.2.7 to 2.2.8

dafriz:
Resolves a couple of CVEs.


chrjohn:
Let's wait, seems there still is a problem in 2.2.8... https://github.com/apache/mina/pull/52#issuecomment-4679493215

PR #1259, opened 2026-06-12
feat: add `SessionStateListener` callback when a PossDup message is discarded

### LINKED ISSUE #1254: Discarding `PossDup` messages with too low seqnum
**Describe the bug**
I have an issue with resent messages being dropped by session due to "too low" sequence numbers. I know I can use the global flag to turn off validation of sequence numbers, but not sure if this is safe.

**Expected behavior**
Messages recieved with PossDupFlag=Y not discarded.

**system information:**
 - OS: Linux
 - Java version: JDK21
 - QFJ Version: 3.0.1

**Additional context**
```
            if(!state.isResetReceived()){
                if (checkTooHigh && isTargetTooHigh(msgSeqNum)) {
                    doTargetTooHigh(msg);
                    return false;
                } else if (checkTooLow && isTargetTooLow(msgSeqNum)) {
                    doTargetTooLow(msg);
                    return false;
                }
            }
```
Seing how doTargetTooLow returns boolean from validatePossDup if it is one, I believe the code above should be changed to:
```
            if(!state.isResetReceived()){
                if (checkTooHigh && isTargetTooHigh(msgSeqNum)) {
                    doTargetTooHigh(msg);
                    return false;
                } else if (checkTooLow && isTargetTooLow(msgSeqNum)) {
                    return doTargetTooLow(msg);
                }
            }
```
alternatively
```
            if(!state.isResetReceived()){
                if (checkTooHigh && isTargetTooHigh(msgSeqNum)) {
                    doTargetTooHigh(msg);
                    return false;
                } else if (checkTooLow && isTargetTooLow(msgSeqNum) && doTargetTooLow(msg) == false) {
                    return false;
                }
            }
```

I know it would help in my case, not sure if this is generally a good fix.

mvanhorn:
## Summary

`SessionStateListener` gains an observe-only callback, `onPossDupMessageDiscarded(SessionID, Message)`, fired when a PossDupFlag=Y message is discarded by the session layer. The spec-mandated discard behavior (FIX session-layer test case 4.5.1(e)) is completely unchanged; applications that need to see these messages, like the GPW WATS case in #1254, can now observe them without forking the engine.

## Why this matters

@chrjohn confirmed in #1254 that the discard is spec-correct and will not change, and proposed exactly this: "some kind of handler or callback that is called on reception of a PossDup message ... I have nothing against putting in callbacks for messages that arrive in the engine anyway." The reporter confirmed a callback covers his use case and is currently running a private fork hack. The new method is a default on `SessionStateListener` (all existing methods are defaults; `onResendRequestSatisfied` is precedent for callbacks with context), invoked via the existing `stateListener` multiplexer at the discard sites in `doTargetTooLow()` and the expected-seqnum `validatePossDup` failure branch in `verify()`. The javadoc states the message was not processed by the application and must not be mutated.

## Testing

`SessionTest` gains cases for the too-low PossDup discard (callback fires, message untouched), the expected-seqnum OrigSendingTime failure (callback fires after the session Reject, target sequence advanced accordingly), and the no-callback path for normal messages. No Java runtime was available in the implementation environment, so CI is the verification run for the suite.

Fixes #1254

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/Session.java
+++ b/quickfixj-core/src/main/java/quickfix/Session.java
@@ -1841,6 +1841,7 @@ public class Session implements Closeable {
             // Handle poss dup where msgSeq is as expected
             // FIX 4.4 Vol 2, test case 2f&g
             if (isPossibleDuplicate(msg) && !validatePossDup(msg)) {
+                stateListener.onPossDupMessageDiscarded(sessionID, msg);
                 return false;
             }
 
@@ -1881,7 +1882,9 @@ public class Session implements Closeable {
             generateLogout(text);
             throw new SessionException(text);
         }
-        return validatePossDup(msg);
+        final boolean validPossDup = validatePossDup(msg);
+        stateListener.onPossDupMessageDiscarded(sessionID, msg);
+        return validPossDup;
     }
 
     private void doBadCompID(Message msg) throws IOException, FieldNotFound {
--- a/quickfixj-core/src/main/java/quickfix/Session.java
+++ b/quickfixj-core/src/main/java/quickfix/Session.java
@@ -1834,6 +1834,7 @@ public class Session implements Closeable {
                     return false;
                 } else if (checkTooLow && isTargetTooLow(msgSeqNum)) {
                     doTargetTooLow(msg);
+                    stateListener.onPossDupMessageDiscarded(sessionID, msg);
                     return false;
                 }
             }
@@ -1874,7 +1875,7 @@ public class Session implements Closeable {
         return true;
     }
 
-    private boolean doTargetTooLow(Message msg) throws FieldNotFound, IOException {
+    private void doTargetTooLow(Message msg) throws FieldNotFound, IOException {
         if (!isPossibleDuplicate(msg)) {
             final int msgSeqNum = msg.getHeader().getInt(MsgSeqNum.FIELD);
             final String text = "MsgSeqNum too low, expecting " + getExpectedTargetNum()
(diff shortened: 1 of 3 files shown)


chrjohn:
@mvanhorn thanks for the PR, will have a look...
ATM unit test is failing.
`Invalid use of argument matchers!`

mvanhorn:
Thanks @chrjohn. Fixed in `edd22ad` - the failing test mixed a raw `session.getSessionID()` with `messageCaptor.capture()` in the same `verify(...)`, which Mockito rejects (all args must be matchers once one is). Wrapped the session ID in `eq(...)` and added the import.

Verified locally on JDK 21: the four `onPossDupMessageDiscarded` tests in `SessionTest` now pass (Tests run: 4, Failures: 0, Errors: 0). CI should go green on the next run.

quickfixj-core/src/main/java/quickfix/Session.java
```diff
@@ -1881,7 +1882,9 @@ private boolean doTargetTooLow(Message msg) throws FieldNotFound, IOException {
             generateLogout(text);
             throw new SessionException(text);
         }
-        return validatePossDup(msg);
+        final boolean validPossDup = validatePossDup(msg);
```
chrjohn: @mvanhorn this assignment is not used at all. Edit: what I mean is that the return value of `validPossDup` is not evaluated when `doTargetTooLow()` is called.
Maybe we can move this block to line 1836? Then it will look more similar to the call in line 1844.
Moreover, I propose to remove the return value of `doTargetTooLow()` since it is not used anywhere.

chrjohn (changes requested):

quickfixj-core/src/main/java/quickfix/Session.java
chrjohn: @mvanhorn let me know if you or your AI can pick this up. 😃

mvanhorn:
Thanks @chrjohn for reviewing and merging the SessionStateListener PossDup callback.

chrjohn:
@mvanhorn you're welcome. Thank you for your contribution. 👍

PR #1271, opened 2026-07-01
feat: Add option to bypass `InetSocketAddress` reverse DNS lookup

t-fitchie-meur:
This is a feature to fix something that's not really a bug, but probably not an intentional feature either, and is definitely an edge case.

### Summary

Currently, all of the host resolution code in quickfixj uses InetSocketAddress::getHostName, which performs a [reverse DNS lookup](https://docs.oracle.com/javase/8/docs/api/java/net/InetSocketAddress.html) that may block for several seconds in certain cases, such as if you are attempting to connect to an IP address that does not have a PTR DNS record. I have observed what I believe is the following race during setup of such a FIX connection: 
- Reverse DNS lookup pre-TCP connect in IoSessionInitiator (this is fine, but may elongate session start-up time)
- Establish TCP connection
- Another Reverse DNS lookup in InitiatorSslFilter whilst attempting to initialize the TLS handshake
- In the time we're blocking waiting for the Reverse DNS lookup, the acceptor times out waiting for us to initialize the TLS handshake
- When the Reverse DNS finally times out, the acceptor has closed the TCP connection, so we get END_OF_STREAM

In scenarios where we are connecting directly to an IP address, most of the time the reverse DNS lookup does not provide any functionality, so I propose that we add the option to bypass this.

### Sample configuration in which the issue was discovered:

ConnectionType=initiator
SocketUseSSL=Y
EnabledProtocols=TLSv1.3
SocketConnectHost=1.2.3.4
SocketConnectPort=1234

JDK: 25
QuickFIX/J version: 3.0.1

Without the fix applied, I see the following pattern in the logs, repeating every retry interval:

MINA session created: local=/---.---.---.--/-----, remote=/---.---.---.--/-----
ERROR ... NullPointerException ... SslFilter.filterWrite ... sslHandler is null
Disconnecting: Encountered END_OF_STREAM

With the fix applied, things connect immediately, with no wait for the reverse DNS lookup.

### Change safety

The reverse DNS setting defaults to enabled, so this should not introduce any regressions


chrjohn:
@t-fitchie-meur thanks for the PR. 👍  I will take a look. First thing that came to my mind: could you please add the config setting to `configuration.md` also? I am also thinking about the naming of the setting, but that can be easily changed later on.
@the-thing any thoughts? Thanks. 😃

quickfixj-core/src/main/java/quickfix/mina/ssl/InitiatorSslFilter.java
```diff
@@ -13,18 +14,20 @@
...
 
     private final String sniHostName;
+    private final HostResolutionStrategy hostResolutionStrategy;
 
-    public InitiatorSslFilter(SSLContext sslContext, String sniHostName) {
+    public InitiatorSslFilter(SSLContext sslContext, String sniHostName, HostResolutionStrategy hostResolutionStrategy) {
         super(sslContext, false);
         this.sniHostName = sniHostName;
+        this.hostResolutionStrategy = hostResolutionStrategy;
     }
```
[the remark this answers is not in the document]
chrjohn: This should be no problem since the class should not be used from outside of QFJ. Maybe we should either document this or prevent its usage outside of QFJ.

quickfixj-core/src/main/java/quickfix/mina/initiator/IoSessionInitiator.java
```diff
@@ -65,9 +66,9 @@ public class IoSessionInitiator {
     private Future<?> reconnectFuture;
 
     public IoSessionInitiator(Session fixSession, SocketAddress[] socketAddresses,
-            SocketAddress localAddress, int connectTimeout, int[] reconnectIntervalInSeconds,
-            ScheduledExecutorService executor, SessionSettings sessionSettings, NetworkingOptions networkingOptions,
-            EventHandlingStrategy eventHandlingStrategy,
+            SocketAddress localAddress, HostResolutionStrategy hostResolutionStrategy, int connectTimeout,
+            int[] reconnectIntervalInSeconds, ScheduledExecutorService executor, SessionSettings sessionSettings,
+            NetworkingOptions networkingOptions, EventHandlingStrategy eventHandlingStrategy,
             IoFilterChainBuilder userIoFilterChainBuilder, boolean sslEnabled, SSLConfig sslConfig,
```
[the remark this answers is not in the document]
chrjohn: This should be no problem since the class should not be used from outside of QFJ. Maybe we should either document this or prevent its usage outside of QFJ.

t-fitchie-meur:
@chrjohn thanks for taking a look!

I've updated the configuration.md and the .html file. Let me know if there's anything else!

quickfixj-core/src/main/java/quickfix/mina/initiator/IoSessionInitiator.java
```diff
@@ -83,7 +84,8 @@ public IoSessionInitiator(Session fixSession, SocketAddress[] socketAddresses,
             reconnectTask = new ConnectTask(sslEnabled, socketAddresses, localAddress,
                     userIoFilterChainBuilder, fixSession, connectTimeoutMillis, reconnectIntervalInMillis,
                     sessionSettings, networkingOptions, eventHandlingStrategy, sslConfig,
-                    proxyType, proxyVersion, proxyHost, proxyPort, proxyUser, proxyPassword, proxyDomain, proxyWorkstation, log);
+                    proxyType, proxyVersion, proxyHost, proxyPort, proxyUser, proxyPassword, proxyDomain,
+                    proxyWorkstation, log, hostResolutionStrategy);
```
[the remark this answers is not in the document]
chrjohn: This should be no problem since the class should not be used from outside of QFJ. Maybe we should either document this or prevent its usage outside of QFJ.

chrjohn:
@t-fitchie-meur thanks. I've deleted the html doc in the meantime since it was outdated. Sorry :) I'll resolve the conflict.

quickfixj-core/src/main/java/quickfix/mina/ssl/InitiatorSslFilter.java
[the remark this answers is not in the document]
t-fitchie-meur: I've added javadoc comments to this class and the IoSessionInitiator to document that the class is internal to QFJ

chrjohn:
@copilot resolve the merge conflicts in this pull request

the-thing:
I had a brief look and the PR looks ok itself, but...

My concern is that the scenario that is being described shouldn't normally happen unless I missed something. The reverse DNS lookup is first performed in `quickfix.mina.initiator.IoSessionInitiator.ConnectTask#getSniHostName` to connect to the next host in the list and later when SSL filter is installed in `quickfix.mina.ssl.InitiatorSslFilter#createEngine`. 

`java.net.InetSocketAddress#getHostName` normally caches successfully resolved hostnames permanently so the resolution before connection should be instantaneous when creating SSL filter, unless DNS caching is disabled or DNS cache TTL is very small.

I'm not sure how hard is this to change, but I have a feeling that making sure that destination host is resolved only before connection is the probably the way to go.

t-fitchie-meur:
Hey @the-thing, thanks for taking a look!

The idea here is to cover the case where the session you're trying to connect to, for whatever reason, might have reverse DNS deliberately configured not to work. When this happens, we don't have a result to cache so we just retry every time `InetSocketAddress::getHostName` is called.

I agree that making sure that the destination host resolution (or resolution failure) occurs pre-connection is a better solution, but this does change the behaviour of the library, and I wasn't sure how strict quickfix is with changes like this? Happy to refactor this PR in that direction if you think it's fine.

the-thing:
No probs.

> I agree that making sure that the destination host resolution (or resolution failure) occurs pre-connection is a better solution, but this does change the behaviour of the library, and I wasn't sure how strict quickfix is with changes like this? Happy to refactor this PR in that direction if you think it's fine.

@t-fitchie-meur 

Now when I think about this. it doesn't require any change. All `java.net.InetSocketAddress` instances share the same cache, so successful connection to the end point will keep the successfully resolved hostname (ignoring some custom dns cache config).

> The idea here is to cover the case where the session you're trying to connect to, for whatever reason, might have reverse DNS deliberately configured not to work. When this happens, we don't have a result to cache so we just retry every time InetSocketAddress::getHostName is called.

@chrjohn 

The change looks good itself. This is an extra functionality, but also an additional complexity too.

the-thing (changes requested):
One suggestion would be to write an end to end test demonstrating functionality with both strategies.

Something similar to `quickfix.mina.ssl.SSLCertificateTest`, but it doesn't need to use SSL etc.

t-fitchie-meur:
> One suggestion would be to write an end to end test demonstrating functionality with both strategies.
> 
> Something similar to `quickfix.mina.ssl.SSLCertificateTest`, but it doesn't need to use SSL etc.

@the-thing 

I've added some E2E tests. To do this it made sense to add the reverse dns enabled configuration option to the acceptor as well as the initiator. Let me know what you think!

quickfixj-core/src/test/java/quickfix/mina/initiator/HostResolutionStrategyTest.java
```diff
@@ -0,0 +1,363 @@
...
+        Map<String, String> hostAliases = new HashMap<>();
+        hostAliases.put(FAKE_HOSTNAME, LOOPBACK_IP);
+        HostResolutionRequestInterceptor.INSTANCE.install(new MappedHostResolver(hostAliases), DefaultHostResolver.INSTANCE);
+    }
+
+    private void installBlockingResolver() {
+        Map<String, String> hostAliases = new HashMap<>();
+        hostAliases.put(FAKE_HOSTNAME, LOOPBACK_IP);
+        reverseDnsResolver = new BlockingReverseDnsResolver(hostAliases);
+        HostResolutionRequestInterceptor.INSTANCE.install(reverseDnsResolver, DefaultHostResolver.INSTANCE);
```
the-thing, line 246: This fails internally during first time static context construction, but this seems to have no effect as the exception is not propagated upstream.

```
java.lang.UnsupportedOperationException
	at java.base/java.util.ImmutableCollections.uoe(ImmutableCollections.java:159)
	at java.base/java.util.ImmutableCollections$AbstractImmutableMap.put(ImmutableCollections.java:1318)
	at org.burningwave.core.classes.Modules.exportToAll(Modules.java:182)
	at org.burningwave.core.classes.Modules.lambda$exportAllToAll$0(Modules.java:89)
	at java.base/java.lang.Iterable.forEach(Iterable.java:75)
	at org.burningwave.core.classes.Modules.lambda$exportAllToAll$1(Modules.java:87)
	at java.base/java.util.HashMap.forEach(HashMap.java:1430)
	at org.burningwave.core.classes.Modules.exportAllToAll(Modules.java:86)
	at org.burningwave.core.assembler.StaticComponentContainer.<clinit>(StaticComponentContainer.java:471)
	at org.burningwave.tools.net.DefaultHostResolver.<clinit>(DefaultHostResolver.java:68)
	at org.burningwave.tools.net.HostResolutionRequestInterceptor.<clinit>(HostResolutionRequestInterceptor.java:61)
	at quickfix.mina.initiator.HostResolutionStrategyTest.installBlockingResolver(HostResolutionStrategyTest.java:209)
	at quickfix.mina.initiator.HostResolutionStrategyTest.initiatorShouldAttemptToCallReverseDnsDuringLogonWhenReverseDnsIsEnabled(HostResolutionStrategyTest.java:123)
```

quickfixj-core/src/test/java/quickfix/mina/initiator/HostResolutionStrategyTest.java
```diff
@@ -0,0 +1,363 @@
...
+        initiator.start();
+
+        initiator.assertLoggedOn(INITIATOR_SESSION_ID);
+        acceptor.assertLoggedOn(ACCEPTOR_SESSION_ID);
+        assertFalse(reverseDnsResolver.awaitReverseDnsLookup(250, TimeUnit.MILLISECONDS));
+        assertEquals(0, reverseDnsResolver.getReverseDnsLookupCount());
+    }
+
+    @Test
+    public void acceptorShouldAttemptToCallReverseDnsDuringLogonWhenReverseDnsIsEnabled() throws Exception {
```
the-thing, line 162: Scenario for both acceptor / initiator resolution enabled seems to be missing.

quickfixj-core/src/test/java/quickfix/mina/initiator/HostResolutionStrategyTest.java
```diff
@@ -0,0 +1,363 @@
...
+        private final AtomicInteger reverseDnsLookupCount = new AtomicInteger();
+
+        BlockingReverseDnsResolver(Map<String, String> hostAliases) {
+            super(hostAliases);
+        }
+
+        @Override
+        public Collection<String> getAllHostNamesForHostAddress(Map<String, Object> requestData) {
+            if (!Arrays.equals((byte[]) getMethodArguments(requestData)[0], LOOPBACK_IP_BYTES)) {
+                return super.getAllHostNamesForHostAddress(requestData);
```
the-thing, line 370: This path is never hit, but this is probably expected just if new test IP are being added to test cases

quickfixj-core/src/main/java/quickfix/mina/initiator/IoSessionInitiator.java
the-thing, line 220: Not sure what to do with this one.

If SNI host name is not provided we might try to do DNS lookup on the endpoint provided. If this fails we will use original address representation. This happens during SSL filter construction, way before ssl engine is created. 

I think we should leave it as you did.

the-thing (changes requested):

quickfixj-core/src/main/java/quickfix/mina/initiator/IoSessionInitiator.java
t-fitchie-meur: I believe SNI host name is prohibited from being a literal IP address (https://datatracker.ietf.org/doc/html/rfc6066#section-3).

If we were to add the host resolution strategy into here, then it would be possible to construct a broken session by disabling reverse DNS, leading to an IP getting through to the SNI name.

If you're passing a literal IP through with SNI enabled that probably suggests some degree of misconfiguration, but the intended behaviour should be either to reverse DNS to get a host name or to return null (i.e. the current behaviour of `::getHostName` is correct in this case)

quickfixj-core/src/test/java/quickfix/mina/initiator/HostResolutionStrategyTest.java
t-fitchie-meur: I think it's nice to be more explicit when writing network-related tests, but agree it is redundant, so I can remove if you would like

quickfixj-core/src/test/java/quickfix/mina/initiator/HostResolutionStrategyTest.java
t-fitchie-meur: I've added this now 👍

quickfixj-core/src/test/java/quickfix/mina/initiator/HostResolutionStrategyTest.java
t-fitchie-meur: Where are you seeing this - I can't reproduce locally or find it when searching through the gh actions workflow logs?

the-thing: When running locally on Windows 11. I was running OpenJDK 25.

`mvn test -pl quickfixj-core -Dmaven.javadoc.skip=true -PskipBundlePlugin,minimal-fix-latest -Dtest=HostResolutionStrategyTest -Dsurefire.failIfNoSpecifiedTests=false`

quickfixj-core/src/test/java/quickfix/mina/initiator/HostResolutionStrategyTest.java
the-thing: That's good, just an observation - expected.

quickfixj-core/src/main/java/quickfix/mina/initiator/IoSessionInitiator.java
the-thing: Correct, should stay as is, but if not explicitly provided it will possibly try DNS resolve which might fail.

quickfixj-core/src/test/java/quickfix/mina/initiator/HostResolutionStrategyTest.java
t-fitchie-meur: Interesting, I'm also running Java 25 (Oracle, 25.0.3) on windows 11. Your command doesn't work for me (I probably have the repo misconfigured locally) but running:

`./mvnw test -am -pl quickfixj-core "-Dmaven.javadoc.skip=true" "-PskipBundlePlugin,minimal-fix-latest" "-Dtest=HostResolutionStrategyTest" "-Dsurefire.failIfNoSpecifiedTests=false"`

Does run for me and I can't recreate the issue.

Does it occur consistently or is it a one-off for you?

the-thing: I built the project with Maven install beforehand. I think with

`mvnw clean install -Dmaven.javadoc.skip=true -DskipTests -PskipBundlePlugin,minimal-fix-latest`

It occurs only one time, because it happens in static {} block in 

`org.burningwave.core.assembler.StaticComponentContainer`

Again. I don't think it is a problem, but I didn't look why it is happening.

the-thing (commented):
LGTM, but very optional feature.

Also, changes to `AcceptorSslFilter` made me realize that probably lack of

`org.apache.mina.filter.ssl.SslFilter#setEndpointIdentificationAlgorithm`
`org.apache.mina.filter.ssl.SslFilter#setWantClientAuth`

for SSL acceptor should be addressed.

t-fitchie-meur:
Morning @chrjohn @the-thing,

What are your thoughts on getting this merged?

chrjohn:
@the-thing I'm currently OOO. Ok if I leave this to you? You're the MINA and networking expert and I would have asked you anyway as soon as I'm back... 😃

the-thing:
Sure. I'm back tomorrow as well. I will have a look ASAP.

the-thing:
@chrjohn 

I was on a train and I initially thought this was a different issue. I have already reviewed it and I'm ok with the changes, but I have mixed feeling about merging this in general. 

Have a look at https://github.com/quickfix-j/quickfixj/pull/1271#issuecomment-5019463285 and 3 messages below.

t-fitchie-meur:
Morning all,

Regarding the optionality of the change: without this fix, QuickFIX/J cannot connect to a counterparty whose IP has no PTR record if they have certain TLS config. This configuration, whilst unusual, is entirely valid and not uncommon, including in modern FIX sessions. One example where this is explicitly configured is the new ETS UK Bond Consolidated tape

Agree on the comment that the DNS lookups are cached, but this is _only for the successful case_ (i.e. when a PTR record is present). For an unsuccessful lookup (what this PR is addressing) `java.net.InetSocketAddress#getHostName` will retry on every call and cause the TLS handshake to time out.

Would be great to get this over the line!

quickfixj-core/src/test/java/quickfix/mina/initiator/HostResolutionStrategyTest.java
```diff
@@ -0,0 +1,395 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * End-to-end test for {@link HostResolutionStrategy} configurations.
+ *
+ * @see SSLSupport#SETTING_REVERSE_DNS_ENABLED
```
chrjohn, line 64: Copy paste?

t-fitchie-meur: Not sure how that got there, I've removed

quickfixj-core/src/test/java/quickfix/mina/initiator/HostResolutionStrategyTest.java
```diff
@@ -0,0 +1,394 @@
...
+                }
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                throw new AssertionError("Interrupted while waiting to release reverse DNS lookup", e);
+            }
+
+            return super.getAllHostNamesForHostAddress(requestData);
+        }
+
+        boolean awaitReverseDnsLookup(long timeout, TimeUnit unit) throws InterruptedException {
```
chrjohn, line 388: Do we need both methods? Lookup and lookup**s**?

chrjohn:
@the-thing @t-fitchie-meur 
I understand both sides. I am all for adding features that are needed. We just need to make sure that the code is not too complex and still maintainable. That said, could you please check that you introduce no additional duplication e.g. in the acceptor and initiator filter. Currently only browsing on my mobile but I think there are some overlaps. Forgive me if I missed something. As said, I can check more thoroughly once back in the office in September. 

@the-thing what is your main pain point? 

Thank you both. 👍

quickfixj-core/src/test/java/quickfix/mina/initiator/HostResolutionStrategyTest.java
t-fitchie-meur: This was an artifact from a previous idea, have removed it now

t-fitchie-meur:
> @the-thing @t-fitchie-meur I understand both sides. I am all for adding features that are needed. We just need to make sure that the code is not too complex and still maintainable. That said, could you please check that you introduce no additional duplication e.g. in the acceptor and initiator filter. Currently only browsing on my mobile but I think there are some overlaps. Forgive me if I missed something. As said, I can check more thoroughly once back in the office in September.
> 
> @the-thing what is your main pain point?
> 
> Thank you both. 👍

Agree it makes sense to remove the duplication here as they are quite similar - I've extracted the duplication into a common base class.

the-thing:
@chrjohn 

> The idea here is to cover the case where the session you're trying to connect to, for whatever reason, might have reverse DNS deliberately configured not to work. When this happens, we don't have a result to cache so we just retry every time InetSocketAddress::getHostName is called.

It's an edge case that only (mostly?) happens when something is broken intentionally or cache TTL is set to very low. If it doesn't break anything it is possibly ok.

t-fitchie-meur:
> @chrjohn
> 
> > The idea here is to cover the case where the session you're trying to connect to, for whatever reason, might have reverse DNS deliberately configured not to work. When this happens, we don't have a result to cache so we just retry every time InetSocketAddress::getHostName is called.
> 
> It's an edge case that only (mostly?) happens when something is broken intentionally or cache TTL is set to very low. If it doesn't break anything it is possibly ok.

I agree that it’s not a common configuration, but I disagree that it’s a broken one. As far as I can tell, neither the FIX nor TLS specifications explicitly require a PTR record or DNS name when connecting to a host by IP address.

chrjohn:
Question: this change does not only apply to SSL connections, does it? 
Because the doc only talks about SSL.

t-fitchie-meur:
> Question: this change does not only apply to SSL connections, does it? Because the doc only talks about SSL.

For acceptors it is SSL only, for initiators it can operate over normal host-name resolution as well, although in practice it won't really have any effect unless misconfigured (reverse DNS disabled and a non-IP hostname)

I've clarified the docs - I still think the setting belongs under secure communication options, but happy to relocate it if need be!

chrjohn:
@t-fitchie-meur @the-thing 
thank you both for PR and review. 👍

PR #1275, opened 2026-07-03
Fix reconnection hang on HTTP proxy connect failures and stalled handshakes

### LINKED ISSUE #1274: Reconnection hangs indefinitely when ConnectFuture fails or stalls through an HTTP proxy (IoSessionInitiator)
**Describe the bug**
`IoSessionInitiator.pollConnectFuture()` does not correctly handle two failure
modes that occur specifically when connecting through an HTTP proxy
(ProxyType=http):

1. **Exceptional future completion is silently ignored in the "still pending" path.**
   When a `ConnectFuture` completes with `getSession() == null` but
   `getException() != null` (e.g. proxy returns 407, or the CONNECT handshake
   is rejected), the original code path only checked `getSession() != null`
   before falling through to logic that treats the future as still pending.
   This left `connectFuture` non-null indefinitely, which blocks
   `shouldReconnect()` from ever returning true again — the initiator never
   retries.

2. **A stalled proxy CONNECT handshake with no exception and no session
   never times out.** MINA's proxy filter can get stuck mid-handshake
   (no session resolved, no exception thrown) if the proxy accepts the TCP
   connection but never completes the CONNECT response. There is no watchdog
   to cancel this and force a reconnect.

This matches the underlying pattern described in #295 (authenticating HTTP
proxy connections) and the older reconnection issues QFJ-868 / QFJ-895, but
those were not addressing the exception-completion and stall-timeout cases
specifically for proxy connects.

**To Reproduce**
1. Configure an initiator session with:
ProxyType=http
ProxyHost=<proxy host>
ProxyPort=<proxy port>
ProxyUser=<user>
ProxyPassword=<password>

2. Point the proxy at an endpoint that either:
- rejects the CONNECT with a 407/4xx (simulates auth failure), or
- accepts the TCP connection but never responds to the CONNECT request
  (simulates a hung/misbehaving proxy).
3. Observe that after the first failed connect attempt, the initiator never
attempts to reconnect again, even though `ReconnectInterval` is configured.

**Expected behavior**
- On an exceptional `ConnectFuture` completion, `connectFuture` should be
cleared immediately and `handleConnectException()` should fire so normal
reconnect backoff resumes.
- On a stalled proxy handshake (no session, no exception) that exceeds
`connectTimeoutMillis` (max pending wait), the pending attempt should be
cancelled, the `ProxyConnector` recreated, and the reconnect timers reset
so the next scheduled reconnect fires correctly instead of firing
immediately or never.

**system information:**
 - OS: [Linux/Windows]
 - Java version [JDK8 and above]
 - QFJ Version [3.0.1]

**Additional context**
Reconnect is not attempted even after endpoint becomes available.
App restart fixes it.

**Related issues**
- #295
- QFJ-868 — IoSessionInitiator can't reconnect the disconnected session
- QFJ-895 — Reconnecting initiator does not work under some circumstances

**Proposed fix**

I've patched `pollConnectFuture()` to:
- Check `future.getException()` explicitly and route it through
`handleConnectException()`, clearing `connectFuture` right away.
- Add a `maxPendingMillis` (`connectTimeoutMillis`) watchdog: if a connect
attempt has been pending longer than this, call a new
`cancelAndResetPendingConnectAttempt()` helper that:
- force-closes any half-open `IoSession` and cancels the future,
- if `ioConnector instanceof ProxyConnector`, cancels the proxy connector's
 internal connect future and recreates the connector via
 `setupIoConnector()`,
- resets `lastConnectTime` and `lastReconnectAttemptTime` to the current
 time so the next reconnect timer tick doesn't fire immediately (since
 `pendingMillis` would otherwise already exceed `ReconnectInterval`).

I have a working patch and a JUnit test suite covering the changes.

Happy to open a PR with the diff and tests attached if maintainers confirm
this is the right approach — wanted to raise the issue first in case there's
a preferred design (e.g. exposing `connectTimeoutMillis` as a new session
setting rather than reusing an existing timeout field).

rishabhagl:
### Summary

Fixes #1274 — reconnection permanently stalls when connecting through an HTTP proxy if the proxy connect fails with an exception, or if the CONNECT handshake stalls without resolving the ConnectFuture at all.

### Problem

`IoSessionInitiator.pollConnectFuture()` only checked `future.getSession() != null`. If the future instead completed exceptionally (e.g. proxy returns 407, or the CONNECT is rejected), the code fell through into the "still pending" branch, which just logged a message and never cleared `connectFuture` or called `handleConnectException()`. Since `shouldReconnect()` requires `connectFuture == null`, the initiator got stuck and never retried.

Separately, a stalled proxy handshake (no session, no exception — MINA's proxy filter wedged mid-negotiation) had no timeout at all and could hang indefinitely.

### Fix

- `pollConnectFuture()` now explicitly checks `future.getException()` and routes it through `handleConnectException()`, clearing `connectFuture` immediately.
- Added a `connectTimeoutMillis`-based watchdog: if a connect attempt has been pending longer than this, `cancelAndResetPendingConnectAttempt()` is invoked to:
  - force-close any half-open `IoSession` and cancel the future,
  - if `ioConnector instanceof ProxyConnector`, cancel the proxy connector's internal connect future and recreate the connector via `setupIoConnector()`,
  - reset `lastConnectTime`/`lastReconnectAttemptTime` to the current time so the next scheduled reconnect doesn't fire immediately (previously `pendingMillis` would already exceed `ReconnectInterval`, causing a premature/rapid retry loop) or never (if left unset).

### Testing

Added `IoSessionInitiatorConnectTaskTest`, which targets the private `IoSessionInitiator.ConnectTask` polling state machine via reflection (the fix lives inside this private class, so tests construct `ConnectTask` directly and invoke `pollConnectFuture()`, `cancelAndResetPendingConnectAttempt()`, and `handleConnectException()` reflectively).

Coverage includes:

- **Pending connect exceeds `connectTimeoutMillis`** — future is cancelled, timeout events are logged, `onConnectException` fires, `connectFuture` clears, failure count increments once, and both timers reset so `shouldReconnect()` correctly waits out the full `ReconnectInterval`.
- **Pending connect within timeout** — future is left untouched, no timeout logs/callbacks fire, failure count stays at zero.
- **Future completes with a real exception (not a timeout)** — routes through the existing exception path, not the new watchdog: no explicit cancel, correct error logged, future cleared, failure count incremented.
- **Successful connect** — session stored, future cleared, failure count and address index reset, `lastConnectTime` updated.
- **Proxy connect timeout** (via a `TestProxyConnector` stub) — confirms the pending future is cancelled, `cancelConnectFuture()` is called, and the connector instance is recreated (not reused).
- **`cancelAndResetPendingConnectAttempt`** — closes a half-open session before cancelling the future and separately confirms proxy-connector cancellation/recreation in isolation.
- **`handleConnectException`** — unwraps nested causes, increments failure count, clears the future, and notifies the state listener.

All existing tests in `quickfixj-core` pass locally with this change.

### Related
- Fixes #1274
- Related to #295, QFJ-868, QFJ-895

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/mina/initiator/IoSessionInitiator.java
+++ b/quickfixj-core/src/main/java/quickfix/mina/initiator/IoSessionInitiator.java
@@ -252,22 +252,113 @@ public class IoSessionInitiator {
         }
 
         private void pollConnectFuture() {
+            ConnectFuture future = connectFuture;
+            if (future == null) {
+                return;
+            }
+
             try {
-                connectFuture.awaitUninterruptibly(CONNECT_POLL_TIMEOUT);
-                if (connectFuture.getSession() != null) {
-                    ioSession = connectFuture.getSession();
+                future.awaitUninterruptibly(CONNECT_POLL_TIMEOUT);
+
+                IoSession session = future.getSession();
+                if (session != null) {
+                    ioSession = session;
                     connectionFailureCount = 0;
                     nextSocketAddressIndex = 0;
                     lastConnectTime = System.currentTimeMillis();
                     connectFuture = null;
-                } else {
-                    fixSession.getLog().onEvent(
-                            "Pending connection not established after "
-                                    + (System.currentTimeMillis() - lastReconnectAttemptTime)
-                                    + " ms.");
+                    return;
                 }
+
+                Throwable exception = future.getException();
+                if (exception != null) {
+                    connectFuture = null;
+
+                    /*
+                     * Do not call cancelPendingConnectAttempt() here.
+                     * The future already completed with an exception.
+                     * Just release our reference to the connector.
+                     */
+
+                    handleConnectException(exception);
+                    return;
+                }
+                long now = System.currentTimeMillis();
+                long pendingMillis = now - lastReconnectAttemptTime;
+
+                fixSession.getLog().onEvent(
+                            "Pending connection not established after "
+                                + pendingMillis
+                                    + " ms.");
+
+                long maxPendingMillis = connectTimeoutMillis;
+
+                if (maxPendingMillis > 0 && pendingMillis >= maxPendingMillis) {
+                    fixSession.getLog().onEvent(
+                            "Pending connection exceeded max wait of "
+                                    + maxPendingMillis
+                                    + " ms; cancelling connect attempt and allowing reconnect."
+                    );
+
+                    try {
+                        cancelAndResetPendingConnectAttempt(future);
+                    } catch (Throwable cancelException) {
+                        fixSession.getLog().onEvent(
+                                "Exception while cancelling pending connect future: "
+                                        + cancelException
+                        );
+                    }
+
+                    /*
+                     * Important:
+                     * Reset reconnect timing from this failure moment.
+                     * Without this, the next timer tick may reconnect immediately
+                     * because pendingMillis has already exceeded ReconnectInterval.
+                     */
+                    lastConnectTime = now;
+                    lastReconnectAttemptTime = now;
+
+                    handleConnectException(new IOException(
+                            "Connect attempt exceeded max pending time of " + maxPendingMillis + " ms"
+                    ));
+                }
+
             } catch (Throwable e) {
                 handleConnectException(e);
+            }
+        }
+
+        synchronized private void cancelAndResetPendingConnectAttempt(ConnectFuture future) throws ConfigError, GeneralSecurityException {
+            try {
+                if (future != null) {
+                    IoSession session = future.getSession();
+                    if (session != null) {
+                        session.closeNow();
+                    }
+
+                    future.cancel();
+                }
+            } catch (Throwable e) {
+                fixSession.getLog().onEvent(
+                        "Exception while cancelling pending connect future: " + e
+                );
+            }
+
+            if (ioConnector instanceof ProxyConnector) {
+                try {
+                    ((ProxyConnector) ioConnector).cancelConnectFuture();
+                } catch (Throwable e) {
+                    fixSession.getLog().onEvent(
+                            "Exception while cancelling proxy connector future: " + e
+                    );
+                }
+
+                try {
+                    setupIoConnector();
+                } catch (Throwable e) {
+                    fixSession.getLog().onEvent("Exception while recreating proxy connector: " + e);
+                    throw e;
+                }
             }
         }
 
--- a/quickfixj-core/src/main/java/quickfix/mina/initiator/IoSessionInitiator.java
+++ b/quickfixj-core/src/main/java/quickfix/mina/initiator/IoSessionInitiator.java
@@ -303,7 +303,7 @@ public class IoSessionInitiator {
                     try {
                         cancelAndResetPendingConnectAttempt(future);
                     } catch (Throwable cancelException) {
-                        fixSession.getLog().onEvent(
+                        fixSession.getLog().onWarnEvent(
                                 "Exception while cancelling pending connect future: "
                                         + cancelException
                         );
@@ -339,7 +339,7 @@ public class IoSessionInitiator {
                     future.cancel();
                 }
             } catch (Throwable e) {
-                fixSession.getLog().onEvent(
+                fixSession.getLog().onWarnEvent(
                         "Exception while cancelling pending connect future: " + e
                 );
             }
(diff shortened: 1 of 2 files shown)


quickfixj-core/src/main/java/quickfix/mina/initiator/IoSessionInitiator.java
```diff
@@ -252,25 +252,116 @@ private void connect() {
...
+                            "Connect attempt exceeded max pending time of " + maxPendingMillis + " ms"
+                    ));
                 }
+
             } catch (Throwable e) {
                 handleConnectException(e);
             }
         }
 
+        synchronized private void cancelAndResetPendingConnectAttempt(ConnectFuture future) throws ConfigError, GeneralSecurityException {
```
chrjohn: Why is this synchronized?

rishabhagl: I thought connector may be used during stop() and may require synchronization. Is it not the case?

quickfixj-core/src/main/java/quickfix/mina/initiator/IoSessionInitiator.java
```diff
@@ -252,25 +252,116 @@ private void connect() {
...
+                if (future != null) {
+                    IoSession session = future.getSession();
+                    if (session != null) {
+                        session.closeNow();
+                    }
+
+                    future.cancel();
+                }
+            } catch (Throwable e) {
+                fixSession.getLog().onEvent(
```
chrjohn: The logging on exceptions should probably be a WARN or ERROR events.

quickfixj-core/src/main/java/quickfix/mina/initiator/IoSessionInitiator.java
chrjohn: I think synchronization on start/stop should only prevent concurrent start/stop, but will check again.

quickfixj-core/src/main/java/quickfix/mina/initiator/IoSessionInitiator.java
rishabhagl: Thanks! I have changed the the log event type on some places. Please help verifing.

quickfixj-core/src/main/java/quickfix/mina/initiator/IoSessionInitiator.java
chrjohn:
```suggestion
        private void cancelAndResetPendingConnectAttempt(ConnectFuture future) throws ConfigError, GeneralSecurityException {
```

I think it can safely be removed. `pollConnect()` is only called from methods which originate from `run()` which is called by a single thread from the executor.

quickfixj-core/src/main/java/quickfix/mina/initiator/IoSessionInitiator.java
```diff
@@ -252,25 +252,116 @@ private void connect() {
...
+                try {
+                    ((ProxyConnector) ioConnector).cancelConnectFuture();
+                } catch (Throwable e) {
+                    fixSession.getLog().onWarnEvent(
+                            "Exception while cancelling proxy connector future: " + e
+                    );
+                }
+
+                try {
+                    setupIoConnector();
```
chrjohn, line 357: Why is this only done on a `ProxyConnector`? I assume because it has some internal state that needs to be handled differently? Maybe add some explanation?
Edit: I'm still thinking about if this could conflict with the dispose-logic.

rishabhagl: I got this exception when reused IoConnector for proxy.
java.lang.IllegalStateException: handler cannot be set while the service is active.
    at org.apache.mina.core.service.AbstractIoService.setHandler(...)
    at org.apache.mina.proxy.ProxyConnector.connect0(...)

This exception does not occure when reusing the IoConnector in non-proxy connection timeout, where the server does not reply TCP SYN.
by explanation, you mean adding code comment?

the-thing:
This is a tough one.

The refactoring looks solid to me and I like it. It seems like this specific scenario was investigated thoroughly, but it is hard for me to prove or disprove the validity of this change.

@rishabhagl 

Would you be able to demo the failing scenario without the fix with some end 2 end unit tests? It doesn't need to happen every time as these failures tend to be time dependent, but it would be cool to see what is actually failing?

There is some code aiding in HTTP proxy auth `quickfix.mina.HttpProxyServer` so it might (or might not) be useful. Also if you know any HTTP proxy servers that can be embedded into unit tests this would help tremendously.

Also what HTTP proxy software and version you are using?

PR #1303, opened 2026-08-12
Add `AllowUnknownEnumValues` setting to allow disabling enum validation

### LINKED ISSUE #1290: Add `AllowUnknownEnumValues` setting to add option to disable enumeration validation
When testing messages from market unknown enum value in messages lead to validation exception, e.g.
if PartyRole has value that is not represented in FIX.xml
```
<field number="452" name="PartyRole" type="INT">
      <value enum="1" description="EXECUTING_FIRM"/>
      <value enum="2" description="BROKER_OF_CREDIT"/>
      <value enum="3" description="CLIENT_ID"/>
```
`Rejecting invalid message: quickfix.IncorrectTagValue: Value is incorrect (out of range) for this tag, field=452 field=452:`
Please introduce settings parameter which could allow such values.

Similar feature implemented in QuickFixN - https://github.com/connamara/quickfixn/pull/929

Thanks in advance.

konradbloor:
Hi - I saw this issue and thought it would be interesting to try to submit a PR for it.

Resolves #1290

AllowUnknownEnumValues session setting added; this then skips validation of field values against the data dictionary enumerated values.  This means they won't be rejected with IncorrectTagValue.  The other validation checks are left unaffected.  Fields without enumerated values behave as they did before.  The per-field `allowOtherValues` dictionary attribute remains the targeted alternative for individual fields.

Mirrors the equivalent QuickFIX/N setting (connamara/quickfixn#929) in name, default (N), and semantics.

Unlike QuickFIX/N, the flag lives in ValidationSettings rather than in DataDictionary, since DataDictionary instances are cached/shared amongst sessions here.

**Testing** (tests written first and confirmed failing before the implementation):
- Unknown enum value rejected by default; accepted with the setting enabled — including a multiple-value field (`ExecInst`) with one invalid token, mirroring the QuickFIX/N test
- `ValidationSettings` copy constructor retains the new flag
- `DefaultSessionFactory` wiring: setting `Y` propagates to the session; absent setting defaults to false
- Full module suites pass: `quickfixj-base` (167 tests), `quickfixj-core` (1677 tests, incl. the acceptance suite)
- Built/tested with the `minimal-fix-latest,skipBundlePlugin` profiles per AGENTS.md
- Docs entry added to `configuration.md`


the-thing (changes requested):
Looks ok, but comments like these are not required. SCM or issue tracking is for history.

```
 /**
     * https://github.com/quickfix-j/quickfixj/issues/1290
     * */
```

konradbloor:
Thanks @the-thing for reviewing - good point!  I removed them.

quickfixj-base/src/main/java/quickfix/DataDictionary.java
```diff
@@ -788,7 +788,10 @@ private void checkValidFormat(ValidationSettings settings, StringField field) th
         }
     }
 
-    private void checkValue(StringField field) throws IncorrectTagValue {
+    private void checkValue(ValidationSettings settings, StringField field) throws IncorrectTagValue {
```
the-thing: I missed this one. Injecting directly boolean flag instead of full object is probably better.

the-thing (changes requested):

quickfixj-base/src/main/java/quickfix/DataDictionary.java
konradbloor: Thanks - I agree, is it worth my doing follow up PR so we have consistency amongst the neighbouring calls?

the-thing: Separate PR? - yes. 

It might be worth keeping settings passed as an object if more than two flags (or two and more) are being read in a single method?

the-thing (approved):
LGTM

quickfixj-base/src/main/java/quickfix/DataDictionary.java
konradbloor: Cool - once I'm back later I'll do this separately.  Agreed too many on one method won't help clarity, I'll have a look through and see what would be clearest 👍  Thanks again for the review.

konradbloor:
Thanks for the review on this.  Just confirming my plan - to make sure that plan works for anyone interested:

Once this has merged I'll then make a followup PR from master to make all the neighbouring calls inside DataDictionary consistent (so ValidationSettings aren't being passed when there's just, say, a single boolean that is relevant).

chrjohn:
@konradbloor thanks for the PR.
@the-thing thanks for your review. I am OOO for the next weeks so will try to keep up as good as possible.

PR #1306, opened 2026-08-19
Extract validation logic from `DataDictionary` into `DataDictionaryValidator`

konradbloor:
This is a follow-up to #1303 — there I said I would make the validation methods consistent. The review feedback on that PR was to pass a single flag rather than the whole `ValidationSettings` when a method only reads one attribute. This PR actually pulls the validation out into its own class which also holds the settings.

Looking at `DataDictionary`, it does three things: loading the XML dictionary, providing a queryable message metadata model, and validation. Ideally a change to how we validate shouldn't mean modifying `DataDictionary`, so this PR moves the validation logic (~240 lines: `validate`, `iterate` and the `check*` methods) into a new `DataDictionaryValidator` class. The existing public API is unaffected — the `DataDictionary.validate(...)` overloads are kept and delegate to the new class.

A couple of notes:

- If this is a step too far, please say and I'll change the PR to be just the scope requested before (making the validation methods consistent).
- Happy to make `DataDictionaryValidator` package-private if we should keep it internal, what do you think?

### DIFF
--- a/quickfixj-base/src/main/java/quickfix/DataDictionary.java
+++ b/quickfixj-base/src/main/java/quickfix/DataDictionary.java
@@ -595,7 +595,7 @@ public class DataDictionary {
      */
     public void validate(Message message, ValidationSettings settings) throws IncorrectTagValue, FieldNotFound,
             IncorrectDataFormat {
-        new DataDictionaryValidator(settings).validate(this, message);
+        DataDictionaryValidator.validateInternal(message, this, this, settings);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


quickfixj-base/src/main/java/quickfix/DataDictionary.java
```diff
@@ -603,7 +595,7 @@ private static <V> void copyCollection(Collection<V> lhs, Collection<V> rhs) {
      */
     public void validate(Message message, ValidationSettings settings) throws IncorrectTagValue, FieldNotFound,
             IncorrectDataFormat {
-        validate(message, false, settings);
+        new DataDictionaryValidator(settings).validate(this, message);
```
chrjohn: Currently OOO so cannot browse all changes in full, but this looks like we create a new object on every call to validate()?

konradbloor: Hi!  This is a good point.  Have fixed this in https://github.com/quickfix-j/quickfixj/pull/1306/commits/539bdf016ed65897d367727b733b3d172d698f61.  The new instance allocation for each call is gone, and nothing new has been added to the public API.

Are you comfortable with the change?  My aim has been to move the validation out so we have one fewer reason to change DataDictionary, but it does come with tradeoffs.  Your experience with the project will give you a much better intuitive sense of whether those tradeoffs are worth it!

If you would prefer actually we wind back and simply make the validation methods consistent as per 1303 I would totally understand and would close this and just follow up with that.

PR #1307, opened 2026-08-23
Add SelfMatchPreventionID field to FIX50SP2.xml

maksoodqp:
This change introduces the new tag added by Tadawul as part of the PTTP SR3 changes. The tag has been incorporated into the relevant message handling to ensure compatibility with the latest Tadawul specifications and message format requirements.


chrjohn:
This tag has been added only with EP211 so it must not be part of the vanilla SP2 dictionary.

PR #1320, opened 2026-09-07
Fixing `NoopStore` state reset

### LINKED ISSUE #1319: Sessions on `NoopStore` keep resetting
**Describe the bug**

If an acceptor with a session schedule is started outside its window, its `NoopStore`-backed sessions
never work again for the lifetime of that process, not even after the window opens. Only a restart
made inside the window helps. `JdbcStore` sessions are fine. This worked on 2.2.0.

While broken: the client logs on, gets a `Logout` about a second later, reconnects, repeats — no
heartbeat is ever exchanged. 

I suspect `NoopStore`. It holds the creation time twice, and `reset()` updates only the `Date`:

```java
private Date creationTime = new Date();
private Calendar creationTimeCalendar = SystemTime.getUtcCalendar(creationTime);

public void reset() {
    creationTime = new Date();
    nextSenderMsgSeqNum = 1;
    nextTargetMsgSeqNum = 1;      // creationTimeCalendar is not touched
}
```

The schedule check reads the other one:

```java
private boolean isCurrentSession(final long time) throws IOException {
    return sessionSchedule == null || sessionSchedule.isSameSession(
            SystemTime.getUtcCalendar(time), state.getCreationTimeCalendar());
}
```

If I read that right, the value compared against the schedule stays pinned at store construction, so
`resetIfSessionNotCurrent()` never stops firing.

**To Reproduce**
The reproducer is in Scala, sorry — that is what I had at hand. `MemoryStore` is the control: same
code, same schedule.

```scala
import java.time.{Instant, LocalDateTime, ZoneOffset}

import scala.concurrent.duration.{DurationInt, FiniteDuration}

import org.scalatest.matchers.should.Matchers
import org.scalatest.wordspec.AnyWordSpec
import quickfix._

class NoopStoreResetSpec extends AnyWordSpec with Matchers {

  private val OneInterval: FiniteDuration = 1.day

  private val schedule = {
    val sessionId = new SessionID("FIX.4.4", "ACCEPTOR", "INITIATOR")
    val settings = new SessionSettings()
    settings.setString(sessionId, "ConnectionType", "acceptor")
    settings.setString(sessionId, "StartTime", "00:00:00")
    settings.setString(sessionId, "EndTime", "23:59:59")
    new DefaultSessionSchedule(settings, sessionId)
  }

  private def isCurrentSession(store: MessageStore, nowMillis: Long): Boolean =
    schedule.isSameSession(SystemTime.getUtcCalendar(nowMillis), store.getCreationTimeCalendar)

  private def fixedAt(millis: Long): SystemTimeSource = new SystemTimeSource {
    override def getTime: Long = millis

    override def getNow: LocalDateTime =
      LocalDateTime.ofInstant(Instant.ofEpochMilli(millis), ZoneOffset.UTC)
  }

  private def currentAfterReset(store: MessageStore): Boolean = {
    assume(schedule.isSessionTime, "run this outside the one-second gap at midnight UTC")

    val nextInterval = System.currentTimeMillis() + OneInterval.toMillis
    SystemTime.setTimeSource(fixedAt(nextInterval))
    try {
      isCurrentSession(store, nextInterval) shouldBe false

      store.reset()

      info(s"after reset(): getCreationTime=${store.getCreationTime}, " +
        s"getCreationTimeCalendar=${store.getCreationTimeCalendar.getTime}")
      isCurrentSession(store, nextInterval)
    } finally {
      SystemTime.setTimeSource(null)
    }
  }

  "MemoryStore" should {
    "be current again after the reset, one schedule interval later" in {
      currentAfterReset(new MemoryStore()) shouldBe true
    }
  }

  "NoopStore" should {
    "be current again after the reset, one schedule interval later" in {
      currentAfterReset(new NoopStore()) shouldBe true
    }
  }
}
```

On 3.0.2:

```
quickfix.DefaultSessionSchedule | [FIX.4.4:ACCEPTOR->INITIATOR] daily, 00:00:00-UTC - 23:59:59-UTC
[info] NoopStoreResetSpec:
[info] MemoryStore
[info] - should be current again after the reset, one schedule interval later
[info]   + after reset(): getCreationTime=Tue Sep 08 15:49:46 GET 2026, getCreationTimeCalendar=Tue Sep 08 15:49:46 GET 2026
[info] NoopStore
[info] - should be current again after the reset, one schedule interval later *** FAILED ***
[info]   false was not equal to true (NoopStoreResetSpec.scala:58)
[info]   + after reset(): getCreationTime=Mon Sep 07 15:49:46 GET 2026, getCreationTimeCalendar=Mon Sep 07 15:49:46 GET 2026
[info] Run completed in 784 milliseconds.
[info] Total number of tests run: 2
[info] Suites: completed 1, aborted 0
[info] Tests: succeeded 1, failed 1, canceled 0, ignored 0, pending 0
```

`MemoryStore.reset()` re-pins to the engine's clock; `NoopStore.reset()` leaves it a day behind.

**Expected behavior**

`reset()` re-pins the creation time returned by `getCreationTimeCalendar()`, so a session reset inside
the current schedule interval counts as current and the session recovers without a restart.

**system information:**
- OS: Linux
- Java version: JDK 21
- QFJ Version: 3.0.2 (worked on 2.2.0)

the-thing:
Fixes https://github.com/quickfix-j/quickfixj/issues/1319

**Changes**
- new calendar created on reset (similar to other stores)
- renamed method to be more consistent with other stores


chrjohn:
@the-thing many thanks for your quick fix. 😄 👍

PR #1322, opened 2026-09-08
Fix `logonSent` state when `MessageStore` fails to persist `Logon`

### LINKED ISSUE #1302: Session state is set incorrectly when `MessageStore` cannot persist Logon message
**Describe the bug**
When QFJ receives a Logon message for an incoming session, it:
1. calls`generateLogon()`, 
2. (which) generates a Logon message, 
3. calls `sendRaw()` 
4. (finally) calls `state.setLogonSent(true)` on the Session object.

Here the return value of `sendRaw()` is not checked. and QFJ assumes a Logon message has been sent to the counterparty (even in cases where it has not). From this point on, the Application will be able to send messages to the counterparty.

if however, the call to `MessageStore.persist()` fails and the outgoing message cannot be persisted (due to an IOException for example), `Session.send()` is never invoked and message is not sent to the counterparty. But still QFJ calls `state.setLogonSent(true)`. Any further outgoing messages to the counterparty will be rejected as they will be waiting for a Logon message

**Expected behavior**
in `Session.generateLogon()`, the return value of `sendRaw()` call should be checked, and `state.setLogonSent(true)` should be called only if `sendRaw()` has returned true

**system information:**
environment/version independent. But is still present as of the latest commit on 18.5.2026 and 3.0.2

aymenjam:
## Problem

When the MessageStore fails to persist the outgoing Logon message (e.g. an IOException), `Session#generateLogon()` still marks `logonSent` as true, even though the Logon was never actually sent. This leaves the session in an inconsistent state.

Fixes #1302

## Solution

Both `generateLogon()` overloads in `Session.java` now set `logonSent` based on the actual boolean result returned by `sendRaw()`, instead of setting it unconditionally beforehand.

## Testing

Added `testLogonNotMarkedAsSentWhenMessageStorePersistFails` in `SessionTest.java`, which mocks a `MessageStore` that throws an `IOException` on persist, and verifies that `logonSent` remains false in that case.

All existing `SessionTest` tests still pass.

### DIFF
--- a/quickfixj-core/src/main/java/quickfix/Session.java
+++ b/quickfixj-core/src/main/java/quickfix/Session.java
@@ -2107,7 +2107,6 @@ public class Session implements Closeable {
         }
         state.setLastReceivedTime(SystemTime.currentTimeMillis());
         state.clearTestRequestCounter();
-        state.setLogonSent(true);
         logonAttempts++;
 
         if (enableNextExpectedMsgSeqNum) {
@@ -2117,7 +2116,9 @@ public class Session implements Closeable {
         }
 
         setLogonTags(logon);
-        return sendRaw(logon, 0);
+        final boolean result = sendRaw(logon, 0);
+        state.setLogonSent(result);
+        return result;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


chrjohn:
Thanks for the PR @aymenjam 👍

